PR #47, opened 2017-07-19
Document new discovery URL

bootstraponline:
Google's new discovery service supports multiple formats. For example:

> https://cloudtrace.googleapis.com/$discovery/swagger2?version=v1
https://cloudtrace.googleapis.com/$discovery/rest?version=v1


renas:
Nice. Thanks @bootstraponline

PR #52, opened 2017-07-24
Add --fail to curl recipe, use full parameter names.

artem-zinnatullin:
This PR add `--fail` which will make curl fail if response code is not 2xx, otherwise curl will happily write something like `404` page to the `--output` file and then `java -jar` will fail with unreadable error :)

Also, changed `-o` and `-L` to their full parameter names so people could better understand what's going on.


renas:
Thanks for contributing.

PR #71, opened 2017-10-23
Update README.md

scompt:
Readme incorrectly references `flank.properties` instead of `config.properties`.


renas:
Thanks for contributing.

PR #73, opened 2017-11-01
add support for --use-orchestrator

### LINKED ISSUE #63: support android test orchestrator flag
[Google Cloud SDK 167](https://cloud.google.com/sdk/docs/release-notes): "Promoted Firebase Test Lab to beta for running instrumentation test with Android Test Orchestrator. Use --use-orchestrator to run with Orchestrator."

This ["allows you to run each of your app's tests within its own invocation of Instrumentation"](https://developer.android.com/training/testing/junit-runner.html#using-android-test-orchestrator) and avoid shared state and crashes taking out other tests.

Would it be possible to have a configuration option for enabling this?

Mauin:
Fixes #63 

Adds support for the test orchestrator. Can be enabled by the `use-orchestrator` flag in the `config.properties` file.


renas:
Looks good. Thanks for contributing.

PR #75, opened 2017-11-02
Include shard index to results-dir name

rciovati:
Using the same result-dir value causes some results artifacts (video, logcat, etc) to be overwritten because they all have the same name. In this PR we are changing the results-dir format to include the shard so that they will be in separate folders.

We additionally have to change the way we delete the upload app/instrumentation apks because looks like Firebase copies them to the all the result directories.


renas:
Looks good! Thanks for contributing.

PR #83, opened 2017-11-28
Added support for devices specification on config.properties

### LINKED ISSUE #32: Add support for device specification on config.properties
As you can do with `gcloud` command, you can specify device ids and versions, or you can specify each device with it's own config (version, orientation, locale and model)

The point is to add support for this type of configuration as well on `config.properties` so you could have a specified device with API 19 and another with API 25.

Today when trying to do this you would end up with four configurations (both devices with API 19 and 25) instead of two.

thalescm:
As [the official documentation](https://cloud.google.com/sdk/gcloud/reference/firebase/test/android/run), Firebase Test Lab has depretaced the use of `--device-ids`, `--locales`, `--orientations` and `--os-version-ids` in order to accept multiple `--device` entries.

This PR adds the possibility to use the `--device` entries by configuring, on `config.properties` the list of devices configurations.  For example:
```
environment-variables=coverage=true,coverageFile=/sdcard/tempDir/coverage.ec
directories-to-pull=/sdcard
shard-timeout=5
shard-duration=120  
devices=model:Nexus5X;version:23;orientation:portrait;locale:en,\
        model:Nexus6P;version:24,\
        model:pixel;version:26

numShards=  
shardIndex=
debug-prints=false  
fetch-xml-files=true
fetch-bucket=true
gcloud-path=
gsutil-path=
```
Will create the devices:
```
--device model=Nexus5X,locale=br,orientation=landscape,version=23
--device model=Nexus6P,locale=en,orientation=portrait,version=24
--device model=pixel,locale=en,orientation=portrait,version=26"
```

PS: Note that the project's default values are still respected
PS2: The project still support the old type of configuration (which has been also deprecated in the project in order for new adopters to use the new and correct way), so there's no breaking changes for old users.

This should fix #32

### DIFF
--- a/Flank/build.gradle
+++ b/Flank/build.gradle
@@ -8,7 +8,7 @@ repositories {
 }
 
 group 'com.walmart.otto'
-version '1.5.0'
+version '1.6.0'
 
 apply plugin: 'java'
 apply plugin: 'findbugs'
--- a/Flank/src/main/java/com/walmart/otto/Flank.java
+++ b/Flank/src/main/java/com/walmart/otto/Flank.java
@@ -4,6 +4,7 @@ import com.linkedin.dex.parser.DexParser;
 import com.linkedin.dex.parser.TestMethod;
 import com.walmart.otto.configurator.ConfigReader;
 import com.walmart.otto.configurator.Configurator;
+import com.walmart.otto.models.Device;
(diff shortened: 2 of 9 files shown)


renas:
Thanks Thales. Looks great.

PR #84, opened 2017-11-30
Test reports aggregation

rciovati:
This PR introduces a new (experimental) feature in Flank: _Test reports aggregation_. 

After each build Flank will generates, starting from the previously fetched bucket, a combined XML and HTML report and pushes them back to the GCS bucket.  The HTML report includes links to  Logcat and Video for both the single test and the entire shard. 

**HTML report in case of failing build**
screenshot  build failure

**HTML report in case of successful build**
screenshot build success

The feature is experimental and thus requires to be explicitly turned on. The idea is to gather feedbacks and fix potential issues.

### DIFF
--- a/Flank/build.gradle
+++ b/Flank/build.gradle
@@ -8,7 +8,7 @@ repositories {
 }
 
 group 'com.walmart.otto'
-version '1.7.0-SOUNDCLOUD'
+version '1.6.0'
 
 apply plugin: 'java'
 apply plugin: 'findbugs'
--- a/Flank/build.gradle
+++ b/Flank/build.gradle
@@ -8,7 +8,7 @@ repositories {
 }
 
 group 'com.walmart.otto'
-version '1.6.0'
+version '1.7.0-SOUNDCLOUD'
 
 apply plugin: 'java'
 apply plugin: 'findbugs'
@@ -22,6 +22,8 @@ tasks.withType(FindBugs) {
 
 dependencies {
     compile 'com.linkedin.dextestparser:parser:1.1.0'
+    compile 'com.samskivert:jmustache:1.13'
+    compileOnly 'com.google.code.findbugs:annotations:3.0.1'
     testCompile group: 'junit', name: 'junit', version: '4.11'
 }
 
--- a/Flank/src/main/java/com/walmart/otto/Flank.java
+++ b/Flank/src/main/java/com/walmart/otto/Flank.java
@@ -2,6 +2,9 @@ package com.walmart.otto;
 
 import com.linkedin.dex.parser.DexParser;
 import com.linkedin.dex.parser.TestMethod;
+import com.walmart.otto.aggregator.HtmlReportGenerationException;
+import com.walmart.otto.aggregator.ReportsAggregator;
+import com.walmart.otto.aggregator.XmlReportGenerationException;
 import com.walmart.otto.configurator.ConfigReader;
 import com.walmart.otto.configurator.Configurator;
 import com.walmart.otto.models.Device;
@@ -14,17 +17,25 @@ import com.walmart.otto.tools.ProcessExecutor;
 import com.walmart.otto.tools.ToolManager;
 import com.walmart.otto.utils.FileUtils;
 import com.walmart.otto.utils.FilterUtils;
-import java.io.*;
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.PrintStream;
+import java.io.UnsupportedEncodingException;
 import java.math.BigDecimal;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.ExecutionException;
 
 public class Flank {
+
   private ToolManager toolManager;
   private Configurator configurator;
+  private GsutilTool gsutilTool;
 
   public void start(String[] args)
       throws RuntimeException, IOException, InterruptedException, ExecutionException {
@@ -53,7 +64,7 @@ public class Flank {
 
     printShards(configurator, testCases.size());
 
-    GsutilTool gsutilTool = toolManager.get(GsutilTool.class);
+    gsutilTool = toolManager.get(GsutilTool.class);
 
     downloadTestTimeFile(gsutilTool, configurator.getShardDuration());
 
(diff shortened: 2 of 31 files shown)


Flank/build.gradle
```diff
@@ -22,6 +22,8 @@ tasks.withType(FindBugs) {
 
 dependencies {
     compile 'com.linkedin.dextestparser:parser:1.1.0'
+    compile 'com.samskivert:jmustache:1.13'
```
runningcode: we should specify `implementation` or `api` i think `compile` is deprecated.

Flank/src/main/java/com/walmart/otto/aggregator/ArtifactsProcessor.java
```diff
@@ -0,0 +1,233 @@
...
+      Instant startOfTargetTest,
+      Instant endOfTargetTest)
+      throws IOException, InterruptedException {
+
+    Path videoFile =
+        reportBaseDir.resolve(
+            testCase.getShardName() + File.separator + matrixName + File.separator + "video.mp4");
+
+    if (!Files.exists(videoFile)) {
+      System.err.printf(
```
runningcode: should this return an error code or throw an exception?

Flank/src/main/java/com/walmart/otto/aggregator/JunitReport.java
```diff
@@ -0,0 +1,202 @@
...
+
+      transformer.transform(new DOMSource(document), result);
+    } catch (Exception e) {
+      throw new XmlReportGenerationException(e);
+    }
+
+    System.out.println("XML report written to: " + outputFile.toString());
+  }
+
+  public TestSuite readFromFile(Path filePath) {
```
runningcode: this is a really big function, can we separate it out and add some comments? :)

Flank/src/main/java/com/walmart/otto/aggregator/JunitReport.java
```diff
@@ -0,0 +1,202 @@
...
+            Element errorElement = document.createElement(ERROR_TAG);
+            errorElement.setTextContent(exceptionMessage);
+            testCaseElement.appendChild(errorElement);
+            break;
+          case FAILURE:
+            Element failureElement = document.createElement(FAILURE_TAG);
+            failureElement.setTextContent(exceptionMessage);
+            testCaseElement.appendChild(failureElement);
+            break;
+          default:
```
runningcode: can we add the `SUCCESS` case and have the default case throw an exception?

Flank/build.gradle
rciovati: Well, I am not a Gradle expert but I think that since we are using the `java` plugin and not the `java-library` one, `compile` scope should be just fine.

bootstraponline: We don't want to be using the deprecated compile Gradle method. There's also a build performance win when using implementation.

rciovati: @bootstraponline AFAIK it's not deprecated. I can change the plugin to use the "java-library"  but, in addition to being out of scope for this PR, doesn't make much sense since Flank is not a library.

bootstraponline: I recommend reading about Gradle best practices to understand why this makes sense. Both Gradle and Google have posted blog articles on the topic. I think it's in scope for the PR. Technical debt isn't great.

rciovati: >I recommend reading about Gradle best practices to understand why this makes sense. Both Gradle and Google have posted blog articles on the topic. I think it's in scope for the PR. Technical debt isn't great.

Could you please elaborate? From my understanding using the `java-library` plugin is beneficial if you are building a library meant to included in other modules. That's why I don't see much benefit in make the change in this PR. Moreover, as I mentioned above, the `java` plugin is not deprecated so we are not adding more tech debt.

bootstraponline:
```diff
diff --git a/Flank/build.gradle b/Flank/build.gradle
index 21febab..21e6840 100644
--- a/Flank/build.gradle
+++ b/Flank/build.gradle
@@ -21,10 +21,10 @@ tasks.withType(FindBugs) {
 }
 
 dependencies {
-    compile 'com.linkedin.dextestparser:parser:1.1.0'
-    compile 'com.samskivert:jmustache:1.13'
-    compileOnly 'com.google.code.findbugs:annotations:3.0.1'
-    testCompile group: 'junit', name: 'junit', version: '4.11'
+    implementation 'com.linkedin.dextestparser:parser:1.1.0'
+    implementation 'com.samskivert:jmustache:1.13'
+    compileOnly 'com.google.code.findbugs:annotations:3.0.1' // 'provided' doesn't exist in the java plugin
+    testImplementation group: 'junit', name: 'junit', version: '4.11'
 }
 
 jar {
diff --git a/gradle/gradle/wrapper/gradle-wrapper.properties b/gradle/gradle/wrapper/gradle-wrapper.properties
index 26b61d7..702c4b6 100644
--- a/gradle/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/gradle/wrapper/gradle-wrapper.properties
@@ -2,4 +2,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.2-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.3.1-all.zip
```

See https://developer.android.com/studio/build/gradle-plugin-3-0-0-migration.html for an explanation of the new dependency configurations. Also see the official Gradle java library docs here:
https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_configurations_graph

I have applied the recommended changes in the above diff.

Flank/src/main/java/com/walmart/otto/Flank.java
```diff
@@ -71,6 +85,17 @@ public void start(String[] args)
...
   }
 
+  private void aggregateTestReports(File dir) {
+    try {
+      new ReportsAggregator(configurator, gsutilTool).aggregate(dir.toPath());
+    } catch (XmlReportGenerationException
+        | HtmlReportGenerationException
+        | IOException
+        | InterruptedException e) {
+      e.printStackTrace(); // don't fail the build for the reports
```
runningcode: if we don't fail to aggregate the reports, will the reports still exist in non-aggregated form? otherwise it would be very hard to debug a build.

rciovati: The source reports are not deleted/changed. 

The rationale for not failing the build in case of an exception during the report processing is that the processing itself might be buggy at this point and we don't want to fail a potentially successful test run. We could be more strict about this in the future. WDYT?

Flank/src/main/java/com/walmart/otto/aggregator/ArtifactsProcessor.java
rciovati: Not sure, I would rather silently shallow it at this point. It's probably better than aborting the entire execution in case for some reason we don't get the video file.

Flank/src/main/java/com/walmart/otto/Flank.java
runningcode: sounds good to me! although this is probably a better question for the maintainers.

renas:
@rciovati Thanks for this. Nice work. Will you be able to help with the maintenance of this?

rciovati:
@renas Yes, I will be able to help with the maintenance. We are actively using this feature and thus we are committed in making it better/stable.

PR #85, opened 2017-11-30
Update test app

rciovati:
Housekeeping


test_app/build.gradle
```diff
@@ -3,9 +3,10 @@
 buildscript {
     repositories {
         jcenter()
+        google()
```
runningcode: `google()` should be listed before `jcenter()` for some reason, but i forget why.

rciovati: I am not sure. Since most dependencies are on jcenter you probably want gradle to look there at first.

bootstraponline: Google is listed first so that their version of the artifacts are used instead of jcenter. The repos are resolved in order.

rciovati: @bootstraponline that's exactly my point. Since they are resolved in order and most dependencies are on jcenter to me it worth to have at first. Btw I'm not strong on this since I feel is a super minor thing, if you guys feel otherwise I'll change it.

bootstraponline: I think the issue is some Google deps are on both services, and in that case the Google maven repo is preferred.

Ideally we'd copy whatever the official Android docs are doing. I've seen Google used first.

PR #92, opened 2018-01-03
CLI command parsing

### LINKED ISSUE #91: CLI command parsing
You can read more about this in https://github.com/TestArmada/flank/issues/89

Requirements:
 - Add https://pholser.github.io/jopt-simple to parse commands
 - new CLI should work like this:
```java -jar Flank.jar -a <app-apk> -t <test-apk> -c <config-path> -p [package-name]```

`a`, `t` are required, `c` and `p` are optional.

### PR DESCRIPTION
**Pull request by Macarse**
Fixes #91

### DIFF
--- a/Flank/src/main/java/com/walmart/otto/Flank.java
+++ b/Flank/src/main/java/com/walmart/otto/Flank.java
@@ -17,19 +17,12 @@ import com.walmart.otto.tools.ProcessExecutor;
 import com.walmart.otto.tools.ToolManager;
 import com.walmart.otto.utils.FileUtils;
 import com.walmart.otto.utils.FilterUtils;
-import java.io.File;
-import java.io.FileNotFoundException;
-import java.io.IOException;
-import java.io.OutputStream;
-import java.io.PrintStream;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: Flank/src/main/java/com/walmart/otto/Flank.java [hunk 1]
```diff
@@ -113,6 +112,24 @@ public static void main(String[] args) {
     }
   }
 
+  static OptionParser getOptionParser() {
+    return new OptionParser() {
```
rciovati: should we declare `p` as well?

Macarse: oh, great catch. I will fix asap.
I don't use -p so when I run this version with my env I didn't find it. My bad.

Macarse: I just fixed it. Thanks again.

rciovati: Cool, let's wait for @renas then :)

renas:
Thanks @Macarse. Looks good!

Issue #90, opened 2018-01-03
Add more tests filtering options


rciovati:
At current state, Flank supports running tests based on:
- package (single)
- class (single)
- test name (single)

It's possible to filter tests in such a way passing the package parameter when invoking Flank. 
In addition to that is possible to run all but some tests defining the `skip-tests` property in the config file.

Taking inspiration from what [AndroidJunitRunner](https://developer.android.com/reference/android/support/test/runner/AndroidJUnitRunner.html) supports, we could add these execution options:
- running multiple tests
- running all tests in multiple classes
- running all tests except those in a particular class
- running all but a single test
- running all tests except a particular package
- running all tests with a given annotation(s)
- running all tests without a given annotation(s)

Related to https://github.com/TestArmada/flank/issues/89#issuecomment-354918303, since it would require a new CLI arguments mechanism.

renas:
@rciovati This would be a great addition! I like the CLI suggested by @Macarse as it will make it more clear what is passed in.

rciovati:
Fixed in #93

PR #93, opened 2018-01-10
Add test filtering support

rciovati:
See #90 for context.

It's now possible to pass one more (comma separated) filtering options via the `-f` option:             
-  `class foo.FooTest, bar.BarTest` executes only the tests in the given class(es);                                  
-  `class foo.FooTest#testFoo, bar.BarTest#testBar` executes only the given test(s);                         
-  `package foo, bar` executes only the tests in the given  package(s);                                              
-  `annotation foo.MyAnnotation, bar.MyAnnotation` executes only the tests annotated with the given annotation(s);   
-  `testFile path/to/file` executes only the tests listed in the given file. The file should contain a list of line separated package names or test classes and optionally methods.                                      

To invert the operator add the **not** prefix. For example, `notPackage` executes all the tests except the ones in the given package.
                                               
As before, tests annotated with an annotation containing the 'Ignore' string are skipped by default.  

Both the `-p` parameter and the `skip-tests` option have been removed. The latter could be replaced by passing the `notTestFile file.txt` option.

cc @Macarse (since you added the skip-tests feature)


Flank/e2e_test.sh
```diff
@@ -9,4 +9,4 @@ APP_APK=$APK_ROOT/app-debug.apk
 TEST_APK=$APK_ROOT/app-debug-androidTest.apk
 
 set -x
-java -jar ./build/libs/Flank-*.jar $APP_APK $TEST_APK
+java -jar ./build/libs/Flank-*.jar -a $APP_APK -t $TEST_APK -f "class com.example.app.ExampleUiTest#testHelloWorld"
```
Macarse: oh I didn't know this existed. how come this is not part of the ci?

Flank/src/main/java/com/walmart/otto/Constants.java
```diff
@@ -8,5 +8,23 @@
 
   public static final String CONFIG_PROPERTIES = "config.properties";
 
+  public static final String APP_APK_OPTION_DESCRIPTION = "Path to the app apk";
```
Macarse: perhaps we should create a separate java class for the CLI description

Flank/src/main/java/com/walmart/otto/Flank.java
```diff
@@ -233,21 +256,29 @@ private void printShards(Configurator configurator, int numberOfShards) {
 
   private List<String> getTestCaseNames(OptionSet options) {
     System.setOut(getEmptyStream());
-    List<String> filteredTests = new ArrayList<>();
+    List<TestMethod> testMethods = DexParser.findTestMethods((String) options.valueOf("t"));
+    System.setOut(originalStream);
```
Macarse: what's this for?

Flank/src/main/java/com/walmart/otto/Flank.java
```diff
@@ -233,21 +256,29 @@ private void printShards(Configurator configurator, int numberOfShards) {
...
+            .filter(filter::shouldRun)
+            .map(TestMethod::getTestName)
+            .collect(Collectors.toList());
+
+    if (filteredTests.size() == 0) {
+      throw new IllegalStateException("No tests to run!");
     }
 
-    System.setOut(originalStream);
+    System.out.printf("Running %d out of %d tests", filteredTests.size(), testMethods.size());
```
Macarse: Aren't there logs in the Flank class which prints similar info?

Flank/src/main/java/com/walmart/otto/filter/TestFilters.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+public class TestFilters {
+
+  private static final Pattern FILTER_ARGUMENT = Pattern.compile("(.+) (.+)");
```
Macarse: It might be less error prone to do `(.+)\s+(.+) `

Flank/src/main/java/com/walmart/otto/configurator/ConfigReader.java
```diff
@@ -141,6 +136,10 @@ private void setProperty(String property, String value) throws IllegalArgumentEx
         configurator.setGenerateSplitVideo(Boolean.parseBoolean(value));
         break;
 
+      case "skip-tests":
```
Macarse: hehe. I think I was the only one using this feature. I would remove it and mention the change in the changelog.

Flank/src/main/java/com/walmart/otto/filter/TestFilters.java
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  private static TestFilter parseSingleFilter(String argument) {
+    Matcher matcher = FILTER_ARGUMENT.matcher(argument);
+    if (!matcher.matches()) {
+      throw new IllegalArgumentException("Invalid argument: " + argument);
+    }
+
+    Collection<String> args =
+        Arrays.stream(matcher.group(2).split(",")).collect(Collectors.toSet());
```
Macarse: I am kind of confused with the API

The main regex is `X Y` where X is the command, how can Y contain spaces?
Doesn't `package com.p1, com.p2; class x` break the regex?

Instead of having `,` and `;` I would prefer having only `,`
User will need to write the `canonicalCommand` all the time, example:
`package com.p1, package com.p2, class x`

wdyt?

Flank/src/test/java/com/walmart/otto/filter/TestFiltersTest.java
```diff
@@ -0,0 +1,118 @@
...
+import static com.walmart.otto.filter.TestMethodFixtures.WITH_FOO_ANNOTATION_AND_PACKAGE;
+import static com.walmart.otto.filter.TestMethodFixtures.WITH_IGNORE_ANNOTATION;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import com.walmart.otto.testsupport.TestUtils;
+import java.nio.file.Path;
+import org.junit.Test;
+
+public class TestFiltersTest {
```
Macarse: I would like to see a test with `package com.p1, package com.p2, class x`

Flank/e2e_test.sh
rciovati: Not sure, maybe @renas could help here. I use that script to test my changes locally.

Flank/src/main/java/com/walmart/otto/Flank.java
rciovati: I suppose we want to discard whatever DexParser prints to stdout. That was already in place, I just moved the lines.

Flank/src/main/java/com/walmart/otto/Flank.java
rciovati: Not really, I think we print just the number of shards.

Flank/src/main/java/com/walmart/otto/configurator/ConfigReader.java
rciovati: Ok, it makes sense to me.

Flank/src/main/java/com/walmart/otto/filter/TestFilters.java
rciovati: Right, I'll change that

Flank/src/main/java/com/walmart/otto/filter/TestFilters.java
rciovati: > Instead of having , and ; I would prefer having only ,
User will need to write the canonicalCommand all the time, example:
package com.p1, package com.p2, class x

I think is a good point. I took inspiration from the [AndroidJUnitRunner](https://developer.android.com/reference/android/support/test/runner/AndroidJUnitRunner.html) interface but I think we are free to diverge and writing the canonical command all the time is cleaner. I'll change that. Thanks for the suggestion!

rciovati: After a second though, I think I'll leave it as it is because such a change would affect the semantic:

- `package com.p1, package com.p2, annotation x` means "run all the tests in package com.p1 AND package com.p2 AND annotation x" 
- `package com.p1, com.p2; annotation x` means "run all the tests in package com.p1 OR package com.p2 AND annotation x`.

Flank/src/main/java/com/walmart/otto/Flank.java
Macarse: oh, I didn't notice it was there before. sorry.

Flank/src/main/java/com/walmart/otto/filter/TestFilters.java
Macarse: Oh, I didn't know you were basing your implementation on `AndroidJUnitRunner`. I know see their API. Why didn't you just copy their implementation?

Flank/src/test/java/com/walmart/otto/filter/TestFiltersTest.java
```diff
@@ -0,0 +1,127 @@
...
+  public void testFilteringByAnnotation() {
+    TestFilter filter = TestFilters.fromCommandLineArguments("annotation Foo,Bar");
+
+    assertTrue(filter.shouldRun(WITH_FOO_ANNOTATION));
+    assertFalse(filter.shouldRun(WITHOUT_FOO_ANNOTATION));
+  }
+
+  @Test
+  public void testFilteringByAnnotationWithSpaces() {
+    TestFilter filter = TestFilters.fromCommandLineArguments("annotation Foo, Bar");
```
Macarse: based on the `AndroidJUnitRunner` api this should fail, right?

Flank/src/test/java/com/walmart/otto/filter/TestFiltersTest.java
Macarse: one more `package x; notPackage x`. What should happen?

Macarse:
>Both the -p parameter and the skip-tests option have been removed. The latter could be replaced by passing the notTestFile file.txt option.

actually, skip-tests should be replaced with `notPackage p1,p2,p3`. Right?

Flank/src/test/java/com/walmart/otto/filter/TestFiltersTest.java
rciovati: Will create a filter that return always false. When you run flank with such a filter it will complain because there are no tests to run.

rciovati:
>actually, skip-tests should be replaced with notPackage p1,p2,p3. Right? 

sure, it's an option as well.

Flank/src/test/java/com/walmart/otto/filter/TestFiltersTest.java
rciovati: I not sure. However, we are lenient and we don't fail.

Flank/src/main/java/com/walmart/otto/filter/TestFilters.java
rciovati: It would have probably taken more time to understand and adapt their code than actually writing this 100 lines of code.

Flank/src/test/java/com/walmart/otto/filter/TestFiltersTest.java
Macarse: ok, I think I see it in the code now. annotation `x, y` works because you are using `.anyMatch(annotations::contains)`. I think `package x, y` should fail because you are using startwith.

Flank/src/main/java/com/walmart/otto/filter/TestFilters.java
```diff
@@ -0,0 +1,121 @@
...
+
+  private static TestFilter parseSingleFilter(String argument) {
+    Matcher matcher = FILTER_ARGUMENT.matcher(argument);
+    if (!matcher.matches()) {
+      throw new IllegalArgumentException("Invalid argument: " + argument);
+    }
+
+    boolean isNegate = matcher.group(1) != null;
+
+    Collection<String> args =
```
Macarse: To make the api consistent, I would force no space in group(3).
Basically:
if group(3) contains \s, `throw IllegalArgumentException("space in " + group(3))`

rciovati: To be fair, it still wouldn't be super-consistent. In AndroidJUnitRunner you have in order to specify multiple options you have to declare multiple `-e` whereas in Flank you pass you just one -f and you wrap everything with quotes. As I said before, this is just inspired by AndroidJUnitRunner, it doesn't aim to be 100% compatible :)

Macarse: I mean consistent against Flank itself. 
I don't like the fact that `package x, y` might have issues and `annotation x, y` will not.

I would remove the possibility of spaces and be done with it.

rciovati: >I mean consistent against Flank itself.
I don't like the fact that package x, y might have issues and annotation x, y will not.

At the moment you are free to add spaces regardless the operator. The test covers just the annotation case but it works at the same way with every operator.

renas:
Great job with this @rciovati and thanks for the review @Macarse. Will release 2.0 early next week.

Macarse:
@rciovati thanks for the change and patience for answering my questions.
Great work!

PR #95, opened 2018-01-18
Trim arguments test filters after splitting

rciovati:
This PR fix an issue in which passing the values separated by comma followed by space would not work properly.

### DIFF
--- a/Flank/src/main/java/com/walmart/otto/filter/TestFilters.java
+++ b/Flank/src/main/java/com/walmart/otto/filter/TestFilters.java
@@ -45,7 +45,9 @@ public class TestFilters {
     boolean isNegate = matcher.group(1) != null;
 
     Collection<String> args =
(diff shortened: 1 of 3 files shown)


Flank/src/main/java/com/walmart/otto/filter/TestFilters.java
```diff
@@ -45,7 +45,9 @@ private static TestFilter parseSingleFilter(String argument) {
     boolean isNegate = matcher.group(1) != null;
 
     Collection<String> args =
-        Arrays.stream(matcher.group(3).split(",")).collect(Collectors.toSet());
+        Arrays.stream(matcher.group(3).split(","))
+            .map(String::trim) //splitting by comma will leave leading or trailing spaces
```
Macarse: again, I prefer consider this a parsing error so we support less cases, but I guess it's up to you.

renas:
Thanks @rciovati

Issue #100, opened 2018-01-24
Aggregated report throws a runtime error


ryanzor:
> Can't process test artifacts for: testSurfaceLogging
> Can't process test artifacts for: testAppQuitViaActivityFinish
> Can't process test artifacts for: testSaveObstruction
> Can't process test artifacts for: testOverflowObstruction
> Can't process test artifacts for: testTriedItEndsImpressions
> Can't process test artifacts for: testSaveObstruction
> Can't process test artifacts for: testSaveObstruction
> java.lang.RuntimeException: java.lang.RuntimeException: java.nio.file.NoSuchFileException: /project/bucket/2018-01-24_18:01:44.378000_Mlhf/41/NexusLowRes-25-en-portrait/logcat

The xml aggregated and uploaded fine.  I, unfortunately, don't know if the not found file was ever created.  I believe that the can't process test artifacts were all shards that partially failed.

renas:
@rciovati can you please have a look at this?

rciovati:
Hey @ryanzor, thanks for using the test aggregation feature and reporting issues :) 

Could you please provide the full tracktrace? We have a theory around why this is happening but we'd like to make sure there is not something else going on.

ryanzor:
Thanks for building the test aggregation feature!

Full trace from the most recent case:

```

java.lang.RuntimeException: java.lang.RuntimeException: java.nio.file.NoSuchFileException: /project/bucket/2018-01-25_17:28:48.977000_dtGC/45/NexusLowRes-25-en-portrait/logcat
--
  | at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
  | at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
  | at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
  | at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
  | at java.util.concurrent.ForkJoinTask.getThrowableException(ForkJoinTask.java:593)
  | at java.util.concurrent.ForkJoinTask.reportException(ForkJoinTask.java:677)
  | at java.util.concurrent.ForkJoinTask.invoke(ForkJoinTask.java:735)
  | at java.util.stream.ForEachOps$ForEachOp.evaluateParallel(ForEachOps.java:160)
[...]

PR #103, opened 2018-01-25
Make sure source logcat file exists before copying it

rciovati:
We were attempting to copy the source logcat file before actually checking if it exists or not. Ops :)

Related to #100


renas:
Thanks @rciovati

PR #107, opened 2018-01-30
printing billable time only when the it is more than 0)

abhagupta:


bootstraponline:
A 0s run will be billed as one minute though. I don't think it's correct to skip this.

PR #112, opened 2018-02-02
Report that tests failed if a gcloud command fails.

### LINKED ISSUE #22: Failed gcloud commands do not end the run
With debug = false, all the gcloud commands can fail and flank will happily report success. We should probably check to see if there's anything in the error stream, and if so end the process with a non-zero exit code.

runningcode:
If a gcloud command fails, that means that the build failed. We should
return a -1 error code.

Without this its impossible to know if a single shard failed while all
the other ones succeeded.

Fixes #22

### DIFF
--- a/Flank/src/main/java/com/walmart/otto/tools/GcloudTool.java
+++ b/Flank/src/main/java/com/walmart/otto/tools/GcloudTool.java
@@ -115,8 +115,8 @@ public class GcloudTool extends Tool {
         latestExecutionTime = Integer.parseInt(timeLine[1].replaceAll("\\D+", ""));
         TimeReporter.addExecutionTime(Integer.parseInt(timeLine[1].replaceAll("\\D+", "")));
       } else if (line.contains("ERROR")) {
-        //TODO retry when error is returned from FTL
-        //throw new RuntimeException(line);
+        System.out.println("Gcloud command failed to run. Reporting that tests failed.\n" + line);
+        testFailed = true;
       }
     }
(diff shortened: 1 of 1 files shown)


bootstraponline:
I think if we throw the exception then the entire run will end as soon as a failure has been encountered. If 1 out of 100 fails, we'd still want the results from the 99 that succeeded.

runningcode:
Yeah, what you're saying makes sense, but I'd rather have a correct build.

bootstraponline:
> Yeah, what you're saying makes sense, but I'd rather have a correct build.

Right, if you set the exit code correctly though you'd have both. A full run & a failed build (due to non-zero exit code).

runningcode:
@bootstraponline good idea. what do you think of this? setting `testFailed` to true and printing out the failed command to the console

bootstraponline:
I like it. The exit code should be successfully set now since this check exists:

```java
if (flank.hasTestFailed()) {
  System.exit(-1);
}
```

Only one person has commit access to Flank so I'm not able to merge this. Code looks good to me.

renas:
Thanks @runningcode makes total sense!

runningcode:
@renas @bootstraponline would it be possible to make a 2.0.1 release with this fix?

bootstraponline:
I think that's a great idea. At the moment, only @renas can merge/publish releases.

renas:
2.0.1 containing this fix has been released.

PR #119, opened 2018-02-08
Create release_notes.md

bootstraponline:


bootstraponline:
As new commits are pushed, we could update `release_notes.md` to communicate the changes to Flank.

PR #121, opened 2018-02-09
Error on large shard counts when limit break is false

bootstraponline:


bootstraponline:
Travis is :white_check_mark:

PR #150, opened 2018-03-23
Improve the format for the --no-auto-google-login flag

sridoodla:
[Negative booleans are confusing.](https://www.quora.com/Is-it-bad-practice-to-have-a-negative-boolean-variable-name-like-didntLose-instead-of-didLose)

Also specifies what the default value is, in the README.

### DIFF
--- a/Flank/src/main/java/com/walmart/otto/configurator/Configurator.java
+++ b/Flank/src/main/java/com/walmart/otto/configurator/Configurator.java
@@ -24,7 +24,7 @@ public class Configurator {
   private boolean fetchXMLFiles = true;
   private boolean debug = false;
   private boolean fetchBucket = false;
-  private boolean disableAutoGoogleLogin = false;
+  private boolean autoGoogleLogin = false;
   private int numShards = -1;
   private int shardIndex = -1;
   private int shardTimeout = 5;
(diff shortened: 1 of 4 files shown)


bootstraponline:
This might be one good reason to keep the mono repo. :) It'd help keep flags consistent. I implemented the flag already like this in Kotlin.

- https://github.com/TestArmada/flank/pull/143

Flank/src/main/java/com/walmart/otto/configurator/Configurator.java
```diff
@@ -24,7 +24,7 @@
   private boolean fetchXMLFiles = true;
   private boolean debug = false;
   private boolean fetchBucket = false;
-  private boolean disableAutoGoogleLogin = false;
+  private boolean autoGoogleLogin = false;
```
bootstraponline: The default for auto google login should be true to match the gcloud CLI.

> Automatically log into the test device using a preconfigured Google account before beginning the test. Enabled by default, use --no-auto-google-login to disable

sridoodla: I made it false by default as there have been issues with it recently. Lemme know if you still want it changed.

sridoodla:
Yeah, but because the source is not the same ( as it should be, in a proper mono repo ), why will people look into the kotlin folder?

bootstraponline:
> Yeah, but because the source is not the same ( as it should be, in a proper mono repo ), why will people look into the kotlin folder?

When I change the flags in both projects, should that be two PRs to two different repos? Or one PR to one repo? 😄 True mono repo means 1 repo for all code, shared or not.

Flank/src/main/java/com/walmart/otto/configurator/Configurator.java
bootstraponline: > I made it false by default as there have been issues with it recently. Lemme know if you still want it changed.

I think we should stay with gcloud defaults as a general principle. Google is rapidly fixing bugs so we'd be out of date with any custom changes.

bootstraponline:
I merged the other PR so there are some conflicts now on this one.

> This branch has conflicts that must be resolved

Flank/src/main/java/com/walmart/otto/configurator/Configurator.java
sridoodla: Done

bootstraponline:
Merged. Thanks!

PR #152, opened 2018-03-24
Add defaults for the configuration options to the README

doodla:
Having defaults visible helps people know exactly what is enabled and what they need to customize.

I also changed my username from `sridoodla` -> `doodla`. So some edits for that. :)


bootstraponline:
```
Auto fixup (add to ~/.gitconfig). git f will automatically commit any changes and squash them into the previous commit retaining the commit message of the previous commit.

[alias]
    f = "!sh -c '(git add . ; git commit --amend -C HEAD)' -"
```

You may want to use this alias. `git f` will then squash the commits instead of making a bunch of individual `f` commits.

bootstraponline:
Code looks great, merged!

PR #172, opened 2018-05-04
Test sharding

bootstraponline:


cloud_testing/kotlin_poc/src/main/kotlin/ftl/run/TestRunner.kt
```diff
@@ -61,8 +61,11 @@ object TestRunner {
         val jobs = arrayListOf<Deferred<TestMatrix>>()
         val runCount = config.testRuns
         val repeatShard = config.testShardChunks.size
+        val testsPerVm = config.testShardChunks.first().size
```
matthewrice345: Possible error, use firstOrNull()?.size ?: 0

bootstraponline: testShardChunks should always be non-empty and have at least one item. I updated to add an assert that'll fail fast if this isn't true.

PR #173, opened 2018-05-04
Link coloring html report

### LINKED ISSUE #164: Kotlin | Make html link colors blue
Links in the Kotlin html report are black. Changing the color to blue would make it more obvious these are clickable.

matthewrice345:
Fix #164


matthewrice345:
I'll try this again in a sec.

PR #181, opened 2018-05-17
iOS execution proof of concept

bootstraponline:


cloud_testing/kotlin_poc/src/main/kotlin/ftl/TmpiOS.kt
```diff
@@ -0,0 +1,75 @@
...
+        // todo: port iOS features from
+        // https://github.com/bootstraponline/gcloud_cli/blob/5bcba57e825fc98e690281cf69484b7ba4eb668a/google-cloud-sdk/lib/googlecloudsdk/api_lib/firebase/test/ios/matrix_creator.py#L109
+        // https://cloud.google.com/sdk/gcloud/reference/alpha/firebase/test/ios/run
+        // https://cloud.google.com/sdk/gcloud/reference/alpha/firebase/test/ios/
+
+        val xctestZipFile = "./fixtures/EarlGreyExampleSwiftTests.zip"
+        if (!File(xctestZipFile).exists()) throw RuntimeException("zip doesn't exist!")
+
+        // TODO: auto create root bucket if it doesn't exist
+        val config = YamlConfig(
```
Macarse: kt version of flank uses Yaml instead of properties file?

cloud_testing/kotlin_poc/src/main/kotlin/ftl/TmpiOS.kt
```diff
@@ -0,0 +1,75 @@
...
+                disablePerformanceMetrics = false,
+                disableVideoRecording = false,
+                testTimeoutMinutes = 5
+        )
+
+        val testTimeoutSeconds = TimeUnit.MINUTES.toSeconds(config.testTimeoutMinutes)
+
+        val runGcsPath = Utils.uniqueObjectName()
+//        val xcTestGcsPath = GcStorage.uploadXCTestZip(config, runGcsPath)
+        val xcTestGcsPath = "gs://earlgrey-swift/2018-05-17_16:46:36.711000_MMtK/EarlGreyExampleSwiftTests.zip"
```
Macarse: since you are hardcoding the gs path, I would remove the zip file from the commit so the repo doesn't grow in size.

cloud_testing/kotlin_poc/src/main/kotlin/ftl/TmpiOS.kt
bootstraponline: Yep

cloud_testing/kotlin_poc/src/main/kotlin/ftl/TmpiOS.kt
bootstraponline: I removed the zip.

PR #189, opened 2018-05-18
WIP | iOS Test Sharding

bootstraponline:
image

**To Do**

- [ ] Add unit tests for sharding
- [ ] Integrate into Flank Kotlin iOS runner
- [ ] Use `xctestrun` override API on `IosXcTest/testsZip ` to skip uploading when sharding


cloud_testing/kotlin_poc/build.gradle.kts
```diff
@@ -84,11 +84,11 @@ dependencies {
...
     // yaml config
-    compile("com.fasterxml.jackson.core:jackson-databind:2.9.4")
-    compile("com.fasterxml.jackson.module:jackson-module-kotlin:2.9.4")
-    compile("com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.9.4")
+    compile("com.fasterxml.jackson.core:jackson-databind:${Versions.JACKSON}")
+    compile("com.fasterxml.jackson.module:jackson-module-kotlin:${Versions.JACKSON}")
+    compile("com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:${Versions.JACKSON}")
 
-    testImplementation("junit:junit:4.12")
+    testImplementation(Libs.JUNIT)
```
Macarse: Why do we have full name dependencies and  `Libs.*`? Can't we use the same method for all dependencies?

cloud_testing/xctest_parser/src/main/kotlin/xctest/Shard.kt
```diff
@@ -0,0 +1,81 @@
...
+                val binaryName = File(binaryRoot).nameWithoutExtension
+                val binaryPath = Paths.get(binaryRoot, binaryName).toString()
+
+                val methods = if (isSwift) {
+                    println("Swift detected")
+                    Parse.parseSwiftTests(binaryPath)
+                } else {
+                    println("Objc detected")
+                    Parse.parseObjcTests(binaryPath)
+                }
```
Macarse: can't `methods` be a mixture of swift + objectiveC?

cloud_testing/xctest_parser/src/main/kotlin/xctest/Shard.kt
```diff
@@ -0,0 +1,81 @@
...
+
+                return methods
+            }
+        }
+
+        return setOf()
```
Macarse: Why would we have a shard without tests? Shouldn't we raise an exception here?

cloud_testing/xctest_parser/src/main/kotlin/xctest/Shard.kt
```diff
@@ -0,0 +1,81 @@
...
+            // run one method for now
+            setOnlyTestIdentifiers(testDictionary, listOf(methods.first()))
+        }
+
+        PropertyListParser.saveAsXML(root, xctestrun)
+    }
+
+    @JvmStatic
+    fun main(args: Array<String>) {
+        val testRunPath = "./src/test/kotlin/xctest/fixtures/EarlGreyExampleSwiftTests/EarlGreyExampleSwiftTests_iphoneos11.3-arm64.xctestrun"
```
Macarse: Add a `TODO` to remember removing?

cloud_testing/xctest_parser/src/test/kotlin/xctest/ShardTest.kt
```diff
@@ -0,0 +1,11 @@
+package xctest
+
+import org.junit.Test
+
+class ShardTest {
+
+    @Test
+    fun shardWorks() {
+        // TODO: test this without committing a giant binary blob to the git repo
```
Macarse: good point. Can't you build the iOS app before the tests run?

cloud_testing/xctest_parser/src/main/kotlin/xctest/Shard.kt
bootstraponline: That's an interesting question. I believe that when writing UI automation, teams will pick one language. I'm not sure Xcode supports having a testing target with tests written in both Swift and Objective C. The teams I've talked with are writing EarlGrey tests in ObjC or Swift, but not both.

cloud_testing/xctest_parser/src/main/kotlin/xctest/Shard.kt
bootstraponline: Good point, yeah I think we should raise an exception here since we don't want empty shards.

cloud_testing/xctest_parser/src/main/kotlin/xctest/Shard.kt
Macarse: Sure, using a single language makes a lot of sense but AFAIK firebase doesn't differentiate them. why wouldn't methods be:

```
  methods = Parse.parseObjcTests(binaryPath) + Parse.parseSwiftTests(binaryPath)
```

cloud_testing/xctest_parser/src/test/kotlin/xctest/ShardTest.kt
bootstraponline: The issue is the iOS app isn't even in this repo. I think we'll need to build a custom download logic to get all the pieces in place for unit testing. I'm hoping I get approval from Walmart to buy Git LFS since that's a better solution to binaries in git repos.

Macarse: mmm what about using git submodules for the iOS repo?

bootstraponline: issue with submodules is that the binary files still end up in a repo that gets bigger each time. I'll probably upload a zip as a GitHub release to a new repo and download that on demand.

cloud_testing/xctest_parser/src/main/kotlin/xctest/Shard.kt
rolyatwilson: Xcode does allow test targets to contain both Objective C and Swift tests. Xcode can also run all the tests in as a single test run. 

We could modify `Parse` to be language agnostic and run aggregate the results of both parsing methods. That would also allow end users to just specify a binary and not have to declare whether it is objc, swift, or mixed.

bootstraponline: I wonder if that's what bluepill does.

bootstraponline: looks like that's the right call, bluepill is running both and concatenating the results.

https://github.com/linkedin/bluepill/blob/f45878114e2b32899ca121c971adb0f638529d94/Source/Shared/BPXCTestFile.m#L56

cloud_testing/xctest_parser/src/test/kotlin/xctest/ShardTest.kt
bootstraponline: I uploaded the test zip here:
https://github.com/Flank/test_artifacts/releases/tag/test

cloud_testing/xctest_parser/src/main/kotlin/xctest/Shard.kt
bootstraponline: Fixed.

cloud_testing/xctest_parser/src/main/kotlin/xctest/Shard.kt
bootstraponline: Done.

PR #190, opened 2018-05-22
Download fixtures for xctest_parser

rolyatwilson:


cloud_testing/xctest_parser/src/test/kotlin/xctest/ParseTest.kt
```diff
@@ -47,6 +47,21 @@ class ParseTest {
             "testWithGreyAssertions"
     )))
 
+    init {
+        val releaseUrl = "https://github.com/Flank/test_artifacts/releases/download/v0.2/EarlGreyExampleTests.zip"
+        val fixtures = File("./src/test/kotlin/xctest/fixtures")
+        if (!fixtures.exists()) {
```
bootstraponline: instead of checking for the zip, we should check md5 of the zip to determine changes. The same release can be updated multiple times.

cloud_testing/xctest_parser/src/test/kotlin/xctest/ParseTest.kt
```diff
@@ -47,6 +51,45 @@ class ParseTest {
...
 
+    init {
+        val assetLink = remoteAssetLink()
+        if (updateRequired(assetLink)) {
+            updateFixtures(assetLink)
+        }
+    }
+
+    private fun remoteAssetLink(): Element {
+        val doc = Jsoup.connect("https://github.com/Flank/test_artifacts/releases/tag/v0.1").get()
```
bootstraponline: I think `https://github.com/Flank/test_artifacts/releases/latest` is the right URL to use, then we're not tied to an individual version.

bootstraponline: jsoup connect would benefit from retries/backoff. If this network request fails, then the test execution breaks.

https://stackoverflow.com/a/22890355

cloud_testing/xctest_parser/src/test/kotlin/xctest/ParseTest.kt
```diff
@@ -47,6 +51,45 @@ class ParseTest {
...
+            it.attr("href").endsWith(".zip") && it.attr("href").contains("releases/download")
+        } ?: throw RuntimeException("Download link not found in html!")
+    }
+
+    private fun updateRequired(remoteAssetLink: Element): Boolean {
+        val remoteAssetName = remoteAssetLink.attr("href").split("/").last()
+        val fixtures = File(fixturesPath)
+        if (!fixtures.exists()) return true
+        val localAssetPath = fixtures.listFiles().find { it.extension == "zip" } ?: return true
+        val localAssetMd5 = localAssetPath.readBytes().md5()
```
bootstraponline: the md5 should be the file name, this avoids us having to recalculate the md5 for each test run which may be expensive for large zips.

cloud_testing/xctest_parser/src/test/kotlin/xctest/ParseTest.kt
```diff
@@ -47,6 +51,45 @@ class ParseTest {
...
+
+    private fun updateFixtures(remoteAssetLink: Element) {
+        val fixtures = File(fixturesPath)
+        if (fixtures.exists()) fixtures.deleteRecursively()
+        fixtures.mkdirs()
+
+        val downloadUrl = "https://github.com${remoteAssetLink.attr("href")}"
+        val assetName = remoteAssetLink.attr("href").split("/").last()
+        val zipPath = "${fixtures.path}/$assetName"
+        val curl = "curl -L $downloadUrl -o $zipPath"
```
bootstraponline: any http request benefits from a retry.

https://stackoverflow.com/a/42873372

cloud_testing/xctest_parser/src/test/kotlin/xctest/ParseTest.kt
```diff
@@ -80,3 +123,9 @@ class ParseTest {
     }
 
 }
+
+fun ByteArray.md5(): String {
```
bootstraponline: for the download asset use case, we shouldn't need to calculate md5s

cloud_testing/xctest_parser/build.gradle.kts
```diff
@@ -64,8 +64,8 @@ tasks.withType<Test> {
 
 dependencies {
     compile(Libs.KOTLIN_STD_LIB)
-
     compile("com.googlecode.plist:dd-plist:1.20")
+    testCompile("org.jsoup:jsoup:1.11.3")
```
bootstraponline: does `testImplementation` work?

cloud_testing/xctest_parser/src/test/kotlin/xctest/ParseTest.kt
rolyatwilson: So true. I’ll look at adding that as well.

cloud_testing/xctest_parser/src/test/kotlin/xctest/ParseTest.kt
rolyatwilson: Nope. You’re right. I was using this to sum the local file. Going off the file name is a better approach anyway.

cloud_testing/xctest_parser/src/test/kotlin/xctest/ParseTest.kt
bootstraponline: I think the value of testing is in developer productivity so we'd want to optimize the feedback loop to be as performant as possible.

I don't think zip corruption is significant enough to slow down every test run. A faster way would be to check the file size. If the remote file size matches the local, then there's less chance of any corruption.

rolyatwilson:
Going to close and reopen to trigger a Bitrise build.

cloud_testing/xctest_parser/src/test/kotlin/xctest/TestArtifact.kt
```diff
@@ -0,0 +1,95 @@
...
+    init {
+        val assetLink = remoteAssetLink()
+        if (updateRequired(assetLink)) {
+            updateFixtures(assetLink)
+        }
+    }
+
+    private fun getHtml(): Document {
+        val githubUrl = "https://github.com/Flank/test_artifacts/releases/latest"
+        val maxRetries = 6
```
bootstraponline: I think it'd be great to abstract the retry code so we're not duplicating the same code for `getHtml()` and `downloadAsset`

wait is only ever used in the retry code. Maybe create a new RetryErrors class.

cloud_testing/xctest_parser/src/test/kotlin/xctest/TestArtifact.kt
```diff
@@ -0,0 +1,95 @@
...
+
+    private fun updateFixtures(remoteAssetLink: Element) {
+        val fixtures = File(fixturesPath)
+        if (fixtures.exists()) fixtures.deleteRecursively()
+        fixtures.mkdirs()
+
+        val downloadUrl = "https://github.com${remoteAssetLink.attr("href")}"
+        val assetName = remoteAssetLink.attr("href").split("/").last()
+        val zipPath = "${fixtures.path}/$assetName"
+        val curl = "curl -L $downloadUrl -o $zipPath"
```
bootstraponline: let's retry curl, I posted before how to do this.

https://stackoverflow.com/questions/42873285/curl-retry-mechanism/42873372#42873372

cloud_testing/xctest_parser/src/test/kotlin/xctest/TestArtifact.kt
```diff
@@ -0,0 +1,95 @@
...
+        downloadAsset(curl)
+
+        val unzip = "unzip $zipPath -d ${fixtures.path}"
```
bootstraponline: is there a way to quote the path? I'm wondering if this works if a user has a zip path that includes a space.

rolyatwilson: I have a fix for this. `zipPath` is now being quoted (will be available in a future PR)
```
unzip "./src/test/kotlin/xctest/fixtures/5D3D195F99DA2CA4A959442025AB3FB1.zip" -d ./src/test/kotlin/xctest/fixtures
```

bootstraponline: great, we should try and quote all paths. In this case, that includes destination.

cloud_testing/xctest_parser/src/test/kotlin/xctest/TestArtifact.kt
rolyatwilson: I have this in place now too.

cloud_testing/xctest_parser/src/test/kotlin/xctest/TestArtifact.kt
rolyatwilson: The `downloadAsset` method is no longer needed now that retries are built into the `curl` command.

bootstraponline: I still think the class refactor makes sense so that there's a path forward for code reuse that doesn't involve copy and paste

PR #191, opened 2018-05-23
iOS test sharding continued

bootstraponline:
image
```
Found xctest: ./EarlGreyExample/Debug-iphoneos/EarlGreyExampleSwift.app/PlugIns/EarlGreyExampleSwiftTests.xctest
YamlConfig
  projectId: 'delta-essence-114723'
  appApk: '',
  testApk: '',
  xctestrunZip: './EarlGreyExample.zip',
  xctestrunFile: './EarlGreyExample/EarlGreyExampleSwiftTests_iphoneos11.3-arm64.xctestrun',
  rootGcsBucket: 'earlgrey-swift',
  autoGoogleLogin: 'true',
  useOrchestrator: true,
  disablePerformanceMetrics: false,
  disableVideoRecording: false,
  testTimeoutMinutes: 30,
  testShards: 2,
  testRuns: 1,
  waitForResults: true,
  testMethods: []
  limitBreak: false

RunTests
  Running 1x using 2 VMs per run. 2 total VMs
  9 tests per VM. 17 total tests per run
```


cloud_testing/kotlin_poc/build.gradle.kts
```diff
@@ -82,6 +82,7 @@ dependencies {
     // https://search.maven.org/#search%7Cga%7C1%7Cg%3A%22com.google.apis%22%20AND%20a%3A%22google-api-services-testing%22
     // compile("com.google.apis:google-api-services-testing:v1-rev30-1.23.0")
```
Macarse: since you are going to master, remove this?

cloud_testing/kotlin_poc/src/main/kotlin/ftl/config/YamlConfig.kt
```diff
@@ -34,7 +36,7 @@ class YamlConfig(
         val testMethods: List<String> = listOf(),
         val limitBreak: Boolean = false,
         val projectId: String = getDefaultProjectId(),
-        var testShardChunks: List<List<String>> = emptyList()) {
+        var testShardChunks: Set<Set<String>> = emptySet()) {
```
Macarse: `) {` in new line?

cloud_testing/kotlin_poc/src/main/kotlin/ftl/gc/GcStorage.kt
```diff
@@ -23,22 +23,29 @@ object GcStorage {
...
-        val apkFileName = Paths.get(apk).fileName.toString()
-        val gcsApkPath = GCS_PREFIX + join(rootGcsBucket, runGcsPath, apkFileName)
+    private fun upload(file: String, rootGcsBucket: String, runGcsPath: String): String {
+        return upload(file = file,
+                fileBytes = Files.readAllBytes(Paths.get(file)),
+                rootGcsBucket = rootGcsBucket,
+                runGcsPath = runGcsPath)
+    }
+
+    private fun upload(file: String, fileBytes: ByteArray, rootGcsBucket: String, runGcsPath: String): String {
```
Macarse: I wonder if we should add some kind of android/ios in the folder path. If users are running android and iOS testing, we will see both builds together, right?

cloud_testing/kotlin_poc/src/main/kotlin/ftl/run/IosTestRunner.kt
```diff
@@ -23,21 +24,30 @@ object IosTestRunner : GenericTestRunner {
         val iosDevice = IosDevice()
                 .setIosModelId(IosCatalog.model("iphone8"))
                 .setIosVersionId(IosCatalog.version("11.2"))
-                .setLocale("en_US") // iOS doesn't support other locales or orientations
+                .setLocale("en_US") // FTL iOS doesn't currently support other locales or orientations
```
Macarse: true, what about model versions? we should be able to configure that right away.

cloud_testing/kotlin_poc/build.gradle.kts
bootstraponline: I'd like to keep the comment because we will be switching back to the official testing jar in 3 months. Currently we have to generate a local testing jar due to the use of private iOS APIs.

cloud_testing/kotlin_poc/src/main/kotlin/ftl/config/YamlConfig.kt
bootstraponline: That's the end of the class constructor. I'm matching the style here:
https://kotlinlang.org/docs/reference/classes.html

cloud_testing/kotlin_poc/src/main/kotlin/ftl/gc/GcStorage.kt
bootstraponline: If you set the root bucket, then android and iOS will go in different folders. For example `rootGcsBucket: earlgrey-swift`

cloud_testing/kotlin_poc/src/main/kotlin/ftl/run/IosTestRunner.kt
bootstraponline: Agreed, I opened a ticket:
https://github.com/TestArmada/flank/issues/193

I'm hesitant to add new features until we comprehensively test the existing code. That's being worked on in https://github.com/TestArmada/flank/pull/190

bootstraponline:
The broken CI will be fixed after https://github.com/TestArmada/flank/pull/190 merges.

PR #199, opened 2018-05-25
Add retry support directly to curl

rolyatwilson:


cloud_testing/xctest_parser/src/test/kotlin/xctest/TestArtifact.kt
```diff
@@ -80,10 +67,19 @@ abstract class TestArtifact {
...
-        val curl = "curl -L $downloadUrl -o $zipPath"
-        downloadAsset(curl)
+        val curl = """
+            curl
+            --connect-timeout 5
+            --max-time 300
+            --retry 5
+            --retry-delay 0
+            --retry-max-time 1000
+            -L $downloadUrl
```
bootstraponline: I think we'd want to quote download URL

PR #201, opened 2018-05-29
Bash

bootstraponline:
Fix https://github.com/TestArmada/flank/issues/200


cloud_testing/xctest_parser/src/main/kotlin/xctest/Bash.kt
```diff
@@ -22,10 +31,19 @@ object Bash {
...
-        process.waitFor(2, TimeUnit.MINUTES)
+        val pool = Executors.newFixedThreadPool(2)
+        val gobbleInput = StreamGobbler(process.inputStream)
+        val gobbleError = StreamGobbler(process.errorStream)
+        pool.submit(gobbleInput)
+        pool.submit(gobbleError)
+
+        process.waitFor(1, TimeUnit.MINUTES)
 
         if (process.failed()) throw RuntimeException("Command failed: $cmd")
```
Macarse: I think we should do something like:
```
if (process.failed()) {
  println("error $gobbleError.output")
  throw RuntimeException("Command failed: $cmd")
}
```

cloud_testing/xctest_parser/src/main/kotlin/xctest/Bash.kt
```diff
@@ -22,10 +31,19 @@ object Bash {
...
 
-        return process.output()
+        pool.shutdown()
+        pool.awaitTermination(1, TimeUnit.MINUTES)
+
+        return (gobbleInput.output + gobbleError.output).trim()
```
Macarse: if we get to this point, we should have only `gobbleInput.output`

cloud_testing/xctest_parser/src/main/kotlin/xctest/Bash.kt
```diff
@@ -0,0 +1,45 @@
...
+                .redirectOutput(PIPE)
+                .redirectError(PIPE)
+                .start()
+
+        val gobbleInput = StreamGobbler(process.inputStream)
+        gobbleInput.start()
+        val gobbleError = StreamGobbler(process.errorStream)
+        gobbleError.start()
+
+        process.waitFor()
```
Macarse:
```
        if (!process.waitFor(2, TimeUnit.MINUTES)) {
            process.destroy()
            throw RuntimeException("execution timed out: $this")
        }
```

bootstraponline:
> why the timeout for the join()?

default join is 

```java
 public final void join() throws InterruptedException {
        join(0);
    }
```

which waits forever.

```
  * Waits at most {@code millis} milliseconds for this thread to
  * die. A timeout of {@code 0} means to wait forever.
```

Macarse:
@bootstraponline ok, what I am saying is that we should be fine with `join()` since the process would be dead already.

bootstraponline:
> @bootstraponline ok, what I am saying is that we should be fine with join() since the process would be dead already.

good point. I removed the timeout on join.

PR #214, opened 2018-06-13
Add Android Device Configuration Support

Blackchai:


cloud_testing/kotlin_poc/src/test/kotlin/ftl/config/YamlConfigTest.kt
```diff
@@ -59,6 +59,7 @@ class YamlConfigTest {
         assert(config.waitForResults, true)
         assert(config.testMethods, listOf(testName))
         assert(config.limitBreak, false)
+        assert(config.devices, listOf(Devices("Nexus6"), Devices(version = "23", orientation = "landscape"), Devices("shamu", "22", "zh_CN", "default")))
```
bootstraponline: I think we'd want to copy `flank.yml` into a `fixtures` dir under `/Users/medwards/code/flank/cloud_testing/kotlin_poc/src/test/kotlin/ftl`. Then we don't have to modify the `flank.yml` we use for running the sample app with the `flank.yml` used by the unit tests.

cloud_testing/kotlin_poc/flank.yml
```diff
@@ -12,3 +12,11 @@ limitBreak: false
 waitForResults: true
 testMethods:
   - com.example.app.ExampleUiTest#testPasses
+devices:
+  - model: Nexus6
```
bootstraponline: I think the default here should be NexusLowRes. Then in the test fixture yaml, we can use the real devices.

cloud_testing/kotlin_poc/src/main/kotlin/ftl/gc/GcAndroidMatrix.kt
```diff
@@ -1,17 +1,16 @@
...
 
 import com.google.api.services.testing.model.AndroidMatrix
-import com.google.common.collect.Lists
+import ftl.config.Devices
 
 object GcAndroidMatrix {
 
-    fun build(
-            modelIds: String, versionIds: String, locales: String, orientations: String): AndroidMatrix {
+    fun build(deviceList: List<Devices>): AndroidMatrix {
```
yogurtearl: I don't think `AndroidMatrix` is what we want here. That will give us all permutations of the listed configs. 
I think want a `com.google.api.services.testing.model.AndroidDeviceList` instead

cloud_testing/kotlin_poc/src/main/kotlin/ftl/gc/GcAndroidMatrix.kt
```diff
@@ -1,17 +1,16 @@
 package ftl.gc
 
 import com.google.api.services.testing.model.AndroidMatrix
-import com.google.common.collect.Lists
+import ftl.config.Devices
 
 object GcAndroidMatrix {
 
-    fun build(
```
yogurtearl: I don't think AndroidMatrix is what we want here. That will give us all permutations of the listed configs.
I think want a com.google.api.services.testing.model.AndroidDeviceList instead

- [Use AndroidDeviceList for configure the devices]

bootstraponline: you're right, we'll fix.

image

bootstraponline:
the CLA bot is unhappy

bootstraponline:
Merged. Thanks!

PR #221, opened 2018-06-15
Fix duplicate OnlyTestIdentifiers (#218)

### LINKED ISSUE #218: Kotlin | Avoid writing OnlyTestIdentifiers twice
OnlyTestIdentifiers is duplicated in the xctestrun files. That's a bug. We should only have one OnlyTestIdentifiers. Adding a JUnit test will help ensure this doesn't regress in the future.

vanwagonet:
The same object was being passed and mutated in each coroutine, causing more
than one set after all the removes. Having each routine operate on their own
copy eliminates this race condition.

Fix #218

### DIFF
--- a/cloud_testing/kotlin_poc/src/main/kotlin/ftl/gc/GcIosTestMatrix.kt
+++ b/cloud_testing/kotlin_poc/src/main/kotlin/ftl/gc/GcIosTestMatrix.kt
@@ -1,6 +1,5 @@
 package ftl.gc
 
-import com.dd.plist.NSDictionary
 import com.google.api.services.testing.Testing
 import com.google.api.services.testing.model.*
 import ftl.config.YamlConfig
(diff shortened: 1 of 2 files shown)


bootstraponline:
what do you think of updating `rewrite` to work on a clone of `root` instead of parsing the same `xctestrunFile` everytime?

vanwagonet:
That was actually my first thought, but couldn't figure out how to deep clone `NSDictionary`. Is there a straightforward way to do that?

bootstraponline:
we may have to define an extension method to return a deep clone

vanwagonet:
I can look more into that, but I'm not sure maintaining a bunch of deep clone code is worth avoiding the extra file read+parse per shard.

bootstraponline:
> I can look more into that, but I'm not sure maintaining a bunch of deep clone code is worth avoiding the extra file read+parse per shard.

If it's only a few lines of code, then I'd like to solve the root of the problem (mutable state). I agree if it's a bunch of code then the current approach is preferable.

bootstraponline:
It's about ~7 lines of code: https://github.com/TestArmada/flank/pull/223

what do you think @vanwagonet ?

vanwagonet:
That looks better than this.

PR #223, opened 2018-06-15
Add Immutable test. Deep clone object in rewrite

### LINKED ISSUE #218: Kotlin | Avoid writing OnlyTestIdentifiers twice

bootstraponline:
Fix #218

### DIFF
--- a/cloud_testing/xctest_parser/src/main/kotlin/xctest/Xctestrun.kt
+++ b/cloud_testing/xctest_parser/src/main/kotlin/xctest/Xctestrun.kt
@@ -2,6 +2,7 @@ package xctest
 
 import com.dd.plist.NSArray
 import com.dd.plist.NSDictionary
+import com.dd.plist.NSObject
 import com.dd.plist.PropertyListParser
 import java.io.ByteArrayOutputStream
 import java.io.File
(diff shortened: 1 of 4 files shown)


cloud_testing/xctest_parser/src/main/kotlin/xctest/Xctestrun.kt
```diff
@@ -87,14 +88,26 @@ object Xctestrun {
         return result
     }
 
+    private fun NSDictionary.deepClone() : NSDictionary {
+        val newDictionary = NSDictionary()
+
+        this.forEach { (key, value) ->
+            val newValue = NSObject.fromJavaObject(value.toJavaObject())
```
vanwagonet: ok, the to/from JavaObject api does makes this easy. It also looks like we probably don't need to do this forEach for the top level dictionary, we could just have the entire method be `return NSObject.fromJavaObject(this.toJavaObject())`, right?

bootstraponline:
> I do think immutable rewrite is a better solution to our issue here, but can we just do the to & from dance at the top level?

that's a great point, I've simplified this to one line now.

PR #226, opened 2018-06-18
Allow xctestrunZip to be specified as a gcs path

### LINKED ISSUE #217: Kotlin | Allow gcs path to xctestrunZip
xctestrunZip should accept a gcs path so we don't have to upload the zip every time.

```yaml
xctestrunZip: gs://apps/EarlGreyExample.zip
```

vanwagonet:
Update validation to allow the gcs path, and skip uploading to gcs if the path
already points to gcs instead of a local zip file.

fix #217

### DIFF
--- a/cloud_testing/kotlin_poc/src/main/kotlin/ftl/config/YamlConfig.kt
+++ b/cloud_testing/kotlin_poc/src/main/kotlin/ftl/config/YamlConfig.kt
@@ -89,7 +89,9 @@ class YamlConfig(
     }
 
     private fun validateIos() {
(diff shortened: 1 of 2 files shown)


vanwagonet:
I was going to update tests for YamlConfig, but the `fatalError` doesn't throw an exception to be easily tested. IosTestRunner doesn't have any unit tests yet, and it would likely need a bunch of mocking I don't know how to do yet. ;)

bootstraponline:
I opened https://github.com/TestArmada/flank/issues/227 to track testing the iOS features. The exception can be caught using a JUnit rule.

The code looks good, I approved.

Issue #193, opened 2018-05-24
Kotlin | Configurable iOS models and versions


bootstraponline:
Flank should support reading models/versions from the yaml file. We should match the gcloud python CLI syntax.

```
val iosDevice = IosDevice()
                 .setIosModelId(IosCatalog.model("iphone8"))
                 .setIosVersionId(IosCatalog.version("11.2"))
```

Android configurable devices already added yaml support. All that's left is to add the iOS specific logic.

Verify multiple devices can be specified. Currently only a single device works.

PR #228, opened 2018-06-18
Filter invalid devices from iOS catalog

### LINKED ISSUE #188: Kotlin | Filter invalid devices from iOS catalog
- https://github.com/TestArmada/flank/blob/master/cloud_testing/kotlin_poc/src/main/kotlin/ftl/ios/IosCatalog.kt
- filter out all models that are missing `supportedVersionIds` 
-  validate based on combined model + supported version

vanwagonet:
If supportedVersionIds is null or empty, then don't include it in iosModelIds.
Also add supported method to check if a model/version pair is supported.

Fix #188

### DIFF
--- a/cloud_testing/kotlin_poc/src/main/kotlin/ftl/ios/IosCatalog.kt
+++ b/cloud_testing/kotlin_poc/src/main/kotlin/ftl/ios/IosCatalog.kt
@@ -19,26 +19,6 @@ Request access for your project via the following form:
 If this project has already been granted access, please email ftl-ios-feedback@google.com for support.""", e)
         }
     }
-    private val iosModelIds by lazy {
-        iosDeviceCatalog.models
-                .filter { it.supportedVersionIds?.isNotEmpty() ?: false }
-                .map { it.id }
-    }
-    private val iosVersionIds by lazy { iosDeviceCatalog.versions.map { it.id } }
-
-    fun model(modelId: String): String {
-        if (!iosModelIds.contains(modelId)) {
-            throw RuntimeException("Invalid iOS modelId '$modelId'.\nValid models are: $iosModelIds")
-        }
-        return modelId
-    }
-
-    fun version(versionId: String): String {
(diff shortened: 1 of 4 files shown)


cloud_testing/kotlin_poc/src/main/kotlin/ftl/ios/IosCatalog.kt
```diff
@@ -36,4 +39,9 @@ If this project has already been granted access, please email ftl-ios-feedback@g
         }
         return versionId
     }
+
+    fun supported(modelId: String, versionId: String): Boolean {
```
vanwagonet: I wasn't sure if we wanted a Boolean result, or to throw, like the above methods. If we do want a throw, what should be returned?

bootstraponline: Boolean is easier to test. We do need to throw an exception to alert the user if they've specified an invalid device though. Maybe that exception should be generated outside of IosCatalog.kt

bootstraponline: Probably YamlConfig should validate after parsing. What do you think?

There's a ticket to add configurable device support on iOS. That seems like a natural location for throwing the exception. #193

vanwagonet: Sounds good. Like in one of IosTestRunner, YamlConfig, or GcIosTestMatrix. When we go to use the values we validate and throw if necessary.

vanwagonet: oh yep. YamlConfig as part of #193 sounds right.

bootstraponline:
I think we can simplify the code to be only:

image

bootstraponline:
the build broke 😭

PR #229, opened 2018-06-18
Add device matrix for ios

### LINKED ISSUE #193: Kotlin | Configurable iOS models and versions

vanwagonet:
The devices specified in the yaml config are validated against the catalog of
supported ios devices, and passed to the gc api.

Fix #193

### DIFF
--- a/cloud_testing/kotlin_poc/src/main/kotlin/ftl/config/YamlConfig.kt
+++ b/cloud_testing/kotlin_poc/src/main/kotlin/ftl/config/YamlConfig.kt
@@ -7,6 +7,7 @@ import com.google.cloud.ServiceOptions
 import com.google.common.math.IntMath
 import com.linkedin.dex.parser.DexParser
 import ftl.config.FtlConstants.useMock
+import ftl.ios.IosCatalog
 import ftl.util.Utils.fatalError
 import xctest.Xctestrun
 import java.io.File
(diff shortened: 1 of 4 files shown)


cloud_testing/kotlin_poc/src/main/kotlin/ftl/config/YamlConfig.kt
```diff
@@ -94,6 +101,8 @@ class YamlConfig(
         }
         assertFileExists(xctestrunFile, "xctestrunFile")
 
+        for (device in devices) assertIosDeviceSupported(device)
```
Macarse: `devices.forEach`?

vanwagonet: I can change it if that's the preferred style. I'm still really new to Kotlin.

bootstraponline:
```kotlin
devices.forEach { device -> assertIosDeviceSupported(device) }
```

i usually prefer named params, either way works

PR #233, opened 2018-06-19
Auto create root bucket if it doesn't exist.

### LINKED ISSUE #183: Kotlin | Auto create root bucket if it doesn't exist
Flank will error if the `rootGcsBucket` doesn't exist. We should use the Google storage API to create the bucket if it doesn't already exist.

See how to do that here: https://github.com/instructure/android-uno/blob/63cef7b172b3bebcc36216e4e2daf7883e951bb6/automation/cloud_build_metrics/src/main/kotlin/support/upload/apks/CloudStorageApp.kt#L63

bradylarson:
Fixes #183

### DIFF
--- a/cloud_testing/kotlin_poc/src/main/kotlin/ftl/config/YamlConfig.kt
+++ b/cloud_testing/kotlin_poc/src/main/kotlin/ftl/config/YamlConfig.kt
@@ -48,9 +48,8 @@ class YamlConfig(
 
 
     private val storage = StorageOptions.newBuilder().setProjectId(projectId).build().service
-    private val bucketLabel: Map<String, String> = mapOf(Pair("flank", "apks"))
+    private val bucketLabel: Map<String, String> = mapOf(Pair("flank", ""))
     private val storageLocation = "us-central1"
-    private val daysToLive = 30
 
     fun getGcsBucket() : String {
         val bucket = storage.list().values?.find { it.name == rootGcsBucket }
@@ -60,7 +59,6 @@ class YamlConfig(
                 .setStorageClass(StorageClass.REGIONAL)
                 .setLocation(storageLocation)
                 .setLabels(bucketLabel)
-                .setDeleteRules(listOf(BucketInfo.AgeDeleteRule(daysToLive)))
                 .build()).name
     }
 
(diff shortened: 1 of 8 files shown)


bootstraponline:
> ftl.config.YamlConfigTest > assertGcsBucket FAILED

gcs requires a token which we can't use on a public job.  As a work around, we're using the `if bitrise return` check. 

- https://github.com/TestArmada/flank/blob/4e309c205103a0a13c50f12d0d08b145905f681e/cloud_testing/kotlin_poc/src/test/kotlin/ftl/android/AndroidCatalogTest.kt#L17

cloud_testing/kotlin_poc/src/main/kotlin/ftl/config/YamlConfig.kt
```diff
@@ -43,6 +47,23 @@ class YamlConfig(
         val directoriesToPull: List<String> = listOf()) {
 
 
+    private val storage = StorageOptions.newBuilder().setProjectId(projectId).build().service
+    private val bucketLabel: Map<String, String> = mapOf(Pair("flank", "apks"))
```
bootstraponline: I think the value can be an empty string instead of `apks`. If flank is used on an iOS app then apks doesn't make sense.


> mapOf(Pair("flank", ""))

cloud_testing/kotlin_poc/src/main/kotlin/ftl/config/YamlConfig.kt
```diff
@@ -43,6 +47,23 @@ class YamlConfig(
...
+    private val storageLocation = "us-central1"
+    private val daysToLive = 30
```
bootstraponline: by default we probably shouldn't set a delete rule. A user can add one manually if they want later. I think it'd be surprising if someone used flank and then realized the artifacts are auto deleted.

bootstraponline:
> ftl.MainTest > mockedTestRun FAILED
>    com.google.cloud.storage.StorageException: Unknown project id: 0

It looks like the mocked test broke.

cloud_testing/kotlin_poc/src/main/kotlin/ftl/config/YamlConfig.kt
```diff
@@ -52,6 +53,8 @@ class YamlConfig(
     private val storageLocation = "us-central1"
 
     fun getGcsBucket() : String {
+        if (bitrise) return rootGcsBucket
```
bootstraponline: we probably want something like `if (FtlConstants.useMock)` instead of `if (bitrise)`

bitrise should use the same logic as local runs

PR #236, opened 2018-06-21
Add end-to-end iOS test

### LINKED ISSUE #227: Kotlin | Figure out how to test the iOS runner
The iOS runner has no end to end tests using the mock server. The ticket should update `MainTest.kt` to use a new ios yaml and verify iOS works as expected.

vanwagonet:
While this does not make any assertions along the way, it ensures the process
completes without exception.
This also adds the mock server response for ios device catalog and downloads
temporary fixture build artifacts to tests against.

Fixes #227

### DIFF
--- a/cloud_testing/kotlin_poc/src/test/kotlin/ftl/TestArtifact.kt
+++ b/cloud_testing/kotlin_poc/src/test/kotlin/ftl/TestArtifact.kt
@@ -1,0 +1,93 @@
+package ftl
+
+import org.jsoup.Jsoup
+import org.jsoup.nodes.Document
+import org.jsoup.nodes.Element
+import java.io.File
+import java.util.concurrent.TimeUnit
+import kotlin.math.pow
+import xctest.Bash
+
+abstract class TestArtifact {
+    internal val fixturesPath = "./src/test/kotlin/ftl/fixtures/tmp"
+
+    init {
+        val assetLink = remoteAssetLink()
+        if (updateRequired(assetLink)) {
+            updateFixtures(assetLink)
+        }
+    }
+
+    private fun getHtml(): Document {
+        val githubUrl = "https://github.com/Flank/test_artifacts/releases/latest"
+        val maxRetries = 6
+        repeat(maxRetries) { attempt ->
+            try {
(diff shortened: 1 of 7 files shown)


cloud_testing/kotlin_poc/src/test/kotlin/ftl/TestArtifact.kt
```diff
@@ -0,0 +1,93 @@
...
+
+import org.jsoup.Jsoup
+import org.jsoup.nodes.Document
+import org.jsoup.nodes.Element
+import java.io.File
+import java.util.concurrent.TimeUnit
+import kotlin.math.pow
+import xctest.Bash
+
+abstract class TestArtifact {
```
bootstraponline: since this is almost a direct copy and paste from the other project, I think we'd want to refactor this to a common `test_fixture` project that's referenced by both `kotlin_poc` and `xctest_parser`

otherwise we'd have to maintain the code in two different places.

cloud_testing/mock_server/src/main/kotlin/ftl/Main.kt
```diff
@@ -35,6 +35,33 @@ object Main {
...
+                    iphone8.id = "iphone8"
+                    iphone8.name = "iPhone 8"
+                    iphone8.supportedVersionIds = listOf("11.2")
+
+                    val iphonex = IosModel()
+                    iphonex.id = "iphonex"
+                    iphonex.name = "iPhone X"
+                    iphonex.supportedVersionIds = listOf("11.2")
+
+                    val iosCatalog = IosDeviceCatalog()
```
bootstraponline: this is awesome 😄

bootstraponline:
The other idea I had is maybe we should combine xctest parser with kotlin_poc and have one project. That would be easier for us to work on, test, and distribute.

vanwagonet:
Should I combine them into one project in this PR, or should that be a step to take in a new PR?

bootstraponline:
> Should I combine them into one project in this PR, or should that be a step to take in a new PR?

I think a new PR to combine them & remove the code duplication makes sense. 

I merged this PR.

Issue #139, opened 2018-03-17
Group tests by time


bootstraponline:
Grouping tests by time instead of count results in faster runs. 

- Calculate time based on aggregated junit.xml results
- Update chunking logic to sort by time (largest tests first)

Reviewing the Flank Java implementation of this feature may be useful.

bootstraponline:
Fixed by https://github.com/TestArmada/flank/pull/385

PR #239, opened 2018-06-22
WIP | First attempt at grouping tests by time for issue #139

Blackchai:
This is my first stab at an algorithm for grouping the tests by their duration using the shard count provided by the user.

In this PR I am modifying the test method list to include duration time and then creating shard buckets based on the existing algorithm and comparing it to a time based approach, the algorithm for this was based on the Java implementation of Flank found here:
https://github.com/TestArmada/flank/blob/551818a0e0a8e054113aea5f98248ad436c9470f/Flank/src/main/java/com/walmart/otto/shards/ShardCreator.java

This version of the algorithm works by setting a bucket size for each shard equal to the longest test run in the list of tests. It then iterates through all tests in descending order attempting to fill each bucket with as many tests as possible. If all buckets are "full" and we have tests remaining, it starts again, with additional bucket space being equal to the remaining longest test.

Here is an example of the results:
WITH TIME BASED SORTING
Slowest shard = 191s
Total Time = 4575s // Time for all tests to be ran synchronously
WITHOUT TIME SORTING
Slowest shard = 329s
Total Time = 4575s // Time for all tests to be ran synchronously

Bear in mind that the algorithms performance is heavily based on the given shard count, number of tests, and their distribution of durations. This example was done with against 30 shards, with 70 tests and time range of 10-120 seconds.

I'd love to get some feedback on this approach, any C&C would be greatly appreciated!


bootstraponline:
It looks like the unit tests are failing.

bootstraponline:
Test scheduling algorithm looks good to me.

bootstraponline:
> This branch has conflicts that must be resolved

inktomi:
I think this is a fine first version of the algorithim. I worry a bit about the number of buckets that might be required - our project has around a thousand tests and given that the longest bucket will contain a single test (whatever test runs longest) we will probably end up with hundreds of buckets for the optimal case. 

Do you know if most projects have a handful / single of really long tests? That seems like an ideal case for this algorithm. Our suite plateaus pretty fast with no huge standout tests.

It would be really neat to be able to plug in algorithms. People could use the provided one, or configure their own based on their needs.

bootstraponline:
> It would be really neat to be able to plug in algorithms. People could use the provided one, or configure their own based on their needs.

Yeah, I think that makes sense. There are unique situations where a custom solution is going to be the best fit. 

> I worry a bit about the number of buckets that might be required

For clarity, the algorithm in this PR is using a user specified bucket maximum. `testShards` is how you control the maximum amount of buckets being created.

bootstraponline:
@inktomi does https://github.com/TestArmada/flank/issues/44 solve your use case?

inktomi:
Yea, it would! We currently use a custom test runner that works off a manifest like that. It's much easier to have Flank generate it, but this would give teams flexibility if needed.

bootstraponline:
Cool. I expect to resume work on Flank in August. I think reading in a list of tests may make sense once the other major changes have landed.

bootstraponline:
This feature is on pause until the new architecture work is complete.

- https://github.com/TestArmada/flank/issues/286

bootstraponline:
work will continue on this in a new PR.

PR #242, opened 2018-06-25
Update cost estimates by device form

### LINKED ISSUE #232: Kotlin | Print actual cost of the run
Now that Flank Kotlin supports multiple devices & platforms, the cost estimation is in need of updating. Using information from the device catalog we can accurately calculate the cost per run. Physical devices are billed at $5/hr and virtual devices are billed at $1/hr.

* Determine how many minutes is used per device type
  * iOS is always physical
* Calculate aggregate cost

2 hours on physical devices + 1 hour of virtual devices would be $11.

vanwagonet:
Calculate the estimated cost from each execution device based its form
(virtual or physical). This is also summing all devices instead of only looking
at the first.

fix #232

### DIFF
--- a/cloud_testing/kotlin_poc/src/main/kotlin/ftl/json/SavedMatrix.kt
+++ b/cloud_testing/kotlin_poc/src/main/kotlin/ftl/json/SavedMatrix.kt
@@ -50,8 +50,7 @@ class SavedMatrix(matrix: TestMatrix) {
         matrix.testExecutions.forEach {
             val step = GcToolResults.getResults(it.toolResultsStep)
             val billableMinutes = Billing.billableMinutes(step.testExecutionStep.testTiming.testProcessDuration.seconds)
-            val androidDevice = it.environment.androidDevice
-            if (androidDevice != null && AndroidCatalog.isVirtualDevice(androidDevice.androidModelId)) {
+            if (AndroidCatalog.isVirtualDevice(it.environment.androidDevice)) {
                 billableVirtualMinutes += billableMinutes
             } else {
                 billablePhysicalMinutes += billableMinutes
--- a/cloud_testing/kotlin_poc/src/main/kotlin/ftl/json/SavedMatrix.kt
+++ b/cloud_testing/kotlin_poc/src/main/kotlin/ftl/json/SavedMatrix.kt
@@ -1,6 +1,7 @@
 package ftl.json
 
 import com.google.api.services.testing.model.TestMatrix
+import ftl.android.AndroidCatalog
 import ftl.gc.GcToolResults
 import ftl.util.Billing
 import ftl.util.MatrixState.FINISHED
@@ -16,9 +17,8 @@ class SavedMatrix(matrix: TestMatrix) {
     var webLink: String = matrix.webLink()
     var downloaded = false
 
-    var billableMinutes: Long = 0
-    private var testDurationSeconds: Long = 0
-    private var runDurationSeconds: Long = 0
+    var billableVirtualMinutes: Long = 0
+    var billablePhysicalMinutes: Long = 0
     var outcome: String = ""
 
     init {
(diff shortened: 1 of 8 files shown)


bootstraponline:
Thanks, I'll review this tomorrow

cloud_testing/kotlin_poc/src/main/kotlin/ftl/json/SavedMatrix.kt
```diff
@@ -45,12 +45,17 @@ class SavedMatrix(matrix: TestMatrix) {
...
-            testDurationSeconds = step.testExecutionStep.testTiming.testProcessDuration.seconds
-            runDurationSeconds = step.runDuration.seconds
-            billableMinutes = Billing.billableMinutes(testDurationSeconds)
+        billableVirtualMinutes = 0
+        billablePhysicalMinutes = 0
+        matrix.testExecutions.forEach {
+            val step = GcToolResults.getResults(it.toolResultsStep)
+            val billableMinutes = Billing.billableMinutes(step.testExecutionStep.testTiming.testProcessDuration.seconds)
+            val androidDevice = it.environment.androidDevice
+            if (androidDevice != null && AndroidCatalog.isVirtualDevice(androidDevice.androidModelId)) {
```
bootstraponline: I think you could do something like this:

> AndroidCatalog.isVirtualDevice(androidDevice)

then the null check / androidModelId would move into the helper method.

cloud_testing/kotlin_poc/src/main/kotlin/ftl/util/Billing.kt
```diff
@@ -36,23 +36,39 @@ object Billing {
...
-        val virtualCost = billableMinutes.multiply(VIRTUAL_COST_PER_MIN).setScale(2, RoundingMode.HALF_UP)
+    private fun estimateCosts(billableVirtualMinutes: BigDecimal, billablePhysicalMinutes: BigDecimal): String {
+        val virtualCost = billableVirtualMinutes.multiply(VIRTUAL_COST_PER_MIN).setScale(2, RoundingMode.HALF_UP)
+        val physicalCost = billablePhysicalMinutes.multiply(PHYSICAL_COST_PER_MIN).setScale(2, RoundingMode.HALF_UP)
+        val totalCost = (virtualCost + physicalCost).setScale(2, RoundingMode.HALF_UP)
 
-        val remainder = billableMinutes.toLong()
-        val hours = TimeUnit.MINUTES.toHours(remainder)
-        val minutes = remainder % 60
+        val virtualTime = prettyTime(billableVirtualMinutes)
```
bootstraponline: I like the clean up into a prettyTime helper!

cloud_testing/kotlin_poc/src/test/kotlin/ftl/android/AndroidCatalogTest.kt
```diff
@@ -21,4 +21,13 @@ class AndroidCatalogTest {
         assertEquals(IncompatibleModelVersion, AndroidCatalog.supportedDeviceConfig("NexusLowRes", "21"))
         assertEquals(SupportedDeviceConfig, AndroidCatalog.supportedDeviceConfig("NexusLowRes", "23"))
     }
+
+    @Test
+    fun validateIsVirtualDevice() {
+        if (bitrise) return
```
bootstraponline: with mocking, are we able to use a fake AndroidCatalog on bitrise? I worry about too many `if (bitrise) return` means we're not running the tests in CI at all.

vanwagonet: the mock server handles the android catalog now, but I got some other credentials error when I tried to allow these tests to run.

vanwagonet: https://app.bitrise.io/build/9ab1863db270de36 was the build that errored.
`The Application Default Credentials are not available.`

bootstraponline:
```
 init {
        FtlConstants.useMock = true
    }
```

may fix that and activate the mock credential logic

PR #245, opened 2018-06-28
Add support for gs uris for APKs

### LINKED ISSUE #216: Kotlin | Allow gcs path to apks
appApk and testApk should accept a gcs path so we don't have to upload them every time.

```yaml
appApk: gs://apps/app-debug.apk
testApk: gs://apps/app-debug-androidTest.apk
```

StephenBrough:
Fixes #216

### DIFF
--- a/cloud_testing/apis/testing/build.gradle
+++ b/cloud_testing/apis/testing/build.gradle
@@ -12,5 +12,4 @@ repositories {
 dependencies {
     // https://mvnrepository.com/artifact/com.google.api-client/google-api-client
     compile 'com.google.api-client:google-api-client:1.23.0'
-    compile 'com.google.cloud:google-cloud-nio:0.30.0-alpha'
 }
--- a/cloud_testing/apis/testing/build.gradle
+++ b/cloud_testing/apis/testing/build.gradle
@@ -12,4 +12,5 @@ repositories {
 dependencies {
     // https://mvnrepository.com/artifact/com.google.api-client/google-api-client
     compile 'com.google.api-client:google-api-client:1.23.0'
+    compile 'com.google.cloud:google-cloud-nio:0.30.0-alpha'
 }
--- a/cloud_testing/kotlin_poc/src/main/kotlin/ftl/run/AndroidTestRunner.kt
+++ b/cloud_testing/kotlin_poc/src/main/kotlin/ftl/run/AndroidTestRunner.kt
@@ -1,6 +1,7 @@
 package ftl.run
 
 import com.google.api.services.testing.model.TestMatrix
+import ftl.config.FtlConstants
 import ftl.config.YamlConfig
 import ftl.gc.GcAndroidMatrix
 import ftl.gc.GcAndroidTestMatrix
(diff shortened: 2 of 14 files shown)


cloud_testing/apis/testing/build.gradle
```diff
@@ -12,4 +12,5 @@ repositories {
 dependencies {
     // https://mvnrepository.com/artifact/com.google.api-client/google-api-client
     compile 'com.google.api-client:google-api-client:1.23.0'
+    compile 'com.google.cloud:google-cloud-nio:0.30.0-alpha'
```
bootstraponline: is this required to compile the API project?

cloud_testing/kotlin_poc/src/main/kotlin/ftl/run/AndroidTestRunner.kt
```diff
@@ -53,4 +46,33 @@ object AndroidTestRunner : GenericTestRunner {
 
         return afterRunTests(jobs, runGcsPath, stopwatch, config)
     }
+
+    /**
+     * If given a Gcs URI, then download the APKs
```
bootstraponline: Is this comment still valid? `then download the APKs` I see the upload logic however I'm not sure where the download code is in `resolveApks`

cloud_testing/kotlin_poc/src/test/kotlin/ftl/android/LocalGcs.kt
```diff
@@ -0,0 +1,42 @@
...
+
+        Assert.assertTrue(
+                GcStorage.storage.list(TEST_BUCKET).values
+                        .map { it.name }
+                        .any { TEST_BLOB_PATH == it }
+        )
+
+        val appApkInputStream = FileInputStream(APP_APK_PATH)
+        val appApkBlobInfo = BlobInfo.newBuilder(TEST_BUCKET, APP_BLOB_PATH).build()
+        GcStorage.storage.create(appApkBlobInfo, appApkInputStream)
```
bootstraponline: Thanks for figuring out how to the the Google cloud storage API!

cloud_testing/kotlin_poc/src/test/kotlin/ftl/fixtures/flank.yml
```diff
@@ -1,5 +1,5 @@
 appApk: ../../test_app/apks/app-debug.apk
-testApk: ../../test_app/apks/app-debug-androidTest.apk
+testApk: gs://tmp_bucket_2/app-debug-androidTest.apk
```
bootstraponline: I think instead of replacing the local test apk, we should add a new fixture. This way both code paths are tested.

flank.local.yml
flank.gcs.yml

I think something like that would work ^

cloud_testing/apis/testing/build.gradle
bootstraponline: I think that may belong in the Kotlin build.gradle file instead:

[cloud_testing/kotlin_poc/build.gradle.kts#L64](https://github.com/TestArmada/flank/blob/88a4f8e5c61b25a8a795114ad295e59ef1803158/cloud_testing/kotlin_poc/build.gradle.kts#L64)

bootstraponline:
> ftl.android.AndroidCatalogTest > validateDeviceConfig FAILED
    java.lang.RuntimeException: java.io.IOException: The Application Default Credentials are not available. They are available if running on Google App Engine, Google Compute Engine, or Google Cloud Shell. Otherwise, the environment variable GOOGLE_APPLICATION_CREDENTIALS must be defined pointing to a file defining the credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.


I think the failure is due to not setting mocked mode. We probably need a custom JUnit runner to always set mocked mode.

https://github.com/TestArmada/flank/blob/6f7de2600d966949677d0d741df0d748f996f690/cloud_testing/kotlin_poc/src/main/kotlin/ftl/config/FtlConstants.kt#L31

You can see here that a mock credential is used if we're in mocked mode.

StephenBrough:
Ah, that's what that was, I'll add that in real quick

StephenBrough:
Looks like we need to use a mock server for catalog tests in Android, but not iOS. Odd that the tests for Android don't pass locally.

bootstraponline:
that's weird, the mock catalog should be used for both.

PR #251, opened 2018-07-05
Refactor yaml fields to match gcloud cli commands

### LINKED ISSUE #238: Kotlin | YAML compatibility with gcloud CLI
**Blocked on: #224**

Flank YAML should be updated to be compatible with gcloud. Test by creating a YAML file and running it with both `gcloud` CLI and `flank`

I think we're using different naming conventions in a few places.

https://cloud.google.com/sdk/gcloud/reference/topic/arg-files

- [ ] TODO: Refactor away from generics and type casting

StephenBrough:
#238


bootstraponline:
```
$ cd cloud_testing/kotlin_poc
$ gcloud firebase test android run flank.yml:flank

Have questions, feedback, or issues? Get support by visiting:
  https://firebase.google.com/support/

ERROR: (gcloud.firebase.test.android.run) [testShards] is not a valid argument name for: gcloud test run.
```

I think we need two top level keys:

- `gcloud` for the gcloud params
- `flank` for the flank specific params

cloud_testing/kotlin_poc/flank.ios.yml
```diff
@@ -1,10 +1,11 @@
...
-testShards: 2
-limitBreak: false
-waitForResults: true
+flank:
+  test: ./src/test/kotlin/ftl/fixtures/tmp/EarlGreyExample.zip
+  xctestrun-file: ./src/test/kotlin/ftl/fixtures/tmp/EarlGreyExampleSwiftTests_iphoneos11.2-arm64.xctestrun
+  results-bucket: earlgrey-swift
+  performance-metrics: false
+  record-video: true
+  timeout: 30
```
bootstraponline: timeout logic requires additional parsing:

```
--timeout 1h is 1 hour
--timeout 5m is 5 minutes
--timeout 200s is 200 seconds
--timeout 100 is 100 seconds
```

The current change takes us from a 60m timeout to a 60 second timeout.

cloud_testing/kotlin_poc/src/main/kotlin/ftl/gc/GcAndroidTestMatrix.kt
```diff
@@ -70,7 +70,7 @@ object GcAndroidTestMatrix {
         testMatrix.testSpecification = TestSpecification()
                 .setAndroidInstrumentationTest(androidInstrumentation)
                 .setDisablePerformanceMetrics(config.disablePerformanceMetrics)
```
bootstraponline: The negatives mean this does the opposite of what's intended.

`performance-metrics` set to true means we want to enable performance metrics. `performance-metrics` is stored as `disablePerformanceMetrics` which is passed to `setDisablePerformanceMetrics`. Enabling performance metrics disables them. 🙂 

we should probably do something like this: `.setDisablePerformanceMetrics(!config.performanceMetrics)`

cloud_testing/kotlin_poc/src/main/kotlin/ftl/gc/GcAndroidTestMatrix.kt
```diff
@@ -70,7 +70,7 @@ object GcAndroidTestMatrix {
...
-                .setDisableVideoRecording(config.disableVideoRecording)
+                .setDisableVideoRecording(config.recordVideo)
```
bootstraponline: I think this has the same logic error as `disablePerformanceMetrics`

cloud_testing/kotlin_poc/src/main/kotlin/ftl/config/AndroidConfig.kt
```diff
@@ -1,18 +1,25 @@
...
         val autoGoogleLogin: Boolean = true,
+        @field:JsonProperty("use-orchestrator")
         val useOrchestrator: Boolean = true,
+        @field:JsonProperty("environment-variables")
         val environmentVariables: Map<String, String> = mapOf(),
+        @field:JsonProperty("directories-to-pull")
         val directoriesToPull: List<String> = listOf(),
-
         rootGcsBucket: String,
         disablePerformanceMetrics: Boolean = true,
```
bootstraponline: I think the rest of these fields are missing the JsonProperty annotation?

cloud_testing/kotlin_poc/src/main/kotlin/ftl/config/YamlConfig.kt
```diff
@@ -21,16 +23,24 @@ import java.net.URI
...
 open class YamlConfig(
+        @field:JsonProperty("results-bucket")
         val rootGcsBucket: String,
+        @field:JsonProperty("performance-metrics")
         val disablePerformanceMetrics: Boolean = true,
-        val disableVideoRecording: Boolean = false,
+        @field:JsonProperty("record-video")
+        val recordVideo: Boolean = false,
+        @field:JsonProperty("timeout")
         val testTimeoutMinutes: Long = 60,
```
bootstraponline: this is now `testTimeout` and a String type.

cloud_testing/kotlin_poc/src/main/kotlin/ftl/config/AndroidConfig.kt
```diff
@@ -1,18 +1,25 @@
...
 
 class AndroidConfig(
+        @field:JsonProperty("app")
         val appApk: String = "",
+        @field:JsonProperty("test")
         val testApk: String = "",
+        @field:JsonProperty("auto-google-login")
         val autoGoogleLogin: Boolean = true,
+        @field:JsonProperty("use-orchestrator")
         val useOrchestrator: Boolean = true,
```
Macarse: this means that `useOrchestrator` is on by default?

bootstraponline: > this means that useOrchestrator is on by default?

Yes, this is the recommended way to run espresso tests with Flank.

cloud_testing/kotlin_poc/src/main/kotlin/ftl/gc/GcAndroidTestMatrix.kt
StephenBrough: Ah, now I get why they were named them that way, thought it was kind of counterintuitive, haha

cloud_testing/kotlin_poc/src/main/kotlin/ftl/config/AndroidConfig.kt
StephenBrough: Looks like the annotations in the base class cover these ones

bootstraponline: Great! maybe add a comment since that isn't obvious.

StephenBrough: Good idea, I'll add one in

bootstraponline:
Thanks, I'll test/review this again tomorrow.

PR #320, opened 2018-09-19
JUnit XML

### LINKED ISSUE #314: Always emit aggregated junit.xml report
For integration with CI systems it's nice to have a report of all the tests that ran, and whether or not they passed. Currently, Flank only returns reports for tests that fail. 

Provide a flag that will result in a report of all the tests that ran & their status in a JUnit xml format.

bootstraponline:
Fix #314

To Do
- [x] Parse iOS and Android JUnit XML
- [x] Generate reports from JUnit XML
- [x] Always output aggregated JUnit XML report


docs/junit_xml_comparison.md
```diff
@@ -0,0 +1,70 @@
...
+      failures
+      errors
+      skipped* Android only.
+      time
+      timestamp* Android only.
+      hostname - always localhost
+    testcase
+      name
+      classname
+      time
```
inktomi: Just curious if this is also the same part of code that we can get the per-test-case timings from? One of these days I'll get to work on the auto-bucketting..

PR #321, opened 2018-09-19
Support running by (not) package, class, annotation, testfile

### LINKED ISSUE #175: Support running by package & class
com.example.MyClass should be a valid test target. Currently the test method detection will flag it as a method and prevent execution.

image

I think something like this YAML may allow us to exactly match the gcloud CLI. Note that the validation logic on Android will need to be reworked to take into account `package/class/test` method. iOS validation is only able to support `method_name` right now

```
testTargets:
 - package package_name
 - class package_name.class_name
 - class package_name.class_name#method_name
```

winterDroid:
This fixes #175. It is based on the implementation from 2.x.
It supports:
```
testTargets:
- package package_name
- class package_name.class_name
- class package_name.class_name1#method_name
- annotation annotation_name
- testfile test_file_path
```
`@Ignored` annotated tests are always filtered.
Supports ignoring tests by `not` e.g. notClass.


bootstraponline:
Thanks, I'll review this soon.

test_runner/src/main/kotlin/ftl/filter/TestFilters.kt
```diff
@@ -0,0 +1,89 @@
...
+import java.util.Locale
+import java.util.regex.Pattern
+
+typealias TestFilter = (TestMethod) -> Boolean
+
+object TestFilters {
+    private const val ARG_CLASS = "class"
+    private const val ARG_PACKAGE = "package"
+    private const val ARG_ANNOTATION = "annotation"
+    private const val ARG_TEST_FILE = "testfile"
```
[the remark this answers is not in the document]
winterDroid: I think then it is better to ignore `testfile` for now. How do you think it would be possible without client side validation since we need to know exactly which methods to run?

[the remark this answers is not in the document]
winterDroid: In the documentation it is mentioned that 
> Any test filter supported by `am instrument -e ...` is supported.

According to the [AndroidJUnitRunner JavaDoc](https://developer.android.com/reference/android/support/test/runner/AndroidJUnitRunner) that would mean:
```
public static final String ARGUMENT_TEST_CLASS = "class";
private static final String ARGUMENT_TEST_SIZE = "size";
private static final String ARGUMENT_ANNOTATION = "annotation";
private static final String ARGUMENT_NOT_ANNOTATION = "notAnnotation";
private static final String ARGUMENT_TEST_PACKAGE = "package";
private static final String ARGUMENT_TEST_FILE = "testFile";
```

rashiq:
if you'd drop the validation then how would the `testShards` flag work? would it shard based on test classes?

bootstraponline:
In looking through https://developer.android.com/reference/android/support/test/runner/AndroidJUnitRunner there's a lot of new functionality since I originally wrote the ticket.

I think the approach you have is the correct one. I didn't realize for example that `testFile` is a supported feature now.

test_runner/src/main/kotlin/ftl/filter/TestFilters.kt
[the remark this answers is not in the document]
bootstraponline: I think that's a great idea. We'd probably want to make sure we're using the latest alpha code. Last time I checked, that source was not up to date with the AndroidX espresso.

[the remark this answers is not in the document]
bootstraponline: re: testFile, that's my mistake. You're correct this is a supported feature. testFile under test-targets makes sense.

bootstraponline:
after talking on Slack the plan is to:

- keep the validation & try to reuse the `AndroidX Test Runner 1.1.0-alpha4` dependency to provide filtering support
- If upstream code can't be reused then the approach in this PR of emulating the logic in Kotlin makes sense.

bootstraponline:
I found a list of supported instrumentation arguments in `RunnerArgs.java` from [runner-1.1.0-alpha4-sources.jar](https://dl.google.com/dl/android/maven2/androidx/test/runner/1.1.0-alpha4/runner-1.1.0-alpha4-sources.jar). Unfortunately there are Android os dependencies so we can't use the upstream code.

Constant | String | Flank support
-- | -- | --
 ARGUMENT_TEST_CLASS | class | :white_check_mark:
 ARGUMENT_CLASSPATH_TO_SCAN | classpathToScan | :x:
 ARGUMENT_NOT_TEST_CLASS | notClass | :white_check_mark:
 ARGUMENT_TEST_SIZE | size | :white_check_mark:
 ARGUMENT_LOG_ONLY | log | :x:
 ARGUMENT_ANNOTATION | annotation | :white_check_mark:
 ARGUMENT_NOT_ANNOTATION | notAnnotation | :white_check_mark:
 ARGUMENT_NUM_SHARDS | numShards | :x:
 ARGUMENT_SHARD_INDEX | shardIndex | :x:
 ARGUMENT_DELAY_IN_MILLIS | delay_msec | :x:
 ARGUMENT_COVERAGE | coverage | :x:
 ARGUMENT_COVERAGE_PATH | coverageFile | :x:
 ARGUMENT_SUITE_ASSIGNMENT | suiteAssignment | :x:
 ARGUMENT_DEBUG | debug | :x:
 ARGUMENT_LISTENER | listener | :x:
 ARGUMENT_FILTER | filter | :x:
 ARGUMENT_RUNNER_BUILDER | runnerBuilder | :x:
 ARGUMENT_TEST_PACKAGE | package | :white_check_mark:
 ARGUMENT_NOT_TEST_PACKAGE | notPackage | :white_check_mark:
 ARGUMENT_TIMEOUT | timeout_msec | :x:
 ARGUMENT_TEST_FILE | testFile | :white_check_mark:
 ARGUMENT_NOT_TEST_FILE | notTestFile | :white_check_mark:
 ARGUMENT_DISABLE_ANALYTICS | disableAnalytics | :x:
 ARGUMENT_APP_LISTENER | appListener | :x:
 ARGUMENT_CLASS_LOADER | classLoader | :x:
 ARGUMENT_REMOTE_INIT_METHOD | remoteMethod | :x:
 ARGUMENT_TARGET_PROCESS | targetProcess | :x:
 ARGUMENT_SCREENSHOT_PROCESSORS | screenCaptureProcessors | :x:
 ARGUMENT_ORCHESTRATOR_SERVICE | orchestratorService | :x:
 ARGUMENT_LIST_TESTS_FOR_ORCHESTRATOR | listTestsForOrchestrator | :x:
 ARGUMENT_SHELL_EXEC_BINDER_KEY | shellExecBinderKey | :x:
 ARGUMENT_RUN_LISTENER_NEW_ORDER | newRunListenerMode | :x:

README.md
```diff
@@ -78,6 +78,8 @@ gcloud:
     - /sdcard/screenshots
   performance-metrics: true
   test-targets:
+    # supported are class, notClass, package, notPackage, annotation, notAnnotation, testfile, notTestfile, size
+    # as described in https://developer.android.com/reference/android/support/test/runner/AndroidJUnitRunner
```
winterDroid: I was wondering whether it would be nicer to actually support the available options directly through yaml like:
```
test-targets:
  class: ...
  notClass: ...
  package: ...
  notPackage: ...
  annotation: ...
  notAnnotation: ...
  testfile: ...
  notTestfile: ...
  size: ...
```
and then either pass the params as a list or just a comma separated string like:
```
test-targets:
  package: 
  - com.foo.bar
  size: 
  - large
```

bootstraponline: @winterDroid we're trying to match test-targets behavior of gcloud CLI. If they support that, great. If not then it's easier to maintain & debug flank if we match the same behavior.

winterDroid: That makes sense. Then I guess we should keep it as is

bootstraponline:
@winterDroid I made some changes, let me know what you think. If you approve then the feature is ready to merge.

bootstraponline:
I added a failing test:

```kotlin
    @Test
    fun multipleTestsFilter() {
        val m1 = TestMethod("a.b#c", listOf(TestAnnotation("Ignore", emptyMap())))
        val m2 = TestMethod("d.e#f", listOf(TestAnnotation("Foo", emptyMap())))
        val m3 = TestMethod("h.i#j", listOf(TestAnnotation("Bar", emptyMap())))

        val testMethods = listOf(m1, m2, m3)
        val testFilter = fromTestTargets(listOf("notAnnotation Foo", "class d.e#f"))

        val list = testMethods
            .asSequence()
            .filter(testFilter::invoke)
            .map(TestMethod::testName)
            .map { "class $it" }
            .toList()

//        expected: [class h.i#j]
//        but was : []
        assertThat(list).isEqualTo(listOf("class h.i#j"))
    }
```

It looks like the filter behavior is buggy when there are conflicting selectors.

bootstraponline:
I pushed an experimental commit to improve debugging by using named methods. Now it's possible to detect which filter is rejecting a test. The filtering logic itself is still failing to pass the `multipleTestsfilter` test.

image

bootstraponline:
I updated the parsing logic. I think the issues are fixed now.

>  allOf(notIgnored(), anyOf(*testMethod, allOf(*otherFilters)))

Let me know if you have any feedback on the approach.

test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
```diff
@@ -83,18 +85,30 @@ class AndroidArgs(
...
         testShardChunks = calculateShards(
             testTargets,
-            validTestMethods,
+            filteredTests,
             testTargetsAlwaysRun,
             testShards
         )
     }
 
+    private fun getTestMethods(testLocalApk: String): List<String> {
```
winterDroid: Do you really want to fail it with an error? Shouldn't a warning be enough?

test_runner/src/main/kotlin/ftl/filter/TestFilters.kt
```diff
@@ -0,0 +1,181 @@
+package ftl.filter
+
+import com.linkedin.dex.parser.TestMethod
+import java.io.IOException
+import java.nio.file.Files
+import java.nio.file.Paths
+import java.util.regex.Pattern
+
+interface ITestFilter {
```
winterDroid: Maybe you want to match more the JUnit test filter? Meaning a method `shouldRun` and a `description`. It would be also possible to have a class that takes the description and a lambda.

test_runner/src/main/kotlin/ftl/filter/TestFilters.kt
```diff
@@ -0,0 +1,181 @@
...
+        } catch (e: IOException) {
+            throw RuntimeException("Unable to read testFile", e)
+        }
+    }
+
+    private fun withPackageName(packageNames: List<String>): ITestFilter = object : ITestFilter {
+        override val name = "withPackageName ${packageNames.joinToString(", ")}"
+        override fun run(testMethod: TestMethod): Boolean {
+            require(packageNames.isNotEmpty()) { MISSING_ARGUMENTS_MSG }
+            return packageNames.any { packageName -> testMethod.testName.startsWith(packageName) }
```
winterDroid: Either all errors should be extracted as a constant or none. Do you actually think the error message here makes still sense?

test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
bootstraponline: I thought about warning. For example in the yaml config, if you have an incorrect param that'll produce a warning via the doctor command.

The motivation behind using Flank is for test sharding. Test sharding requires accurate information about test methods. If we're not able to get the test methods, then we can't shard the tests. That seems like a fatal error to me where we'd want to exit immediately and alert the user that something is wrong.

If it was only a warning then the user ends up getting billed for tests that may not have executed as intended.

test_runner/src/main/kotlin/ftl/filter/TestFilters.kt
bootstraponline: > Maybe you want to match more the JUnit test filter?

These aren't JUnit filters so I'm not sure it buys us much to match that API.
 
>  It would be also possible to have a class that takes the description and a lambda.

I'm interested in that idea. Do you think a class with a description/lambda is cleaner than the current interface and anonymous class approach?

If you want to try that refactor I'll take a look.

test_runner/src/main/kotlin/ftl/filter/TestFilters.kt
bootstraponline: I inlined the message and added a few more error handling statements.

I think it makes sense to have sanity checks in the filters, that helps reduce the risk of an invalid test run.

test_runner/src/main/kotlin/ftl/filter/TestFilters.kt
bootstraponline: > It would be also possible to have a class that takes the description and a lambda.

I pushed a new commit with that refactor.

bootstraponline: Updated to describe/shouldRun to match https://junit.org/junit4/javadoc/4.12/org/junit/runner/manipulation/Filter.html

PR #327, opened 2018-09-25
Make flank run with multiple test classes

rashiq:
Right now you can run one test class by specifying

test-targets:
	- class Foo.Bar

But that doesn't work for multiple test classes.

### DIFF
--- a/test_runner/src/main/kotlin/ftl/filter/TestFilters.kt
+++ b/test_runner/src/main/kotlin/ftl/filter/TestFilters.kt
@@ -18,7 +18,7 @@ data class TestFilter(
 )
 
 private fun TestFilter.isTestMethod(): Boolean {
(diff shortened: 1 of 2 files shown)


rashiq:
Doing this however, it still doesn't shard the test methods; so test methods from `class Bar` are run in one matrix, and test methods from `class Foo` are run in one matrix

bootstraponline:
```
ftl.filter.TestFiltersTest > testFilteringByClassName FAILED
    expected to be false
        at ftl.filter.TestFiltersTest.testFilteringByClassName(TestFiltersTest.kt:51)
```

looks like this broke another part of filtering.

I'll take a look at the code. In theory we should be able to resolve classes down to methods since we have a full list of tests from dex test parse.

bootstraponline:
fixed in https://github.com/TestArmada/flank/pull/328

PR #345, opened 2018-10-10
Add flank cancel command to cancel unfinished matrices from the last run

### LINKED ISSUE #194: Cancel builds on ctrl+c
When cancelling the Kotlin process, it'd be nice if we also cancelled the running matrices. Additionally a `flank cancel` command could be added to cancel matrix ids in the most recent flank.json. That's useful for when async mode is used.

Cancel should work for both iOS and Android runs.

vtsoup:
Fix #194 

`flank cancel`

The matrices from the last run are processed. Any matrix with state != FINISHED will be cancelled asynchronously.

### DIFF
--- a/test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
+++ b/test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
@@ -33,10 +33,10 @@ import java.nio.file.Path
 class AndroidArgs(
     gcloudYml: GcloudYml,
     androidGcloudYml: AndroidGcloudYml,
-    flankYml: FlankYml
+    flankYml: FlankYml,
+    override val data: String
 ) : IArgs {
     private val gcloud = gcloudYml.gcloud
-    override var filePath: Path? = null
     override val resultsBucket: String
     override val recordVideo = gcloud.recordVideo
     override val testTimeout = gcloud.timeout
@@ -159,11 +159,7 @@ ${listToString(testTargetsAlwaysRun)}
             mergeYmlMaps(GcloudYml, AndroidGcloudYml, FlankYml)
         }
 
-        fun load(data: Path): AndroidArgs {
-            val rtn = load(String(Files.readAllBytes(data)))
-            rtn.filePath = data
-            return rtn
-        }
+        fun load(data: Path): AndroidArgs = load(String(Files.readAllBytes(data)))
 
         fun load(data: String): AndroidArgs {
             val flankYml = yamlMapper.readValue(data, FlankYml::class.java)
@@ -173,7 +169,8 @@ ${listToString(testTargetsAlwaysRun)}
             return AndroidArgs(
                 gcloudYml,
                 androidGcloudYml,
-                flankYml
+                flankYml,
+                data
             )
         }
     }
--- a/test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
+++ b/test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
@@ -36,6 +36,7 @@ class AndroidArgs(
     flankYml: FlankYml
 ) : IArgs {
     private val gcloud = gcloudYml.gcloud
+    override var filePath: Path? = null
     override val resultsBucket: String
     override val recordVideo = gcloud.recordVideo
     override val testTimeout = gcloud.timeout
@@ -158,7 +159,11 @@ ${listToString(testTargetsAlwaysRun)}
             mergeYmlMaps(GcloudYml, AndroidGcloudYml, FlankYml)
         }
 
-        fun load(data: Path): AndroidArgs = load(String(Files.readAllBytes(data)))
+        fun load(data: Path): AndroidArgs {
+            val rtn = load(String(Files.readAllBytes(data)))
+            rtn.filePath = data
+            return rtn
+        }
 
         fun load(data: String): AndroidArgs {
             val flankYml = yamlMapper.readValue(data, FlankYml::class.java)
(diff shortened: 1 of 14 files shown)


bootstraponline:
It looks like the unit test suite is failing.

test_runner/src/main/kotlin/ftl/cli/firebase/test/ios/IosCancelCommand.kt
```diff
@@ -0,0 +1,37 @@
...
+        synopsisHeading = "%n",
+        descriptionHeading = "%n@|bold,underline Description:|@%n%n",
+        parameterListHeading = "%n@|bold,underline Parameters:|@%n",
+        optionListHeading = "%n@|bold,underline Options:|@%n",
+        header = ["Cancels the last iOS Firebase Test Lab run"],
+        description = ["""Selects the most recent run in the results/ folder.
+Reads in the matrix_ids.json file. Cancels any incomplete matrices.
+"""]
+)
+class IosCancelCommand : Runnable {
```
bootstraponline: The cancel logic should be the same for both Ios and Android, I think we could have one command for both.

Are you sure we need to load in the IosArgs? The folders are named using timestamps so we should be able to pick the latest folder and then read the matrix_ids.json without needing a flank.yml

test_runner/src/main/kotlin/ftl/gc/GcTestMatrix.kt
```diff
@@ -48,4 +49,21 @@ object GcTestMatrix {
 
         throw RuntimeException("Failed to refresh matrix")
     }
+
+    fun cancel(testMatrixId: String, config: IArgs): CancelTestMatrixResponse {
+        val cancelMatrix = GcTesting.get.projects().testMatrices().cancel(config.projectId, testMatrixId)
```
bootstraponline: if we need the projectId, it may make sense to add that in the `matrix_id.json` serialization, that'll decouple the cancel logic from the configuration.

test_runner/src/main/kotlin/ftl/run/TestRunner.kt
```diff
@@ -278,6 +308,18 @@ object TestRunner {
         ReportManager.generate(matrixMap)
     }
 
+    // used to cancel and update results from an async run
+    suspend fun cancelLastRun(config: IArgs) {
+        val matrixMap = lastMatrices()
+
+        cancelMatrices(matrixMap, config)
+        pollMatrices(matrixMap, config)
```
bootstraponline: if we're cancelling then I don't think we need to poll, fetch artifacts, or generate reports.

test_runner/src/main/kotlin/ftl/cli/firebase/test/ios/IosCancelCommand.kt
vtsoup: That is an idea. I basically copied the refresh command and its logic flow. I will look into creating one cancel command.

bootstraponline:
for ctrl+c, we can add a shutdown hook:

https://stackoverflow.com/questions/1611931/catching-ctrlc-in-java

bootstraponline:
new idea: vendor the config used to schedule the run into the corresponding results/ folder

bootstraponline:
with the vendored configs, we'd want to keep the two different commands (one for each platform)

vtsoup:
So I am not sure that we should support cancelling via ctrl+c. I can see a scenario where we have long running process, and we need to close the terminal for whatever reason. Or if I accidentally do a ctrl+c out of habit. Unless we prompt the user to confirm before starting a cancel, I would think that having the user explicitly issue a separate `flank cancel` command is a safer way to cancel. Thoughts?

bootstraponline:
> I would think that having the user explicitly issue a separate `flank cancel` command is a safer way to cancel. Thoughts?

yeah that's a good point, I agree with you.

test_runner/src/main/kotlin/ftl/cli/firebase/CancelCommand.kt
```diff
@@ -0,0 +1,29 @@
...
+        descriptionHeading = "%n@|bold,underline Description:|@%n%n",
+        parameterListHeading = "%n@|bold,underline Parameters:|@%n",
+        optionListHeading = "%n@|bold,underline Options:|@%n",
+        header = ["Cancels the last Firebase Test Lab run"],
+        description = ["""Selects the most recent run in the results/ folder.
+Reads in the matrix_ids.json file. Cancels any incomplete matrices.
+"""]
+)
+class CancelCommand : Runnable {
+    override fun run() {
```
vtsoup: I consolidated the cancel to just one command because calling it will just cancel the matrices in last run folder. If we want to specify whether we want to cancel an iOS or Android last run, we will need to structure our results folder differently, like have `results-ios` and `results-android` folders or prefix our run folders with `ios-` or `android-`.

bootstraponline:
image

test failed on bitrise

test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
```diff
@@ -158,7 +159,11 @@ ${listToString(testTargetsAlwaysRun)}
             mergeYmlMaps(GcloudYml, AndroidGcloudYml, FlankYml)
         }
 
-        fun load(data: Path): AndroidArgs = load(String(Files.readAllBytes(data)))
+        fun load(data: Path): AndroidArgs {
```
bootstraponline: ?

test_runner/src/main/kotlin/ftl/cli/firebase/CancelCommand.kt
```diff
@@ -0,0 +1,29 @@
...
+        runBlocking {
+            TestRunner.cancelLastRun()
+        }
+    }
+
+    @CommandLine.Option(names = ["-h", "--help"], usageHelp = true, description = ["Prints this help message"])
+    var usageHelpRequested: Boolean = false
+}
```
bootstraponline: missing newline

PR #351, opened 2018-10-11
Modified refresh command

### LINKED ISSUE #349: Modify refresh to read from last run config and to also continually poll
Right now, there is a separate `flank android refresh` and `flank ios refresh`. Both commands have a config as command line option. We should consolidate them to a single refresh. It would be similar to `flank cancel` where the config file is read from the last run run folder. This means we will also remove the --config option from refresh.

Also, refresh is only getting the current state. We should add an option to poll so that the matrices state is continually updated: `flank refresh --poll`.

vtsoup:
Fix #349.

- Consolidated refresh commands
- Use the config saved in the last run folder
- Added polling to occur after current state is refreshed


bootstraponline:
looks like the CI failed

PR #354, opened 2018-10-12
Lazy load the discovery of test methods in args files. 

### LINKED ISSUE #350: Lazy load test methods IosArgs and AndroidArgs
Sometimes we don't need the test methods (cancel/refresh). It'd be better to calculate these on demand instead of always on init.

vtsoup:
Fix #350


test_runner/src/test/kotlin/ftl/args/AndroidArgsFileTest.kt
```diff
@@ -108,11 +108,11 @@ class AndroidArgsFileTest {
...
-    @Test
-    fun calculateShards_0() {
-        exceptionRule.expectMessage("Test APK has no tests")
-        configWithTestMethods(0)
-    }
+//    @Test
+//    fun calculateShards_0() {
+//        exceptionRule.expectMessage("Test APK has no tests")
+//        configWithTestMethods(0)
+//    }
```
vtsoup: This is the failing test, at least locally.

vtsoup: Fixed. I just referenced the lazy property from the test.

PR #376, opened 2018-11-02
Fetch artifacts in parallel. Treat warnings as errors

### LINKED ISSUE #323: Fetch Artifacts in parallel
Can we use e.g. coroutines to fetch all the artifacts in the test runner in parallel?

bootstraponline:
Fix #323

### DIFF
--- a/test_runner/src/main/kotlin/ftl/run/TestRunner.kt
+++ b/test_runner/src/main/kotlin/ftl/run/TestRunner.kt
@@ -208,28 +208,29 @@ object TestRunner {
             finished && notDownloaded
         }
 
-        // TODO: fetch in parallel https://github.com/TestArmada/flank/issues/323
         print(indent)
-        filtered.forEach { matrix ->
-            val prefix = Storage.BlobListOption.prefix(matrix.gcsPathWithoutRootBucket)
-            val result = GcStorage.storage.list(matrix.gcsRootBucket, prefix, fields)
+        runBlocking {
+            filtered.forEach { matrix ->
+                launch {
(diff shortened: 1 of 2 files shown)


test_runner/src/main/kotlin/ftl/run/TestRunner.kt
```diff
@@ -208,28 +208,29 @@ object TestRunner {
...
-                if (
-                    blobPath.matches(ArtifactRegex.testResultRgx)
-                ) {
-                    val downloadFile = Paths.get(FtlConstants.localResultsDir, blobPath)
-                    print(".")
-                    if (!downloadFile.toFile().exists()) {
-                        downloadFile.parent.toFile().mkdirs()
-                        blob.downloadTo(downloadFile)
+        runBlocking {
+            filtered.forEach { matrix ->
```
winterDroid: should the amount of parallel executions be limited?

bootstraponline: with the current code, there's one parallel execution per matrix that's finished & not yet downloaded. 

I don't think this will be a problem. The download requests per matrix are still run synchronously. 

if it is an issue then we could look into placing an upper limit on the amount of concurrency.

bootstraponline: https://github.com/Kotlin/kotlinx.coroutines/blob/d6a5a399d1724ff56bbb285b25df071dbc98b715/core/kotlinx-coroutines-core/src/CommonPool.kt#L23

 > If the property is not specified, `Runtime.getRuntime().availableProcessors() - 1` will be used 
>
> https://github.com/Kotlin/kotlinx.coroutines/issues/261

based on the above, looks like we're good

PR #377, opened 2018-11-02
Add a parameter to shuffle the tests before sharding

doodla:
A use case we have is we have a lot of tests which have arbitrary sleeps ( probably something we should work to eliminate ). These take a longer time to complete than our other tests. As it so happens, all of these are in the same class. So when sharding, most of these end up in the same shard.

Shuffling will ensure some variance in where these tests end up, and will ensure faster CI.


doodla:
Discussed via Slack. With #139 coming in a couple of weeks, it doesn't make sense to have this, even temporarily.

PR #380, opened 2018-11-06
Modify the WebLink to point to the matrix root if there are more than one device

doodla:
Previously, if a matrix had multiple devices, the link would take you into the details of the first device.

Expected:
image
Actual:
image

Now, if there's a single device, it will take you directly to the execution. But if there are multiple devices, it will take you to the root matrix page which lists all the devices.

So, for **ONE** device, you'll still see the execution as before ( second screenshot above ). But for **MULTIPLE** devices, you'll see the matrix base screen ( first screenshot above ).


bootstraponline:
Great idea! Merged.

PR #385, opened 2018-11-09
Smart Flank

bootstraponline:
After each test run, the JUnit XML file is uploaded to Google storage

## To Do

- [x] Download JUnit XML
- [x] Merge JUnit XML (retain only passing tests with timing info)
- [x] Upload JUnit XML
- [x] Shard tests based on JUnit XML timing (see https://github.com/TestArmada/flank/pull/239)
- [x] Restore test methods always run
- [x] Verify Android shards correctly
- [x] Add iOS support

## Example

smart_flank_0
smart_flank_1
smart_flank_2
smart_flank_3

### DIFF
--- a/test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
+++ b/test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
@@ -11,16 +11,15 @@ import com.google.cloud.storage.BucketInfo
 import com.google.cloud.storage.Storage
 import com.google.cloud.storage.StorageClass
 import com.google.cloud.storage.StorageOptions
-import com.google.common.math.IntMath
 import ftl.args.yml.IYmlMap
 import ftl.config.FtlConstants
 import ftl.config.FtlConstants.GCS_PREFIX
 import ftl.config.FtlConstants.JSON_FACTORY
 import ftl.config.FtlConstants.defaultCredentialPath
 import ftl.gc.GcStorage
+import ftl.shard.TestShard
 import ftl.util.Utils
 import java.io.File
-import java.math.RoundingMode
 import java.net.URI
 import java.nio.file.Files
 import java.nio.file.Path
@@ -95,39 +94,10 @@ object ArgsHelper {
         if (validTestMethods.isEmpty()) Utils.fatalError("$from has no tests")
     }
 
-    fun calculateShards(
-        testMethodsToShard: Collection<String>,
-        testMethodsAlwaysRun: Collection<String>,
-        testShards: Int
-    ): List<List<String>> {
-        val testShardMethods = testMethodsToShard.distinct().toMutableList()
-        testShardMethods.removeAll(testMethodsAlwaysRun)
-
-        val oneTestPerChunk = testShards == -1
-        var chunkSize = IntMath.divide(testShardMethods.size, testShards, RoundingMode.UP)
-
-        if (oneTestPerChunk || chunkSize < 1) {
-            chunkSize = 1
-        }
-
-        val testShardChunks = testShardMethods.asSequence()
-            .chunked(chunkSize)
-            .map { testMethodsAlwaysRun + it }
-            .toList()
-
-        // Ensure we don't create more VMs than requested. VM count per run should be <= testShards
-        if (!oneTestPerChunk && testShardChunks.size > testShards) {
-            Utils.fatalError("Calculated chunks $testShardChunks is > requested $testShards testShards.")
-        }
-        if (testShardChunks.isEmpty()) Utils.fatalError("Failed to populate test shard chunks")
-
-        return testShardChunks
-    }
-
     fun createJunitBucket(projectId: String, junitGcsPath: String) {
         if (FtlConstants.useMock || junitGcsPath.isEmpty()) return
         val bucket = junitGcsPath.drop(GCS_PREFIX.length).substringBefore('/')
-        createGcsBucket(projectId, bucket)
+            createGcsBucket(projectId, bucket)
     }
 
     fun createGcsBucket(projectId: String, bucket: String): String {
@@ -182,6 +152,7 @@ object ArgsHelper {
 
     // https://stackoverflow.com/a/2821201/2450315
     private val envRegex = Pattern.compile("\\$([a-zA-Z_]+[a-zA-Z0-9_]*)")
+
     private fun evaluateEnvVars(text: String): String {
         val buffer = StringBuffer()
         val matcher = envRegex.matcher(text)
@@ -201,5 +172,11 @@ object ArgsHelper {
         val searchDir = Paths.get(filePath).parent
 
         return ArgsFileVisitor("glob:$filePath").walk(searchDir)
+    }
+
+    fun convertShards(shards: List<TestShard>): List<List<String>> {
+        return shards.map { shard ->
+            shard.testMethods.map { it.name }
+        }
     }
 }
--- a/test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
+++ b/test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
@@ -17,7 +17,10 @@ import ftl.config.FtlConstants.GCS_PREFIX
 import ftl.config.FtlConstants.JSON_FACTORY
 import ftl.config.FtlConstants.defaultCredentialPath
 import ftl.gc.GcStorage
-import ftl.shard.TestShard
+import ftl.reports.xml.model.JUnitTestResult
+import ftl.shard.Shard
+import ftl.shard.StringShards
+import ftl.shard.stringShards
 import ftl.util.Utils
 import java.io.File
 import java.net.URI
@@ -174,9 +177,21 @@ object ArgsHelper {
         return ArgsFileVisitor("glob:$filePath").walk(searchDir)
     }
 
-    fun convertShards(shards: List<TestShard>): List<List<String>> {
-        return shards.map { shard ->
-            shard.testMethods.map { it.name }
+    fun calculateShards(filteredTests: List<String>, args: IArgs): List<List<String>> {
+        val oldTestResult = GcStorage.downloadJunitXml(args) ?: JUnitTestResult(mutableListOf())
+        val shardsByTime = Shard.calculateShardsByTime(filteredTests, oldTestResult, args.testShards)
+
+        return testMethodsAlwaysRun(shardsByTime.stringShards(), args)
+    }
+
+    private fun testMethodsAlwaysRun(shards: StringShards, args: IArgs): StringShards {
+        val alwaysRun = args.testTargetsAlwaysRun
+
+        shards.forEach { shard ->
+            shard.removeAll(alwaysRun)
+            shard.addAll(0, alwaysRun)
         }
+
+        return shards
     }
 }
--- a/test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
+++ b/test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
@@ -124,25 +124,31 @@ object ArgsHelper {
         return testShardChunks
     }
 
-    fun getGcsBucket(projectId: String, resultsBucket: String): String {
+    fun createJunitBucket(projectId: String, junitGcsPath: String) {
+        if (FtlConstants.useMock || junitGcsPath.isEmpty()) return
+        val bucket = junitGcsPath.drop(GCS_PREFIX.length).substringBefore('/')
+        createGcsBucket(projectId, bucket)
+    }
+
+    fun createGcsBucket(projectId: String, bucket: String): String {
         // com.google.cloud.storage.contrib.nio.testing.FakeStorageRpc doesn't support list
         // when testing, use a hard coded results bucket instead.
-        if (FtlConstants.useMock) return resultsBucket
+        if (FtlConstants.useMock) return bucket
         // test lab supports using a special free storage bucket
         // because we don't have access to the root account, it won't show up in the storage list.
-        if (resultsBucket.startsWith("test-lab-")) return resultsBucket
+        if (bucket.startsWith("test-lab-")) return bucket
 
         val storage = StorageOptions.newBuilder().setProjectId(projectId).build().service
         val bucketLabel = mapOf(Pair("flank", ""))
         val storageLocation = "us-central1"
 
-        val bucketListOption = Storage.BucketListOption.prefix(resultsBucket)
+        val bucketListOption = Storage.BucketListOption.prefix(bucket)
         val storageList = storage.list(bucketListOption).values?.map { it.name } ?: emptyList()
-        val bucket = storageList.find { it == resultsBucket }
-        if (bucket != null) return bucket
+        val targetBucket = storageList.find { it == bucket }
+        if (targetBucket != null) return targetBucket
 
         return storage.create(
-            BucketInfo.newBuilder(resultsBucket)
+            BucketInfo.newBuilder(targetBucket)
                 .setStorageClass(StorageClass.REGIONAL)
                 .setLocation(storageLocation)
                 .setLabels(bucketLabel)
--- a/test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
+++ b/test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
@@ -179,7 +179,7 @@ object ArgsHelper {
 
     fun calculateShards(filteredTests: List<String>, args: IArgs): List<List<String>> {
         val oldTestResult = GcStorage.downloadJunitXml(args) ?: JUnitTestResult(mutableListOf())
-        val shardsByTime = Shard.calculateShardsByTime(filteredTests, oldTestResult, args.testShards)
+        val shardsByTime = Shard.calculateShardsByTime(filteredTests, oldTestResult, args)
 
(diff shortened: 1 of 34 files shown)


test_runner/src/main/kotlin/ftl/reports/util/ReportManager.kt
```diff
@@ -81,6 +82,9 @@ object ReportManager {
 
         JUnitReport.run(matrices, testSuite)
 
+        val localJunitXmlPath = JUnitReport.reportPath(matrices)
+        GcStorage.uploadJunitXml(localJunitXmlPath, args)
```
sridhar-vadlamani-ck: I think this should only be uploaded if all tests passed.

bootstraponline: In the algorithm part of this feature, we'll be ignoring failed tests. Does that address your feedback?

It's exceptionally rare for all tests to pass on large test suites. If 99.99% of tests pass and one fails, then there's a lot of value in using the timing data from the passed tests. I'm not sure  discarding all test results if a single failure is present makes sense.

sridhar-vadlamani-ck: I think it's perfectly fine to toss the whole thing if tests are failing. Think of it this way. There's a `master` branch where _ALL_ tests are passing, and to merge into `master`, you need to pass tests anyway.

Failing tests just means you don't go update a perfectly working dataset of test times with your failing test data.

Fix your tests and it'll go update the timing xml.

test_runner/src/main/kotlin/ftl/args/yml/FlankYml.kt
```diff
@@ -2,24 +2,35 @@ package ftl.args.yml
...
     val testShards: Int = 1,
     val repeatTests: Int = 1,
+    val junitGcsPath: String = "",
 
     @field:JsonProperty("test-targets-always-run")
     val testTargetsAlwaysRun: List<String> = emptyList()
 ) {
     companion object : IYmlKeys {
-        override val keys = listOf("testShards", "repeatTests", "test-targets-always-run")
+        override val keys = listOf("testShards", "repeatTests", "junitGcsPath", "test-targets-always-run")
```
sridhar-vadlamani-ck: Question unrelated to PR.

What's this?

test_runner/src/main/kotlin/ftl/args/IosArgs.kt
```diff
@@ -67,6 +70,9 @@ class IosArgs(
     }
 
     init {
+        resultsBucket = createGcsBucket(projectId, gcloud.resultsBucket)
+        createJunitBucket(projectId, flank.junitGcsPath)
```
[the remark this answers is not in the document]
bootstraponline: Yep, firebase provides JUnit XML for both Android and iOS.

test_runner/src/main/kotlin/ftl/reports/util/ReportManager.kt
bootstraponline: The only problem I see is there are test suites where the tests never pass 100%. At that point the smart flank feature does nothing for them.

Let's have an example scenario of 1,000 tests with 1 failure. The failure is different each time because UI tests are flaky. In the 0 failures allowed model, smart flank does nothing until the test suite passes completely. In the always upload model, the test suite benefits regardless of if the suite perfectly executed.

I'm not sure why we wouldn't always want the benefit, even if sometimes it'll be partially beneficial. I agree with disregarding individual failing tests.

test_runner/src/main/kotlin/ftl/args/yml/FlankYml.kt
bootstraponline: `IYmlKeys` is used by the `flank doctor` command to detect unregonized YAML keys.

https://github.com/TestArmada/flank/blob/5715cd44bf784d2b5508d766f01c30aea7bbd26d/test_runner/src/main/kotlin/ftl/doctor/Doctor.kt#L37

test_runner/src/main/kotlin/ftl/reports/util/ReportManager.kt
sridhar-vadlamani-ck: It just needs to pass once. For an entire repository. As the location of the file is not variable between branches/builds, we need to be safe.

Imagine a scenario where _all_ tests fail. And it just happily goes and updates the xml file with wonky test times, maybe zeros.

Now, not only are the tests for your branch messed up. You also messed up sharding for the rest of your team.

sridhar-vadlamani-ck: Imho, It's not flank's responsibility to worry about cases where a team has flaky tests.

Smart flank will work if you have ONE successful test run.

bootstraponline: I don't think that's how [Smart Flank works in the Java code](https://github.com/TestArmada/flank/tree/flank_java). I'm worried about users getting surprised by this behavior after upgrading.

With the CLI option coming to flank, it's easy to have multiple timing files. You could dynamically define the timing by branch:

- `flank android run --junitGcsPath gs://myapp/flank/$BRANCH_NAME.xml`

> Now, not only are the tests for your branch messed up. You also messed up sharding for the rest of your team.

Ok, let's say a bad branch was pushed with all failing tests. Sharding is messed up. As soon as the next run happens, the timing info is good again.

Now let's say we have a flaky test suite of ~99% with no 100% successful runs. The timing info will never be good. There are a lot of companies out there with UI test suites that don't pass. Unfortunately flakiness is the norm.

Flank should do the right thing for the majority of users by default. While also [providing an API in the future](https://github.com/TestArmada/flank/issues/286) for users to customize behavior. I could see, for example, a team that has test suites that pass 100% wanting to enforce that for JUnit XML uploads. I'm not persuaded that makes sense for the default option.

bootstraponline: I think if we had a `--no-upload` option that might be useful. Local runs/WIP PRs could then optionally use the existing XML without having to always upload a new one.

sridhar-vadlamani-ck: I find it hard to believe an organization would not devote some time to tests which _always_ fail ( which is the only scenario where you would want to upload without checking for success. )

The thing is, when you upload without ensuring success, the timing info is _never_ good. You always shard off of potentially bad data. If a test is failing, it might fail in 0.5 seconds. This would only make the implementation wonky as it raises questions like - 

- If a test has failed fast in our reference junit xml, do I shard it using its very low time value? Even though it might be a junk value as it failed quickly? For example a test may fail in 0.5s when it usually takes 3s. Now you're sharding subsequent tests based off of wrong data anyway.

- If the test hangs, you might have large values. How do you handle that?

- Are you just going to ignore failed tests when it comes to deciding shard assignment?

sridhar-vadlamani-ck: Alternatively, just make storing regardless of failure a configurable option. gRPC is coming. But that's sometime in Q1/Q2 2019.

sridhar-vadlamani-ck: > I don't think that's how Smart Flank works in the Java code. I'm worried about users getting surprised by this behavior after upgrading.

Not necessarily a bad thing. Flank java was never very good with this as Firebase's junit was messed up. So everytime a new test was added, they'd single shard the whole test suite.

I remember trying to do that, and it took forever. So I scrapped time based sharding alltogether. And We have one of the smaller test suites out there.

sridhar-vadlamani-ck: > In the algorithm part of this feature, we'll be ignoring failed tests. Does that address your feedback?

I only just saw this. 🤦‍♂️ 

In a way, yes. I suggest we do some extra work and only replace/add the test times for non-failing tests. So instead of uploading the firebase provided junit xml, we could process and upload our own or even use json.

This would be the best of both worlds. As if a test is failing, its time is not updated. But a time from when it _did_ work isn't lost either.

bootstraponline: > The thing is, when you upload without ensuring success, the timing info is _never_ good. You always shard off of potentially bad data.

I think the timing info for tests that passed is good? The XML includes passed/failed/skipped. If 1 out of 1000 fails then I have good timing data for 999 tests. The failure is discarded.

> * Are you just going to ignore failed tests when it comes to deciding shard assignment?

I think so. This is an existing feature in Flank Java that worked well at scale. No other test runner that I'm aware of requires perfect pass rates to take advantage of test timing information. We have to ignore new tests anyway since they won't have previous timing data.

If we discover this is a real problem I'm happy to revisit in the future. With Flank Java, we didn't have any such reports.

sridhar-vadlamani-ck: > If we discover this is a real problem I'm happy to revisit in the future. With Flank Java, we didn't have any such reports.

Probably because it wasn't used much. 🙈

bootstraponline: Here are the options as I understand them:

- Upload only on 100% pass rate
    - Pros
        - Timing info is accurate for all tests on a perfect run
    - Cons
        - Timing info is out of date if the last run wasn't a perfect run
        - Unhealthy test suites (< 100% pass rate) don't benefit from time data

- Always upload
    - Pros
        - Timing info is accurate for passed tests
    - Cons
        - Not all tests may have passed.
        - Test failures remove test timing info from the XML

- Always upload + merge success
    - Pros
        - Timing info is accurate for passed tests.
        - If a test failed previously, then we keep the last known good timing
    - Cons
        - Additional complexity to implement

sridhar-vadlamani-ck: Option 3 has complexity at the implementation ( and not _that_ much either ). Option 1 and 2 are detrimental to the workflows of the consumers. A subset in each case. ( 2 in my case. 1 in yours )

2 especially so (imo) , as _some_ tests may fail, or _all_ tests may fail. Making sharding properly the exception rather than the norm due to the inconsistency of the data.

sridhar-vadlamani-ck: I can help with the implementation of 3.

bootstraponline: I agree that `3) Always upload + merge success`  makes the most sense.

test_runner/src/main/kotlin/ftl/reports/xml/model/JUnitTestSuite.kt
```diff
@@ -71,4 +71,59 @@ data class JUnitTestSuite(
 
         return this
     }
+
+    fun mergeTestTimes(other: JUnitTestSuite): JUnitTestSuite {
+        if (this.name != other.name) throw RuntimeException("Attempted to merge ${other.name} into ${this.name}")
+
+        // for each new JUnitTestCase
+        //   iif it failed then pull timing info from old
```
sridhar-vadlamani-ck: typo here.

test_runner/src/main/kotlin/ftl/reports/xml/model/JUnitTestSuite.kt
```diff
@@ -71,4 +71,59 @@ data class JUnitTestSuite(
...
+        //   remove if not successful in either new or old
```
sridhar-vadlamani-ck: inconsistent spacing?

test_runner/src/main/kotlin/ftl/reports/xml/model/JUnitTestSuite.kt
```diff
@@ -71,4 +71,59 @@ data class JUnitTestSuite(
 
         return this
     }
+
+    fun mergeTestTimes(other: JUnitTestSuite): JUnitTestSuite {
+        if (this.name != other.name) throw RuntimeException("Attempted to merge ${other.name} into ${this.name}")
```
sridhar-vadlamani-ck: I didn't see the `name` when I was coding. 🙈

test_runner/src/test/kotlin/ftl/args/AndroidArgsTest.kt
```diff
@@ -197,6 +197,7 @@ AndroidArgs
     flank:
       testShards: 7
       repeatTests: 8
+      junitGcsPath:${' '}
```
sridhar-vadlamani-ck: What's the `${' '}`?

test_runner/src/test/kotlin/ftl/args/ArgsHelperTest.kt
```diff
@@ -153,7 +153,7 @@ class ArgsHelperTest {
 
     @Test
     fun getGcsBucket_succeeds() {
-        getGcsBucket("123", "results_bucket")
+        createGcsBucket("123", "results_bucket")
```
sridhar-vadlamani-ck: Test name needs to be updated.

test_runner/src/test/kotlin/ftl/args/AndroidArgsTest.kt
bootstraponline: That's an empty space. Regular empty spaces are auto trimmed by IntelliJ.

test_runner/src/main/kotlin/ftl/gc/GcStorage.kt
```diff
@@ -76,7 +99,7 @@ object GcStorage {
         return gcsFilePath
     }
 
-    private fun download(gcsUriString: String): String {
+    private fun download(gcsUriString: String, ignoreError: Boolean = false): String {
```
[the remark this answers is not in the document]
bootstraponline: agreed, we can probably make this an empty string.

test_runner/src/main/kotlin/ftl/reports/xml/model/JUnitTestSuite.kt
bootstraponline: `iif` = if and only if

https://math.stackexchange.com/questions/68293/what-is-the-difference-between-only-if-and-iff

bootstraponline: i'll fix 😭

bootstraponline: Fixed.

test_runner/src/main/kotlin/ftl/reports/xml/model/JUnitTestSuite.kt
bootstraponline: Fixed.

test_runner/src/main/kotlin/ftl/gc/GcStorage.kt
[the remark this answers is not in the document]
bootstraponline: Fixed. Renamed to "tmp"

test_runner/src/test/kotlin/ftl/args/ArgsHelperTest.kt
bootstraponline: Fixed.

test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
```diff
@@ -102,25 +102,31 @@ object ArgsHelper {
...
+
+    fun createGcsBucket(projectId: String, bucket: String): String {
         // com.google.cloud.storage.contrib.nio.testing.FakeStorageRpc doesn't support list
         // when testing, use a hard coded results bucket instead.
-        if (FtlConstants.useMock) return resultsBucket
+        if (FtlConstants.useMock) return bucket
         // test lab supports using a special free storage bucket
         // because we don't have access to the root account, it won't show up in the storage list.
-        if (resultsBucket.startsWith("test-lab-")) return resultsBucket
+        if (bucket.startsWith("test-lab-")) return bucket
```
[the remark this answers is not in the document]
bootstraponline: the file must be hosted on gcs for us to fetch it on subsequent runs.

`test-lab-` is a prefix of the free bucket, not a test helper.

there are limitations in the google cloud storage server, that's why there's an early return on `useMock`. If we enhanced the server API support then the code would be cleaner.

test_runner/src/main/kotlin/ftl/gc/GcStorage.kt
```diff
@@ -40,6 +41,24 @@ object GcStorage {
             runGcsPath = runGcsPath
         )
 
+    fun uploadJunitXml(localJunitXml: String, args: IArgs) {
```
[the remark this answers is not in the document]
bootstraponline: no because the firebase run is a random path, we need a fixed path that's consistent across all runs

test_runner/src/main/kotlin/ftl/shard/Shard.kt
```diff
@@ -0,0 +1,113 @@
...
+    val name: String,
+    val time: Double
+)
+
+data class TestShard(
+    val time: Double,
+    val testMethods: MutableList<TestMethod>
+)
+
+data class TestShards(
```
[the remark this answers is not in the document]
bootstraponline: it's a helper to make the code easier to work with.

PR #386, opened 2018-11-11
Allow blob, tilde and env vars in app paths

### LINKED ISSUE #336: Sravan | Always find xctestrun file - avoid hard coding file name
search for `$DD/Build/Products/*.xctestrun` or similar instead of requiring a hard coded path that changes as new iOS versions are released

Support environment variables and tilda as well

we should enable auto discovery by default

---

Also support wildcard for APK discovery. Look into supporting `~` expansion as well.

sravanmedarapu:
app, test and xctestrun-file values in Flank config should support below formats along with current formats.

1. tilda expansion(`~`)
2. environment variable expansion
3. globbing with `*` and `**`

```
Ex: 
app: ~/repo/app-*.apk
xctestrun-file: $DERIVED_DATA/**/*. xctestrun
```

Fix #336


test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
```diff
@@ -45,6 +48,24 @@ object ArgsHelper {
         }
     }
 
+    fun evaluateFilePath(filePath: String): String {
+        var file = filePath.trim().replaceFirst("~", System.getProperty("user.home"))
```
bootstraponline: ~ should be replaced only if it's the first character in the string
https://stackoverflow.com/a/7163455

the regex should be `^~`, we should add a JUnit test for this (for both the happy path and failure case).

`~/tmp` should be replaced
`/tmp/~` should not be replaced

test_runner/src/main/kotlin/ftl/args/ArgsFileVisitor.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    companion object {
+        private const val RECURSE = "/**"
+    }
+
+    @Throws(java.nio.file.NoSuchFileException::class)
+    fun walk(searchPath: Path): List<Path> {
+        val searchString = searchPath.toString()
+        // /Users/tmp/code/flank/test_app/** => /Users/tmp/code/flank/test_app/
+        val beforeGlob = Paths.get(searchString.substringBefore(RECURSE))
```
bootstraponline: we probably need more testing around paths with `/**/` and `/*/`

i'm not sure splitting by `/**/` works for all cases

test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
```diff
@@ -150,4 +172,24 @@ object ArgsHelper {
         // Allow users control over projectId by checking using Google's logic first before falling back to JSON.
         return ServiceOptions.getDefaultProjectId() ?: serviceAccountProjectId()
     }
+
+    // https://stackoverflow.com/a/2821201/2450315
+    private val envRegex = Pattern.compile("\\$([a-zA-Z_]+[a-zA-Z0-9_]*)")
+    private fun substituteEnvVars(text: String): String {
```
bootstraponline: we should check this works for duplicate environment variables via more JUnit tests. There are probably other edge cases as well. For example the `${HOME}` syntax which we should probably document as unsupported.

> $HOME/$HOME/tmp

doodla: We should probably _WARN_ but not fail for duplicate env variables. The warning is to notify the user that he's duplicated an env variable. Not failing, in case its intentional.

Regardless, I don't think we should handle custom env vars. That would likely be a pain to maintain.

test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
```diff
@@ -150,4 +172,24 @@ object ArgsHelper {
...
+        val buffer = StringBuffer()
+        val matcher = envRegex.matcher(text)
+        while (matcher.find()) {
+            val envName = matcher.group(1)
+            val envValue = System.getenv(envName) ?: ""
```
doodla: Should we raise an error/warning here? Silently ignoring un-found env variables seems like a bug.

test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
```diff
@@ -45,6 +48,24 @@ object ArgsHelper {
...
     }
 
+    fun evaluateFilePath(filePath: String): String {
+        var file = filePath.trim().replaceFirst("~", System.getProperty("user.home"))
+        file = substituteEnvVars(file)
+        // avoid File(..).canonicalPath since that will resolve symlinks
+        file = Paths.get(file).toAbsolutePath().normalize().toString()
+
+        // Avoid walking the folder's parent dir if we know it exists already.
+        if (File(file).exists()) return file
```
doodla: I would prefer if we also printed this value to console. Something like - `Resolved app file path - "$file"`.
Under a conditional. `if (filePath != file)`.

bootstraponline: I think it should be printed already when the args are printed to stdout at the start of a test run.

test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
bootstraponline: bash doesn't warn on duplicates. I'm using bash as the model for the parsing logic.

test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
bootstraponline: this is standard behavior in Bash, that's how environment variables work sadly.

bootstraponline: discussed on slack, printing a warning makes sense.

test_runner/src/test/kotlin/ftl/args/ArgsHelperTest.kt
```diff
@@ -161,4 +173,74 @@ class ArgsHelperTest {
...
+
+        return file.absolutePath
+    }
+
+    private fun String.absolutePath(): String {
+        return File(this).absolutePath
+    }
+
+    @Test
+    fun tmpTest() {
```
bootstraponline: This should be deleted.

Issue #365, opened 2018-10-23
CLI support for all args


bootstraponline:
Match gcloud CLI in providing CLI arguments for all options.

- [Android Run docs](https://cloud.google.com/sdk/gcloud/reference/firebase/test/android/run)
- [iOS Run Docs](https://cloud.google.com/sdk/gcloud/reference/alpha/firebase/test/ios/run)

### Android

- gcloudYml
  - [x] resultsBucket `--results-bucket=RESULTS_BUCKET` #413
  - [x] recordVideo `--record-video` `--no-record-video` #414
  - [x] timeout `--timeout=TIMEOUT` #415
  - [x] async `--async`#416
  - [x] project `--project=PROJECT` #417
  - [x] resultsHistoryName `--results-history-name=RESULTS_HISTORY_NAME` #418
- androidGcloudYml
  - [x] app `--app=APP` https://github.com/TestArmada/flank/pull/389
  - [x] test `--test=TEST` https://github.com/TestArmada/flank/pull/393
  - [x] autoGoogleLogin `--auto-google-login` `--no-auto-google-login` #405
  - [x] useOrchestrator `--use-orchestrator` `--no-use-orchestrator` https://github.com/TestArmada/flank/pull/395
  - [x] environmentVariables `--environment-variables=[KEY=VALUE,…]` #407
  - [x] directoriesToPull `--directories-to-pull=[DIR_TO_PULL,…]` #408
  - [x] performanceMetrics `--performance-metrics` `--no-performance-metrics` #406
  - [x] testTargets `--test-targets=TEST_TARGET,[TEST_TARGET,…]]` https://github.com/TestArmada/flank/pull/394
  - [x] device `--device=DIMENSION=VALUE,[…]` #409
- flankYml
  - [x] testShards `--test-shards=TEST_SHARDS` #419
  - [x] repeatTests `--repeat-tests=REPEAT_TESTS` #420
  - [x] testTargetsAlwaysRun `--test-targets-always-run=TEST_TARGET,[TEST_TARGET,…]]` #421
  - ? smartFlankGcsPath

### iOS

- gcloudYml
    - [x] resultsBucket `--results-bucket=RESULTS_BUCKET` #413
    - [x] recordVideo `--record-video` `--no-record-video` #414
    - [x] timeout `--timeout=TIMEOUT` #415
    - [x] async `--async` #416
    - [x] project `--project=PROJECT` #417
[...]

PR #394, opened 2018-11-14
--test-targets param

Macarse:
#365

### DIFF
--- a/test_runner/src/test/kotlin/ftl/args/AndroidArgsTest.kt
+++ b/test_runner/src/test/kotlin/ftl/args/AndroidArgsTest.kt
@@ -315,4 +315,25 @@ AndroidArgs
         assertThat(androidArgs.testApk).isEqualTo(appApk)
         assertThat(androidArgs.cli).isEqualTo(cli)
     }
+
+    @Test
+    fun androidArgs_overrideTestTargetsFromCmdLine() {
+        val cli = AndroidRunCommand()
+        CommandLine(cli).parse("--test-targets", "class com.foo.ClassName")
+
+        val androidArgs = AndroidArgs.load(
+            """
+        gcloud:
+          app: $appApk
+          test: $testApk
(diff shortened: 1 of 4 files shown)


test_runner/src/test/kotlin/ftl/cli/firebase/test/android/AndroidRunCommandTest.kt
```diff
@@ -74,4 +75,13 @@ class AndroidRunCommandTest {
         CommandLine(cmd).parse("--test", "myTestApp.apk")
         assertThat(cmd.test).isEqualTo("myTestApp.apk")
     }
+
+    @Test
+    fun testTargets_parse() {
+        val cmd = AndroidRunCommand()
+        CommandLine(cmd).parse("--test-targets", "class com.foo.Clazz", "--test-targets", "package com.my.package")
```
bootstraponline: probably better to extract `"class com.foo.Clazz", "--test-targets", "package com.my.package"` to a variable.

I think you can make them a list and then destructure with *.

val expectedTargets = listOf("class com.foo.Clazz", "--test-targets", "package com.my.package")
...

test_runner/src/test/kotlin/ftl/args/AndroidArgsTest.kt
```diff
@@ -315,4 +315,25 @@ AndroidArgs
         assertThat(androidArgs.testApk).isEqualTo(appApk)
         assertThat(androidArgs.cli).isEqualTo(cli)
     }
+
+    @Test
+    fun androidArgs_overrideTestTargetsFromCmdLine() {
+        val cli = AndroidRunCommand()
+        CommandLine(cli).parse("--test-targets", "class com.foo.ClassName")
```
bootstraponline: extract `"class com.foo.ClassName")` to a variable

PR #395, opened 2018-11-15
--use-orchestrator param

Macarse:
#365

### DIFF
--- a/test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
+++ b/test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
@@ -50,6 +50,8 @@ class AndroidArgs(
     val appApk = cli.app ?: androidGcloud.app
     val testApk = cli.test ?: androidGcloud.test
     val autoGoogleLogin = androidGcloud.autoGoogleLogin
+
+    // We use not() on noUseOrchestrator because if the flag is on, useOrchestrator needs to be false
     val useOrchestrator = cli.useOrchestrator ?: cli.noUseOrchestrator?.not() ?: androidGcloud.useOrchestrator
     val environmentVariables = androidGcloud.environmentVariables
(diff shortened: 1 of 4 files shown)


test_runner/src/main/kotlin/ftl/cli/firebase/test/android/AndroidRunCommand.kt
```diff
@@ -51,4 +51,11 @@ class AndroidRunCommand : Runnable {
           See https://developer.android.com/reference/android/support/test/runner/AndroidJUnitRunner for more
           information."""])
     var testTargets: List<String>? = null
+
+    @Option(names = ["--use-orchestrator"], description = ["""Whether each test runs in its own Instrumentation instance
+        |with the Android Test Orchestrator (default: Orchestrator is not used, same as specifying --no-use-orchestrator).
```
bootstraponline: > default: Orchestrator is not used

The default in flank is to use orchestrator. I think we may want to add `--no-use-orchestrator` to this PR

test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
```diff
@@ -50,7 +50,7 @@ class AndroidArgs(
     val appApk = cli.app ?: androidGcloud.app
     val testApk = cli.test ?: androidGcloud.test
     val autoGoogleLogin = androidGcloud.autoGoogleLogin
-    val useOrchestrator = androidGcloud.useOrchestrator
+    val useOrchestrator = cli.useOrchestrator ?: cli.noUseOrchestrator?.not() ?: androidGcloud.useOrchestrator
```
Macarse: `not()` because if the flag is on, useOrchestrator needs to be `false`

bootstraponline: let's add this as a comment, the double negative is confusing

`no.not()` 😂

PR #397, opened 2018-11-15
SmartFlank algo

Macarse:


test_runner/src/main/kotlin/ftl/shard/Shard.kt
```diff
@@ -1,123 +1,60 @@
...
-        while (testList.size > 0) {
-            if (shardCount == maxTestShards) {
-                // we have "filled" all shards, reset starting duration/position
-                startingDuration = testList.first().time
-                System.out.println("Starting duration $startingDuration")
-                shardCount = 0
+        // Create a map with information from previous junit run
+        testResult.testsuites?.forEach { testsuite ->
+            testsuite.testcases?.forEach { testcase ->
+                junitMap[testcase.name] = testcase.time.toDouble()
```
bootstraponline: testcase name isn't unique. The fully qualified name is `suite.name + testCase.classname + testCase.name`

https://github.com/TestArmada/flank/blob/bd8dd925f6f10a61fe583b03b63228daa0f6a885/test_runner/src/main/kotlin/ftl/reports/HtmlErrorReport.kt#L32

test_runner/src/main/kotlin/ftl/shard/Shard.kt
```diff
@@ -1,123 +1,60 @@
...
-                testMethods.remove()
-
-                continue
-            }
+        // Create the list of shards we will return
+        var shards = List(shardsCount) { TestShard(0.0, mutableListOf()) }
 
-            val noTestsAdded = timeBudget == targetShardDuration
-            val testOverBudget = test.time >= timeBudget
+        // We want to iterate over testcase going from higher to lower
```
bootstraponline: >  higher to lower

I think slowest to fastest is more descriptive.

test_runner/src/main/kotlin/ftl/shard/Shard.kt
```diff
@@ -1,123 +1,60 @@
...
-                return TestShard(shardTime, shardTests)
-            }
+            val tm = TestMethod(it.name, it.time)
+            val shard = shards[0]
+
+            shard.testMethods.add(tm)
+            shard.time += tm.time
+
+            // Sort the shards to keep the most empty shard first
+            shards = shards.sortedBy { it.time }
```
bootstraponline: it is shadowed here

image

test_runner/src/main/kotlin/ftl/shard/Shard.kt
```diff
@@ -1,123 +1,60 @@
...
+        val testCount = testcases.size
 
-        // Ensure we don't create more VMs than requested. VM count per run should be <= testShards
-        if (testShardChunks.size > maxTestShards) {
-            fatalError("Calculated chunks $testShardChunks is > requested $maxTestShards testShards.")
-        }
-        if (testShardChunks.isEmpty()) fatalError("Failed to populate test shard chunks")
-    }
+        // If maxShards is infinite or we have more shards than tests, let's match it
+        val shardsCount = if (maxShards == -1 || maxShards > testCount) testCount else maxShards
```
bootstraponline: I think this is great, we should be able to drop the old chunk based sharding method.

We'll still want to add in test-methods-always-run, that could be in a different method to follow the single responsibility principle.

test_runner/src/main/kotlin/ftl/shard/Shard.kt
```diff
@@ -1,123 +1,60 @@
...
-                // Always add at least 1 test per shard regardless of budget
-                shardTime += test.time
-                shardTests.add(test)
-                testMethods.remove()
+        testcases.forEach {
 
-                return TestShard(shardTime, shardTests)
-            }
+            val tm = TestMethod(it.name, it.time)
+            val shard = shards[0]
```
bootstraponline:
```kotlin
val shard = shards.first()
```

test_runner/src/main/kotlin/ftl/shard/Shard.kt
Macarse: great! I wanted to test that by hooking with a real run.

test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
```diff
@@ -80,11 +83,22 @@ class AndroidArgs(
 
         val filteredTests = getTestMethods(testLocalApk)
 
-        calculateShards(
-            filteredTests,
-            testTargetsAlwaysRun,
-            testShards
-            )
+        if (true) {
```
bootstraponline: probably we should remove this?

test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
```diff
@@ -80,11 +83,22 @@ class AndroidArgs(
...
+
+            val start = System.nanoTime()
+            val result = convertShards(Shard.calculateShardsByTime(filteredTests, oldTestResult, testShards))
+            val shardCreationTime = TimeUnit.SECONDS.convert(System.nanoTime() - start, TimeUnit.NANOSECONDS)
+
+            println("Shard calculation took: $shardCreationTime")
+
+            result
+        } else {
+            calculateShards(
```
bootstraponline: testTargetsAlwaysRun support is missing. I think calculateShards should be removed

test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
```diff
@@ -80,11 +83,22 @@ class AndroidArgs(
...
-            testShards
-            )
+        if (true) {
+            val oldTestResult = GcStorage.downloadJunitXml(this) ?: JUnitTestResult(mutableListOf())
+
+            val start = System.nanoTime()
+            val result = convertShards(Shard.calculateShardsByTime(filteredTests, oldTestResult, testShards))
+            val shardCreationTime = TimeUnit.SECONDS.convert(System.nanoTime() - start, TimeUnit.NANOSECONDS)
+
+            println("Shard calculation took: $shardCreationTime")
```
bootstraponline: I don't think users will benefit from shard calculation numbers. This makes more sense in a JUnit test.

test_runner/src/main/kotlin/ftl/shard/Shard.kt
```diff
@@ -1,64 +1,61 @@
...
+            testsuite.testcases?.forEach { testcase ->
+                val key = "${testsuite.name}${testcase.classname}#${testcase.name}".trim()
+                junitMap[key] = testcase.time.toDouble()
             }
+        }
+
+        val testcases = mutableListOf<TestMethod>()
+        runningTests.forEach {
+            // junitMap doesn't include `class `, we remove it to search in the map
+            val key = it.replace("class ", "")
```
bootstraponline: let's use `replaceFirst`

test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
Macarse: Yep. will remove.

PR #441, opened 2018-12-21
Add files-to-download

winterDroid:
Support downloading directories after test run from the result bucket
Upgrade Guava to 25.1 due to NoSuchMethodError while running tests (see https://github.com/google/truth/issues/473)
Fix missing download of PNGs after test run


winterDroid:
Fixes https://github.com/TestArmada/flank/issues/438

Macarse:
@winterDroid What's the diff between `directories-to-download` and `--directories-to-pull` from https://github.com/TestArmada/flank/pull/408?

winterDroid:
This one will download the directories specified from the bucket to your local storage, the other one downloads from the device to the bucket

Macarse:
Oh, I see. I guess the cmd is not part of gcloud and that's why it was not part of https://github.com/TestArmada/flank/issues/365. 
I wonder if we should add `bucket` to the cmd somehow to avoid confusion. Something like `directories-to-download-from-bucket`. wdyt?

winterDroid:
I thought about that, but I think the problem with that is that it is hard to know the full path.. Maybe it would be better to have it `files-to-download` which takes a regex and not just a string..

Macarse:
I tested this and it works as expected. I will test this on iOS tomorrow but it should work.
@bootstraponline plz review and merge :+1:

bootstraponline:
I'll take a look when I'm back at work tomorrow.

bootstraponline:
Auto downloading pngs was intentionally removed because it's slow and there's no way to disable the behavior. I think the new dir-to-download feature will allow users to opt into downloading what they want (pngs included).

bootstraponline:
> Maybe it would be better to have it `files-to-download` which takes a regex and not just a string..

I think that's a great suggestion. I updated the code to work on a regex that downloads files.

Macarse:
Tested on iOS and Android. Merge! :+1:

bootstraponline:
The feature has been merged into Flank. It'll be part of the v4.2 release. Thanks for contributing!

PR #449, opened 2019-01-09
flank auth login

### LINKED ISSUE #398: gcloud auth login support
Some teams don't allow sharing the service account json. Support user auth as an option in flank.

Requested by Slack.

Looking for additional feedback on if this is a useful feature to add or not. Code to add the feature is here: https://github.com/TestArmada/flank/pull/209

Update: Add this in Jan 2019

bootstraponline:
Fix #398 

Flank now allows you to login as a user. Type `flank auth login` and follow the web prompt. User auth is an alternative to using a service account.

### DIFF
--- a/test_runner/src/main/kotlin/ftl/Main.kt
+++ b/test_runner/src/main/kotlin/ftl/Main.kt
@@ -1,5 +1,6 @@
 package ftl
 
+import ftl.cli.AuthCommand
 import ftl.cli.FirebaseCommand
 import ftl.cli.firebase.CancelCommand
 import ftl.cli.firebase.RefreshCommand
@@ -16,7 +17,8 @@ import picocli.CommandLine
         IosCommand::class,
         AndroidCommand::class,
         RefreshCommand::class,
-        CancelCommand::class
+        CancelCommand::class,
+        AuthCommand::class
     ]
 )
 class Main : Runnable {
@@ -36,6 +38,10 @@ class Main : Runnable {
     private var printVersion = false
 
     companion object {
+        init {
+//            GoogleApiLogger.logAllToStdout()
+        }
+
         @JvmStatic
         fun main(args: Array<String>) {
             // val args = arrayOf("firebase", "test", "android", "run") // for debugging. run test from IntelliJ IDEA
--- a/test_runner/src/main/kotlin/ftl/gc/GcAuth.kt
+++ b/test_runner/src/main/kotlin/ftl/gc/GcAuth.kt
@@ -1,0 +1,59 @@
+package ftl.gc
+
+import com.google.api.client.auth.oauth2.Credential
+import com.google.api.client.extensions.java6.auth.oauth2.AuthorizationCodeInstalledApp
+import com.google.api.client.extensions.jetty.auth.oauth2.LocalServerReceiver
+import com.google.api.client.googleapis.auth.oauth2.GoogleAuthorizationCodeFlow
+import com.google.api.client.googleapis.auth.oauth2.GoogleCredential
+import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport
+import com.google.api.client.json.jackson2.JacksonFactory
+import com.google.api.client.util.store.FileDataStoreFactory
+import ftl.config.FtlConstants
+import java.io.File
+import java.io.IOException
+
+// https://github.com/googleapis/google-oauth-java-client
+// GoogleAuthorizationCodeFlow usage based on https://developers.google.com/sheets/api/quickstart/java
+object GcAuth {
+    private val HOME = System.getProperty("user.home")
+    private val CRED_FOLDER = File(HOME, ".flank/")
+    val CRED = File(CRED_FOLDER, "StoredCredential")
+
+    private val JSON_FACTORY = JacksonFactory.getDefaultInstance()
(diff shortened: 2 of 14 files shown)


test_runner/src/main/kotlin/ftl/Main.kt
```diff
@@ -36,6 +39,10 @@ class Main : Runnable {
     private var printVersion = false
 
     companion object {
+        init {
+//            GoogleApiLogger.logAllToStdout()
```
valeraz: Is this left here for debugging purposes?

test_runner/src/main/kotlin/ftl/gc/GcAuth.kt
```diff
@@ -0,0 +1,76 @@
...
+import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport
+import com.google.api.client.http.HttpTransport
+import com.google.api.client.json.jackson2.JacksonFactory
+import com.google.api.client.util.store.FileDataStoreFactory
+import ftl.config.FtlConstants
+import java.io.File
+import java.io.IOException
+
+// https://github.com/googleapis/google-oauth-java-client
+// Code from: https://developers.google.com/sheets/api/quickstart/java
```
valeraz: I'd help to add some context to the comment (like why you chose to copy from sheets)

test_runner/src/main/kotlin/ftl/gc/GcAuth.kt
```diff
@@ -0,0 +1,76 @@
...
+    // https://github.com/bootstraponline/gcloud_cli/blob/40521a6e297830b9f652a9ab4d8002e309b4353a/google-cloud-sdk/platform/gsutil/gslib/utils/system_util.py#L177
+    private val clientId = "32555940559.apps.googleusercontent.com"
+    private val clientSecret = "ZmssLNjJy2998hD4CTg2ejr2"
+
+    init {
+        try {
+            HTTP_TRANSPORT = GoogleNetHttpTransport.newTrustedTransport()
+            DATA_STORE_FACTORY = FileDataStoreFactory(CRED_FOLDER)
+        } catch (t: Throwable) {
+            t.printStackTrace()
```
valeraz: Not sure if this is copied code or not, but why not let the exception propagate or just rethrow it as a RuntimeException?

test_runner/src/main/kotlin/ftl/gc/GcAuth.kt
```diff
@@ -0,0 +1,76 @@
...
+            System.exit(1)
+        }
+    }
+
+    fun activateUserAuth(): Boolean {
```
valeraz: missing kdoc. Also, it seems strange that this would return a boolean (based on the naming of the method).

test_runner/src/main/kotlin/ftl/gc/GcAuth.kt
```diff
@@ -0,0 +1,76 @@
...
+    private val HOME = System.getProperty("user.home")
+    private val JSON_FACTORY = JacksonFactory.getDefaultInstance()
+    private var HTTP_TRANSPORT: HttpTransport? = null
+    private var DATA_STORE_FACTORY: FileDataStoreFactory? = null
+
+    private val CRED_FOLDER = File(HOME, ".flank/")
+    val CRED = File(CRED_FOLDER, "StoredCredential")
+
+    // https://github.com/bootstraponline/gcloud_cli/blob/40521a6e297830b9f652a9ab4d8002e309b4353a/google-cloud-sdk/platform/gsutil/gslib/utils/system_util.py#L177
+    private val clientId = "32555940559.apps.googleusercontent.com"
```
valeraz: Is it by design that these are hardcoded?

test_runner/src/main/kotlin/ftl/Main.kt
bootstraponline: Yeah, I think it'll be cleaned up as part of https://github.com/TestArmada/flank/issues/446

test_runner/src/main/kotlin/ftl/gc/GcAuth.kt
bootstraponline: I updated the comment. The usage of `GoogleAuthorizationCodeFlow` is the same for any Google service. When the API changes, Google updates their official reference documentation.

test_runner/src/main/kotlin/ftl/gc/GcAuth.kt
bootstraponline: That's a good point. Updated.

test_runner/src/main/kotlin/ftl/gc/GcAuth.kt
bootstraponline: renamed to `hasUserAuth`

test_runner/src/main/kotlin/ftl/gc/GcAuth.kt
bootstraponline: Yep, they're hardcoded in gcloud CLI as well.

test_runner/src/main/kotlin/ftl/gc/GcAuth.kt
bootstraponline: kdoc isn't used at the moment on the methods. A PR will be adding [detekt](https://github.com/arturbosch/detekt) support soon. At that point we may decide to enforce function docs.

PR #470, opened 2019-01-22
Issue #399: Smart Flank - Dynamic bucket counts

### LINKED ISSUE #399: Smart Flank - Dynamic bucket counts
The current smart flank algorithm shards into user defined buckets based on time. A user may want to define a time value for buckets, and then let flank decide how many buckets to use.

Mention this in design doc. Maybe not for core flank due to resource exhaustion anti-pattern. Specifying a max shard may resolve the resource issue.

Macarse:
Fixes #399

### DIFF
--- a/test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
+++ b/test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
@@ -178,9 +178,17 @@ object ArgsHelper {
 
     fun calculateShards(filteredTests: List<String>, args: IArgs): List<List<String>> {
         val oldTestResult = GcStorage.downloadJunitXml(args) ?: JUnitTestResult(mutableListOf())
-        val shardsByTime = Shard.calculateShardsByTime(filteredTests, oldTestResult, args)
 
-        return testMethodsAlwaysRun(shardsByTime.stringShards(), args)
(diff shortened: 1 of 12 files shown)


test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
```diff
@@ -178,9 +178,17 @@ object ArgsHelper {
 
     fun calculateShards(filteredTests: List<String>, args: IArgs): List<List<String>> {
         val oldTestResult = GcStorage.downloadJunitXml(args) ?: JUnitTestResult(mutableListOf())
-        val shardsByTime = Shard.calculateShardsByTime(filteredTests, oldTestResult, args)
 
-        return testMethodsAlwaysRun(shardsByTime.stringShards(), args)
+        var shards: MutableList<MutableList<String>>
+
+        try {
+            shards = Shard.createShardsByShardTime(filteredTests, oldTestResult, args)
```
Macarse: This can be improved. After applying this algo to my apk's with `shardTime=110`, the run ended up creating: `Shard times: 40s, 106s, 107s, 108s, 108s, 108s, 109s`

Another possibility can be running `Shard.createShardsByShardTime` to get the shard count, and using that value in `Shard.createShardsByShardCount`. That will make tests more distributed across shards.

PR #478, opened 2019-01-27
Issue #467: iOS support globs for test-targets

### LINKED ISSUE #467: iOS support globs (*) for test-targets
It is a feature request. 

Now only support:

```
test-targets:
    - ClassTest/testMethod
```

Would be good to have:

```
test-targets:
    - Class1Test/*
    - Class2Test/testLogin*
```

Macarse:
Fixes #467 
cc @gtroshin

### DIFF
--- a/test_runner/src/main/kotlin/ftl/args/IosArgs.kt
+++ b/test_runner/src/main/kotlin/ftl/args/IosArgs.kt
@@ -6,7 +6,6 @@ import ftl.args.ArgsHelper.createGcsBucket
 import ftl.args.ArgsHelper.createJunitBucket
 import ftl.args.ArgsHelper.evaluateFilePath
 import ftl.args.ArgsHelper.mergeYmlMaps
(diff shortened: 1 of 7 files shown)


test_runner/src/main/kotlin/ftl/args/IosArgs.kt
```diff
@@ -64,11 +64,8 @@ class IosArgs(
     override val testShardChunks: List<List<String>> by lazy {
         val validTestMethods = Xctestrun.findTestNames(xctestrunFile)
```
bootstraponline: I think we want to add this:
https://github.com/TestArmada/flank/blob/master/test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt#L77

gtroshin:
Super excited to test it when it's built!

gtroshin:
Worked well! Thank you again!

PR #479, opened 2019-01-29
Improve invalid iOS regex error message

bootstraponline:


test_runner/src/main/kotlin/ftl/args/IosArgs.kt
```diff
@@ -159,15 +159,19 @@ ${listToString(testTargets)}
     }
 }
 
-fun filterTests(validTestMethods: List<String>, testTargets: List<String>): List<String> {
-    if (testTargets.isEmpty()) {
+fun filterTests(validTestMethods: List<String>, testTargetsRgx: List<String>): List<String> {
```
Macarse: mmm remember the list contains regex and NON regex like fullClassName/fullTestName.

bootstraponline: the code converts everything in that list to a regex `target.toRegex()`, right?

even non-regex becomes a regex.

Macarse: true :)

Issue #484, opened 2019-01-31
: breaks Windows


bootstraponline:
colon in folder name is not valid on Windows.

https://stackoverflow.com/questions/10386344/how-to-get-a-file-in-windows-with-a-colon-in-the-filename

convert from 

> YYYY-MM-DD_hh:mm:ss.ssssss_rrrr

to
> YYYY-MM-DD_hh-mm-ss.ssssss_rrrr

https://github.com/TestArmada/flank/blob/55c2d53fbfc5de1ef1fde3e8001980ba24b30a12/test_runner/src/main/kotlin/ftl/util/Utils.kt#L71-L75

bootstraponline:
Fixed by https://github.com/TestArmada/flank/pull/485

PR #485, opened 2019-02-01
Issue #484: Name with ':' breaks in windows

Sunil-plsr:
Fixes issue: 484


bootstraponline:
Thanks for the pull request! Merged.

PR #509, opened 2019-02-27
Auto publish to Maven

bootstraponline:


winterDroid:
Amazing 👍

test_runner/build.gradle.kts
```diff
@@ -11,6 +14,53 @@ plugins {
...
+        })
+    })
+}
+
+publishing {
+    publications {
+        register("mavenJava", MavenPublication::class) {
+            from(components["java"])
+
+            groupId = "flank"
```
runningcode: should the groupId include `testarmada` or some other unique identifier?

bootstraponline: I think flank is unique? I hesitate to use `testarmada` since I'm not sure that'll be true forever.

runningcode: Makes sense. Flank is an English word, I think its not the best to use english words as group ids. Usually it is a website identifier.

Anyways, the maven jar is working now! I really appreciate it!

PR #514, opened 2019-03-07
Fixes #490 Rename testShards to maxTestShards

### LINKED ISSUE #490: Rename testShards to maxTestShards
testShards defines an upper limit for the amount of test shards to use

miguelslemos:
This PR rename testShards parameter to maxTestShards

Fix #490

### DIFF
--- a/test_runner/flank.ios.yml
+++ b/test_runner/flank.ios.yml
@@ -70,10 +70,10 @@ flank:
 
   ## test shards - the amount of groups to split the test suite into
   ## set to -1 to use one shard per test. default: 1
-  # testShards: 1
+  # maxTestShards: 1
 
   ## shard time - the amount of time tests within a shard should take
-  ## when set to > 0, the shard count is dynamically set based on time up to the maxmimum limit defined by testShards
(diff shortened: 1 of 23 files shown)


miguelslemos:
This PR is breaking change, should I make it retro compatibility or just bump a major version?

bootstraponline:
Thanks for the pull request! I'll take a look at this today.

bootstraponline:
The code looks great.

To make it easy for people to upgrade, I'd like to dynamically rewrite the `flank.yml` file. If we don't have an auto-migration strategy then this will break immediately after merging since they run from the snapshot.

What do you think of that approach? I can add the code for this if the scope is too large.

The idea would be to read in the input YAML (`flank.yml` or `flank.ios.yml`), detect any old keys `testShards`, and then rename those keys. After the file is rewritten to disk then Flank would proceed normally.

bootstraponline:
I talked with the team and once https://github.com/TestArmada/flank/issues/515 is figured out then I'll merge this pull request. It's an open question how to best preserve backwards compatibility.

PR #522, opened 2019-03-15
Test retry

### LINKED ISSUE #456: Merge rerun results when --flaky-test-attempts is used
flakyTestAttempts runs tests multiple times. Currently any failures will be reported as a failure in Flank, even if the test passes on retry.

Fetch the test rerun results, and consider the full history of a test to determine if it's passed or not.

bootstraponline:
Fix #456 

Read in JUnitReport.xml and remove duplicate results when `flaky-test-attempts` is > 0. For each test keep first result. If next result for the same test is successful, keep last successful result.

### DIFF
--- a/test_runner/src/main/kotlin/ftl/reports/util/JUnitDedupe.kt
+++ b/test_runner/src/main/kotlin/ftl/reports/util/JUnitDedupe.kt
@@ -1,0 +1,29 @@
+package ftl.reports.util
+
+import ftl.reports.xml.model.JUnitTestCase
+import ftl.reports.xml.model.JUnitTestResult
+
+// Read in JUnitReport.xml and remove duplicate results when `flaky-test-attempts` is > 0
+// for each test `name="testFails" classname="com.example.app.ExampleUiTest"`
+// Keep first result. If next result for the same test is successful, keep last successful result.
+object JUnitDedupe {
+
+    private fun JUnitTestCase.key(): String {
+        return "${this.classname}#${this.name}"
+    }
+
+    fun modify(testResult: JUnitTestResult?) {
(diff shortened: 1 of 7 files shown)


gtroshin:
Is there information about the same tests (with the same name) in JUnitReport.xml been run on different devices and how does it handled if for some device they failed for some all passed?

bootstraponline:
> Is there information about the same tests (with the same name) in JUnitReport.xml been run on different devices and how does it handled if for some device they failed for some all passed?

That's a good point. I'll push a fix for multi-device.

valeraz:
From a quick look, this approach would probably break legitimate cases of duplicate test results in the final rollup (e.g. the use of repeatShards, unless you want to not allow the use of repeatShards with flakyTestAttempts). Deduping at the `shard/device` level seems like a safer approach.

bootstraponline:
I pushed a new commit that names `JUnitTestSuite` after the device string (`NexusLowRes-28-en-portrait`). That should resolve concerns around the same tests being deduplicated across multiple devices.

bootstraponline:
> unless you want to not allow the use of repeatShards with flakyTestAttempts

I think you can use repeatShards with flakyTestAttempts, however `flakyTestAttempts` will activate deduplication. I agree about device level dedupe, that should work now.

gtroshin:
I have a suite with several flaky tests and that new implementation showed it well! Everything is green. 
Will you keep it in snapshot until the release or there are any plans to unmerge it?

bootstraponline:
> Will you keep it in snapshot until the release or there are any plans to unmerge it?

I merged the PR. It'll continue to be in the snapshots. Thanks for testing!

PR #523, opened 2019-03-15
Local result dir

### LINKED ISSUE #500: Allow /results to be stored to different local path
Currently all of the Flank test results seem to be stored to /results

It would be nice if we had and optional parameter to set a different local folder to write the results to.

bootstraponline:
Fix #500


miguelslemos:
This PR contains commits related to #456. But local result dir implementation looks good to me.

bootstraponline:
I updated to base branch to avoid the unrelated commits.

PR #533, opened 2019-03-28
Update matrix file each time a matrix changes while polling

jschear:
When a Flank execution hangs (due to a broken test or an FTL outage) and is killed by a timeout (on Jenkins, Bitrise, etc.), it's nice to be able to find the FTL console link associated with each shard in order to debug the issue. These are normally printed to stdout at the end of a test run, but they might not be logged if the run is killed before completing.

Fortunately, Flank also persists a `matrix_ids.json` file to disk that contains the FTL link. This PR increases the frequency with which that file is updated, so that FTL links will make it into the file as soon as they're available.

I was hoping to add a test here but I was having trouble writing one that really covered this logic -- let me know if you have an idea for how to do that 😄

### DIFF
--- a/test_runner/src/main/kotlin/ftl/run/TestRunner.kt
+++ b/test_runner/src/main/kotlin/ftl/run/TestRunner.kt
@@ -259,7 +259,6 @@ object TestRunner {
     /** Synchronously poll all matrix ids until they complete. Returns true if test run passed. **/
     private fun pollMatrices(matrices: MatrixMap, args: IArgs) {
         println("PollMatrices")
-        val map = matrices.map
         val poll = matrices.map.values.filter {
             MatrixState.inProgress(it.state)
         }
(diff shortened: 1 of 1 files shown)


bootstraponline:
Thanks. The change looks good. There's not a great way to test this until FTL open sources their mock server. The current mock server we have doesn't support polling. We could update it to support polling but I'm hoping to use the official FTL mock server in the future.

bootstraponline:
```kotlin
if (MatrixState.completed(refreshedMatrix.state)) {
  break
}
```

that means we exit after a matrix reaches a final state. The write/update logic happens after that, and thus misses the final update to the the matrix_ids.json. I fixed this in: https://github.com/TestArmada/flank/pull/534

I found the bug by using flank on our [iOS Student app](https://github.com/instructure/canvas-ios) :)

jschear:
🤦‍♂️ nice find, sorry about that!

bootstraponline:
no problem, it's a gap in our testing coverage 😄 I will hopefully be improving that in the future.

PR #537, opened 2019-04-04
Add option to prevent uploading results for Smart Flank

elihart:
Hi, thanks for this great library! I noticed that https://github.com/TestArmada/flank/blob/master/docs/smart_flank.md mentioned the possible improvement to disable uploading smart flank results, so I implemented it.

I have recently added Flank to our test suite at Airbnb, and I'd love this support to allow us to run a subset of tests on some branches without overwriting the master set of results.

### DIFF
--- a/test_runner/flank.yml
+++ b/test_runner/flank.yml
@@ -107,6 +107,11 @@ flank:
   ## Google cloud storage path to store the JUnit XML results from the last run.
   # smart-flank-gcs-path: gs://tmp_flank/flank/test_app_android.xml
 
+  ## Whether the JUnit XML results should be uploaded for Smart Flank.
+  ## This can be disabled to prevent new results from overriding previous results.
+  ## Default: true
+  # smart-flank-upload-enabled: true
+
   ## Disables sharding. Useful for parameterized tests.
   # disable-sharding: false
 
--- a/test_runner/flank.yml
+++ b/test_runner/flank.yml
@@ -107,10 +107,9 @@ flank:
   ## Google cloud storage path to store the JUnit XML results from the last run.
   # smart-flank-gcs-path: gs://tmp_flank/flank/test_app_android.xml
 
-  ## Whether the JUnit XML results should be uploaded for Smart Flank.
-  ## This can be disabled to prevent new results from overriding previous results.
-  ## Default: true
-  # smart-flank-upload-enabled: true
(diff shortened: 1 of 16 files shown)


bootstraponline:
Thanks! That's definitely a feature we want to support.

bootstraponline:
If you run `./gradlew check` locally it should auto fix some of the lint issues.

elihart:
Thanks for the tip, fixed!

I was going off the contributing doc which says to use `./gradlew ktlintFormat`, which threw me off. Would be nice to update that.

bootstraponline:
ah, yeah we replaced ktlint with detekt. I'll update the doc

test_runner/flank.yml
```diff
@@ -107,6 +107,11 @@ flank:
   ## Google cloud storage path to store the JUnit XML results from the last run.
   # smart-flank-gcs-path: gs://tmp_flank/flank/test_app_android.xml
 
+  ## Whether the JUnit XML results should be uploaded for Smart Flank.
+  ## This can be disabled to prevent new results from overriding previous results.
+  ## Default: true
+  # smart-flank-upload-enabled: true
+
   ## Disables sharding. Useful for parameterized tests.
   # disable-sharding: false
```
bootstraponline: I think for consistency with `disable-sharding`, I'd like the flag renamed to:

`smart-flank-disable-upload`

then in the code we can check `args.smartFlankDisableUpload` instead of `!args.smartFlankUploadEnabled`.

The description can be written like this:
> Disables smart flank JUnit XML uploading. Useful for preventing timing data from being updated.

I'm happy to make both changes if you'd like.

elihart: sure thing, I'll do it

bootstraponline: Perfect, thanks. After that I'll merge the PR.

elihart:
All cleaned up, I think it should be good to go. Thanks for the quick review!

Any idea when you might be able to push a new release with this change? I'd love to get it into our app soon.

bootstraponline:
Awesome work! The Flank release process is fully automated. The snapshot jar is released after every merged PR. The stable releases are auto uploaded after I tag a version in git. I'll tag a `5.1.0` release after merging.

bootstraponline:
image

PR #542, opened 2019-04-10
Add additional-app-test-apks support

### LINKED ISSUE #511: Execute Additional Test apks
For projects that have many modules it's not uncommon to include some small number of android tests in a submodule. Flank is currently unable to run these tests because they are not packaged in the main app's test apk. 

In order to run these tests today, a user would need to write a script that passes each library to flank along with the main app's apk for the "app". This is doable, but it would be better if flank could handle this for us. 

I'd like to propose the following: 

1. A new `--module-apks` parameter in the Flank yaml. This is a `List<String> = emptyList()`
1. Modify `AndroidArgs` so that it's a data class, allowing us to use things like `copy`
1. `AndroidArgs` needs to be updated to allow the `app` parameter to be adjusted on the fly  
1. Update `AndroidTestRunner` to upload the new APKs in the same way the main app apk is uploaded today. 
1. Update `AndroidTestRunner` to iterate through the `--module-apks` and pass each to the `AndroidArgs` before creating a `GcAndroidTestMatrix` for the shards created in each additional apk. 

The changes to `AndroidTestRunner` would be something along these lines: 

    suspend fun runTests(androidArgs: AndroidArgs): MatrixMap = coroutineScope {
        val (stopwatch, runGcsPath) = GenericTestRunner.beforeRunTests(androidArgs)

        val jobs = arrayListOf<Deferred<TestMatrix>>()

        // Add all the app tests. 
        jobs.addAll(runTestsInternal(androidArgs, runGcsPath))

        // Add any submodule tests as well.
        val resolvedModuleUris = resolveSubmoduleApks(androidArgs, runGcsPath)
        resolvedModuleUris.forEach { apk ->
            // Submodules package everything they need in the test apk, we can use any random app apk
            val moduleArgs = androidArgs.copy(testApk = apk)

            jobs.addAll(runTestsInternal(moduleArgs, runGcsPath))
        }

        GenericTestRunner.afterRunTests(jobs.awaitAll(), runGcsPath, stopwatch, androidArgs)
    }

Things to consider.. 
1. I'm not really sure modifying the `AndroidArgs` like this is the best option, but not doing it this way requires that we also modify the sharding logic so that we can track which APK a given test is part of so that a shard does not span more than one apk. Since the sharding logic is built into the AndroidArgs, copying them and allowing it to re-run against a new APK seems to work. 

I'd like to work on this, so please let me know if anyone has better ideas on how this might be implemented. I've sort of hacked together the above solution - it's _very_ hacky but it does work and I could clean it up for a PR if it sounds good.

bootstraponline:
Fix #511

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,6 +307,12 @@ flank:
 
   ## Local folder to store the test result. Folder is DELETED before each run to ensure only artifacts from the new run are saved.
   # local-result-dir: flank
+    
+  ## Include additional app/test apk pairs in the run. If app is omitted, then the top level app is used for that pair.
+  # additional-app-test-apks:
+  #  - app: ../test_app/apks/app-debug.apk
+  #    test: ../test_app/apks/app-debug-androidTest.apk
+  #  - test: ../test_app/apks/app-debug-androidTest.apk
 ```
 
 ### Android code coverage
--- a/test_runner/src/main/kotlin/ftl/run/AndroidTestRunner.kt
+++ b/test_runner/src/main/kotlin/ftl/run/AndroidTestRunner.kt
@@ -2,7 +2,8 @@ package ftl.run
 
 import com.google.api.services.testing.model.TestMatrix
 import ftl.args.AndroidArgs
-import ftl.config.FtlConstants
+import ftl.args.AndroidTestShard
+import ftl.args.yml.AppTestPair
 import ftl.gc.GcAndroidDevice
 import ftl.gc.GcAndroidTestMatrix
 import ftl.gc.GcStorage
@@ -17,39 +18,46 @@ import kotlinx.coroutines.coroutineScope
 
 object AndroidTestRunner {
 
-    suspend fun runTests(androidArgs: AndroidArgs): MatrixMap = coroutineScope {
+    suspend fun runTests(androidArgs: AndroidArgs): Pair<MatrixMap, List<List<String>>> = coroutineScope {
         val (stopwatch, runGcsPath) = GenericTestRunner.beforeRunTests(androidArgs)
 
         // GcAndroidMatrix => GcAndroidTestMatrix
         // GcAndroidTestMatrix.execute() 3x retry => matrix id (string)
         val androidDeviceList = GcAndroidDevice.build(androidArgs.devices)
 
-        val apks = resolveApks(androidArgs, runGcsPath)
         val jobs = arrayListOf<Deferred<TestMatrix>>()
         val runCount = androidArgs.repeatTests
-        val shardCount = androidArgs.testShardChunks.size
         val shardCounter = ShardCounter()
         val history = GcToolResults.createToolResultsHistory(androidArgs)
+        val apks = resolveApks(androidArgs, runGcsPath)
+        val testShardChunks: MutableList<List<String>> = mutableListOf()
 
-        println(GenericTestRunner.beforeRunMessage(androidArgs))
-        repeat(runCount) {
-            repeat(shardCount) { testShardsIndex ->
-                jobs += async {
-                    GcAndroidTestMatrix.build(
-                        appApkGcsPath = apks.first,
-                        testApkGcsPath = apks.second,
-                        runGcsPath = runGcsPath,
-                        androidDeviceList = androidDeviceList,
-                        testShardsIndex = testShardsIndex,
-                        args = androidArgs,
-                        shardCounter = shardCounter,
-                        toolResultsHistory = history
-                    ).executeWithRetry()
+        apks.forEach { apk ->
+            testShardChunks += AndroidTestShard.getTestShardChunks(androidArgs, apk.test)
+        }
+
+        apks.forEach { apk ->
+            repeat(runCount) {
+                testShardChunks.forEach { testTargets ->
+                    jobs += async {
+                        GcAndroidTestMatrix.build(
+                            appApkGcsPath = apk.app ?: androidArgs.appApk,
+                            testApkGcsPath = apk.test,
+                            runGcsPath = runGcsPath,
+                            androidDeviceList = androidDeviceList,
+                            testTargets = testTargets,
+                            args = androidArgs,
+                            shardCounter = shardCounter,
+                            toolResultsHistory = history
+                        ).executeWithRetry()
+                    }
                 }
             }
         }
 
-        GenericTestRunner.afterRunTests(jobs.awaitAll(), runGcsPath, stopwatch, androidArgs)
+        println(GenericTestRunner.beforeRunMessage(androidArgs, testShardChunks))
+        val matrixMap = GenericTestRunner.afterRunTests(jobs.awaitAll(), runGcsPath, stopwatch, androidArgs)
+        matrixMap to testShardChunks
     }
 
     /**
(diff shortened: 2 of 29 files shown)


README.md
```diff
@@ -307,6 +307,12 @@ flank:
 
   ## Local folder to store the test result. Folder is DELETED before each run to ensure only artifacts from the new run are saved.
   # local-result-dir: flank
+    
+  ## Include additional app/test apk pairs in the run. If app is omitted, then the top level app is used for that pair.
+  # additional-app-test-apks:
+  #  - app: ../test_app/apks/app-debug.apk
+  #    test: ../test_app/apks/app-debug-androidTest.apk
```
doodla: Have you thought of making `test` a list, so you don't have a bunch of usages without `app`?

Something like

```yml

- app: ../app.apk
  test:
    - testapk1.apk
    - testapk2.apk

- app: app2.apk
  test:
     - testfinal.apk
```

bootstraponline: I like that idea.

doodla:
Did firebase add support for multiple test apks?

bootstraponline:
> Did firebase add support for multiple test apks?

Nope. We talked with the FTL team and this isn't a roadmap priority for them.

I think modules on Android are a good use case for this feature. I've been working with American Express to add support for this use case directly in Flank to unblock them.

doodla:
Haven't gone through the code, but are you merging all test apks into a single one?

bootstraponline:
> Haven't gone through the code, but are you merging all test apks into a single one?

Previously flank only allowed one test/app apk. In the new design, Flank accepts an array of test/app apks. The apks aren't merged. If you have one app and a dozen test apks for that single app, now you can get the results within a single flank run.

doodla:
Oh of course. You can just trigger multiple test lab runs, each with a different test APK.

carlospaulino:
I tested these changes locally, and it is not working as expected.

Most of the shard results are reporting no tests.

```xml
<?xml version='1.0' encoding='UTF-8' ?>
<testsuite name="" tests="1" failures="1" errors="0" skipped="0" time="0.544" timestamp="2019-04-23T17:16:05" hostname="localhost">
  <properties />
  <testcase name="" classname="no tests found" time="0.0">
    <failure>java.lang.ClassNotFoundException: Invalid name: no tests found
at java.lang.Class.classForName(Native Method)
at java.lang.Class.forName(Class.java:324)
at androidx.test.internal.runner.TestLoader.doCreateRunner(TestLoader.java:72)
at androidx.test.internal.runner.TestLoader.getRunnersFor(TestLoader.java:104)
at androidx.test.internal.runner.TestRequestBuilder.build(TestRequestBuilder.java:789)
at androidx.test.runner.AndroidJUnitRunner.buildRequest(AndroidJUnitRunner.java:543)
at androidx.test.runner.AndroidJUnitRunner.onStart(AndroidJUnitRunner.java:386)
at android.app.Instrumentation$InstrumentationThread.run(Instrumentation.java:1879)
</failure>
  </testcase>
</testsuite>
```

bootstraponline:
Thanks. I'll look into it

bootstraponline:
I'm unable to reproduce the failure using my apks. Pinged on Slack with more details on how to debug the problem.

carlospaulino:
@bootstraponline I found the problem. The Matrix is receiving test targets that are not part of the test apk.

See here: https://github.com/TestArmada/flank/pull/542/files#diff-1e35d5e13a5fd960dfea32faa4b6e6e0R41

test_runner/src/main/kotlin/ftl/run/AndroidTestRunner.kt
```diff
@@ -17,66 +18,70 @@ import kotlinx.coroutines.coroutineScope
...
-                        testShardsIndex = testShardsIndex,
-                        args = androidArgs,
-                        shardCounter = shardCounter,
-                        toolResultsHistory = history
-                    ).executeWithRetry()
+        apks.forEach { apk ->
+            repeat(runCount) {
+                testShardChunks.forEach { testTargets ->
+                    jobs += async {
+                        GcAndroidTestMatrix.build(
```
carlospaulino: The matrix is receiving test shards that are not part of the test apk. We should create each matrix out the shards of a given apk.

```kotlin
        apks.forEach { apk ->
            val apkTestShardChunks = AndroidTestShard.getTestShardChunks(androidArgs, apk.test)
            testShardChunks.addAll(apkTestShardChunks)
            repeat(runCount) {
                apkTestShardChunks.forEach { testTargets ->
                    jobs += async {
                        GcAndroidTestMatrix.build(
                            appApkGcsPath = apk.app ?: androidArgs.appApk,
                            testApkGcsPath = apk.test,
                            runGcsPath = runGcsPath,
                            androidDeviceList = androidDeviceList,
                            testTargets = testTargets,
                            args = androidArgs,
                            shardCounter = shardCounter,
                            toolResultsHistory = history
                        ).executeWithRetry()
                    }
                }
            }
        }
```

test_runner/src/main/kotlin/ftl/run/AndroidTestRunner.kt
```diff
@@ -17,66 +18,69 @@ import kotlinx.coroutines.coroutineScope
...
-                        shardCounter = shardCounter,
-                        toolResultsHistory = history
-                    ).executeWithRetry()
+        apks.forEach { apk ->
+            // ensure we only shard tests that are part of the test apk
+            val testShardChunks = AndroidTestShard.getTestShardChunks(androidArgs, apk.test)
+            allTestShardChunks += testShardChunks
+            repeat(runCount) {
+                testShardChunks.forEach { testTargets ->
+                    jobs += async {
```
carlospaulino: We should probably specify a `Dispatcher` to run things in parallel because is inheriting the context from `runBlocking`

test_runner/src/main/kotlin/ftl/run/AndroidTestRunner.kt
```diff
@@ -17,66 +18,69 @@ import kotlinx.coroutines.coroutineScope
...
 
-        val appApkGcsPath = async {
-            if (!args.appApk.startsWith(FtlConstants.GCS_PREFIX)) {
-                GcStorage.uploadAppApk(args, gcsBucket, runGcsPath)
-            } else {
-                args.appApk
-            }
-        }
+        appTestApks.forEach { apks ->
+            val appApkGcsPath = async { GcStorage.upload(apks.app ?: args.appApk, gcsBucket, runGcsPath) }
```
carlospaulino: Same here we should specify a `Dispatcher`

test_runner/src/main/kotlin/ftl/run/AndroidTestRunner.kt
bootstraponline: Fixed

test_runner/src/main/kotlin/ftl/run/AndroidTestRunner.kt
bootstraponline: Fixed

test_runner/src/main/kotlin/ftl/run/AndroidTestRunner.kt
bootstraponline: Fixed

README.md
bootstraponline: I prototyped adding list support, in particular this complicated the code when parsing from CLI. 

- https://github.com/remkop/picocli/issues/677

I think for the first version of his feature, the more verbose format is easier to maintain. The test run loop also becomes more complex as we have yet another list to iterate through.

```yaml
- app: ../app.apk
  test: test1.apk
- app: ../app.apk
  test: test2.apk
```

I'm open to revisiting this based on feedback in the future.

PR #560, opened 2019-05-27
Add --test-runner-class

### LINKED ISSUE #459: Add support for --test-runner-class
Can we support https://cloud.google.com/sdk/gcloud/reference/firebase/test/android/run#--test-runner-class ?

jschear:
Adds the `--test-runner-class` parameter (which is only applicable for the Android run command): https://cloud.google.com/sdk/gcloud/reference/firebase/test/android/run#--test-runner-class

Closes #459.


bootstraponline:
The code looks perfect! I tested this on my app and the feature works as expected.

PR #589, opened 2019-08-13
fix java.lang.NumberFormatException: empty String

vfadc:
On my default local `"%.3f".format()` return string like 1,23 instead of 1.23.
So, the `clean()` extention removes the comma and it causes the acc var to get "infinity", which becomes `""` after `clean()` and cause `NumberFormatException`

If I run the unit tests locally it fails: 
```ftl.reports.xml.JUnitXmlTest > merge android FAILED
expected: 6.148
but was : 6,148
    at ftl.reports.xml.JUnitXmlTest.merge android(JUnitXmlTest.kt:51)
```
https://stackoverflow.com/questions/43963633/java-double-to-string-format

### DIFF
--- a/test_runner/src/main/kotlin/ftl/reports/xml/model/JUnitTestSuite.kt
+++ b/test_runner/src/main/kotlin/ftl/reports/xml/model/JUnitTestSuite.kt
@@ -2,6 +2,7 @@ package ftl.reports.xml.model
 
 import com.fasterxml.jackson.annotation.JsonInclude
 import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlProperty
+import java.util.Locale
 
 data class JUnitTestSuite(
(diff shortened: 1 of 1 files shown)


bootstraponline:
Interesting. I didn't realize the formatting was locale specific. That explains why I wasn't able to reproduce the problem. Thanks for the pull request!

The CLA bot wants you to update the commit to match your GitHub email.
- https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user

bootstraponline:
I think you have to squash merge so the PR contains a total of one commit. Then the commit with the invalid author info will be gone from the history. Sorry it's kind of a pain.

bootstraponline:
Merged! Thanks for tracking down the bug and fixing it.

PR #597, opened 2019-09-10
FTL is now sometimes adding a `testLabExecutionId` on <testsuite

yogurtearl:
We started getting some failures because some of the junit xml's coming back from FTL had an `testLabExecutionId` attribute on `<testsuite`.

This adds the property to the model class so it can deserialize correctly.

### DIFF
--- a/test_runner/src/main/kotlin/ftl/reports/xml/model/JUnitTestSuite.kt
+++ b/test_runner/src/main/kotlin/ftl/reports/xml/model/JUnitTestSuite.kt
@@ -31,6 +31,10 @@ data class JUnitTestSuite(
     @JacksonXmlProperty(isAttribute = true)
     val hostname: String, // String.
 
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    @JacksonXmlProperty(isAttribute = true)
+    val testLabExecutionId: String?, // String.
+
     @JacksonXmlProperty(localName = "testcase")
     var testcases: MutableCollection<JUnitTestCase>?,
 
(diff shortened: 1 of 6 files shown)


bootstraponline:
FTL is working on a rollback to unbreak Flank users.

Thanks for the pull request. I'd like a junit test before merging.

narenkmanoharan:
Also making the JacksonXmlModule configured with `DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES` to be `false` might address the root of the problem and make sure that this does not happen again.

mohitagrawal:
@bootstraponline - How can we track FTL updates? Would you be posting once Firebase fixes the issue?

bootstraponline:
Let's make sure we agree on what the root problem is. There's a software development ecosystem built on parsing the FTL JUnit XML reports. I can't have the format changing silently in production. 

I'd expect some type of heads up that the XML will be changing. I know bitrise would also be interested as they're building a dashboard based on parsing the XML.

If FTL is going to continue to make silent changes, then setting fail on unknown would be a reasonable work around. At the moment, I consider silent changes a legitimate bug that impacts multiple downstream businesses and tools.

bootstraponline:
> - How can we track FTL updates? 

That's a great question for the FTL team. There's currently no way to track FTL updates for external customers.

> Would you be posting once Firebase fixes the issue?

Yep, I'll be posting an update in the Flank release notes. I'll also post in the #flank channel.

yogurtearl:
@bootstraponline Adjusted tests to verify that `testLabExecutionId` is parsed if present.

bootstraponline:
The new commit looks good. CLA bot is unhappy.

yogurtearl:
Fixed CLA issue. :)

bootstraponline:
Awesome. Merged.

PR #615, opened 2019-10-29
Add --keep-file-path

### LINKED ISSUE #611: Option to preserve directory paths
For Android, some users have requested to preserve the original directory paths. Currently Flank downloads files to a flattened directory which requires unique file names.

I'm looking for feedback on this as it seems like an edge case that's not super useful. Files should be uniquely named, especially images.

tahirhajizada:
Fix #611

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,6 +311,10 @@ flank:
 
   ## Local folder to store the test result. Folder is DELETED before each run to ensure only artifacts from the new run are saved.
   # local-result-dir: flank
+
+  ## Downloaded files preserves the original path of file. Required when file names are not unique.
+  ## Default: false  
+  # preserve-original-path: false
     
   ## Include additional app/test apk pairs in the run. If app is omitted, then the top level app is used for that pair.
   # additional-app-test-apks:
--- a/README.md
+++ b/README.md
@@ -312,9 +312,9 @@ flank:
   ## Local folder to store the test result. Folder is DELETED before each run to ensure only artifacts from the new run are saved.
   # local-result-dir: flank
 
-  ## Downloaded files preserves the original path of file. Required when file names are not unique.
+  ## Keeps the full path of downloaded files. Required when file names are not unique.
   ## Default: false  
-  # preserve-original-path: false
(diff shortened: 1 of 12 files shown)


bootstraponline:
Thanks for submitting a pull request! I'll take a look soon.

README.md
```diff
@@ -311,6 +311,10 @@ flank:
 
   ## Local folder to store the test result. Folder is DELETED before each run to ensure only artifacts from the new run are saved.
   # local-result-dir: flank
+
+  ## Downloaded files preserves the original path of file. Required when file names are not unique.
+  ## Default: false  
+  # preserve-original-path: false
```
bootstraponline: Thanks for working on this! What do you think of `keep-file-path` for the key name?

test_runner/src/main/kotlin/ftl/util/ObjPath.kt
```diff
@@ -16,12 +17,16 @@ data class ObjPath(
             val objName = parsed.getName(0).toString()
             val shardName = parsed.getName(1).toString()
             val deviceName = parsed.getName(2).toString()
+            val filePathName = if (parsed.nameCount > 4) {
+                parsed.subpath(3, parsed.nameCount - 1).toString()
```
bootstraponline: What do you think of `parsed.parent.drop(3).joinToString("/")` as a way to simplify the `parsed.subpath` logic?

bootstraponline:
I'm happy to make the suggested changes if that'd be easier.

tahirhajizada:
I am ok with your suggestions. Makes sense. It would be nice if you apply them or I can do after 9-10 hours :)

bootstraponline:
Sounds good. I'll apply the fixes in ~4 hours after work.

bootstraponline:
I updated, tested, and merged your pull request! Nice work.

I released Flank v8.1.0 with the feature and updated the change log.

Thanks for contributing.

tahirhajizada:
Thank you for the changes and release. Welcome.

PR #621, opened 2019-12-01
Sharding

bootstraponline:


test_runner/src/main/kotlin/ftl/shard/Shard.kt
```diff
@@ -136,7 +136,8 @@ object Shard {
         testcases.sortByDescending { it.time }
 
         testcases.forEach { testMethod ->
-            val shard = shards.first()
+            // num_shards must be > 1, and <= 50
```
runningcode: just for clarification. num shards must be less than 50 or number of tests per shard must be less than 50?

bootstraponline: > The number of shards should be less than the total number of test cases. The number of shards specified must be >= 1 and <= 50.
>
> https://cloud.google.com/sdk/gcloud/reference/alpha/firebase/test/android/run

I think (hopefully) there's no limit on the number of tests per shard.

bootstraponline: I think this is >= 1, and <= 50

test_runner/src/main/kotlin/ftl/shard/Shard.kt
```diff
@@ -136,7 +136,8 @@ object Shard {
...
+            val shard = shards.first { it.testMethods.size + args.testTargetsAlwaysRun.size < 50 }
```
jan-gogo: There is a problem here. According to doc the number of shards should be less than 50, but this condition prevents from putting more than 50 test methods per shard so probably it's not what we want.

jan-gogo:
Hi, a have added some additional apks for specific test cases. Currently we are checking if new sharding is working properly on android, also with num-flaky-test-attempts flag on. And if everything will be ok I guess we can merge that. 
Additionally the output directory structure for shards will change slightly, because now gc API are handling shards and can returns correct directories in output.

bootstraponline:

bootstraponline:
recheckcla

jan-gogo:

pawelpasterz:

pawelpasterz:

bootstraponline:
@pawelpasterz hey, the code looks great. Let's move all the apks into https://github.com/Flank/test_artifacts

We don't want large binary files in the repo increasing the repo size. To solve for this, I publish the artifact bundle saparately which is downloaded as part of the test execution process.

For now, delete the apks and I can merge the PR. As a follow up, we can add the APKs to the test_artifacts repo and I'll publish a new test bundle.

jan-gogo:
@bootstraponline @pawelpasterz I have removed test apks but there is still one bug during test results fetch related to output directory, so we cannot merge until it will be resolved. Also i have 404 error when i tried to upload test apks in to https://github.com/Flank/test_artifacts. I need some time to investigate it.

bootstraponline:
@jan-gogo I added you to `Flank/test_artifacts` with write access. Let me know when this PR is ready to merge.

bootstraponline:
@pawelpasterz Thank you so much! Awesome work. I'll merge this.

Issue #632, opened 2020-01-28
Default test run time is too low


RainNapper:
### Context
- Enable smart flank
- Set a reasonable test run timeout (e.g. 5m, not 1h)
- Set a shard time below the timeout (e.g. 2m)
- Add a bunch of tests to the test suite

When new tests are discovered (miss in the smart flank cache), [they default to a 10.0 sec runtime.](https://github.com/Flank/flank/blob/d872f9514616b6c9cb137fa9c7215bde76b4cc54/test_runner/src/main/kotlin/ftl/shard/Shard.kt#L69)

### Problem
This presents an issue when too many tests get put into a single shard.
For example, say your tests on average run in 60 seconds. A shard time of 2m should mean you  will have about 2 tests per shard, with 3 minutes of buffer time in case your tests run longer than usual.

If you add 12 tests to your suite, all 12 of these tests will not be found my smart flank, and default to 10 seconds. It is very possible that 12 of the new tests will be placed in a single shard (12 * 10 = 120 = shard time). This now puts your expected run-time at 12 minutes, which will very likely time-out the suite.

To make matters worse, Firebase gives fairly cryptic errors in this scenario:
A few examples:
`Test run failed to complete. Expected 10 tests, received 4`
`java.lang.IllegalStateException: UiAutomation not connected!`
`java.lang.RuntimeException: androidx.test.services.shellexecutor.ClientNotConnected`

### Proposal
Option 1 - Make the 10 second default significantly higher (2-5m)
Having an over-estimated default means that we err on the side of too many shards rather than too few. This effectively trades cost for stability. Enabling smart flank will mitigate this cost.

Option 2 - Add an input param to specify this default, e.g. `--test-run-time-estimate`, to allow users  to choose if they want to prefer cost vs stability with a high vs low value.

pawelpasterz:
Resolved by #635

PR #635, opened 2020-02-04
Pass test time estimate arg to stabilize smart flank without cached results

RainNapper:
Context: https://github.com/Flank/flank/issues/632

Ran into some local issues running all of the tests (stacktrace below[1]), but I think I covered most if not all of the places that this change touches.

Please leave comments on what I missed, or what I need to get a local build working. I can run unit tests in Android Studio, but hit some issues[2] running `./gradlew build`

[1]
```
java.lang.VerifyError: Stack map does not match the one at exception handler 77
Exception Details:
  Location:
    com/fasterxml/jackson/databind/deser/std/StdDeserializer._parseDate(Lcom/fasterxml/jackson/core/JsonParser;Lcom/fasterxml/jackson/databind/DeserializationContext;)Ljava/util/Date; @77: astore
  Reason:
    Type 'com/fasterxml/jackson/core/JsonParseException' (current frame, stack[0]) is not assignable to 'com/fasterxml/jackson/core/exc/StreamReadException' (stack map, stack[0])
```

[2]
```
ftl.args.GcloudYmlTest > initializationError FAILED
    java.lang.NoClassDefFoundError: Could not initialize class ftl.test.util.FlankTestRunner
```

### DIFF
--- a/test_runner/src/main/kotlin/ftl/shard/Shard.kt
+++ b/test_runner/src/main/kotlin/ftl/shard/Shard.kt
@@ -1,5 +1,6 @@
 package ftl.shard
 
+import com.google.common.annotations.VisibleForTesting
 import ftl.args.AndroidArgs
 import ftl.args.IArgs
 import ftl.args.IosArgs
@@ -44,6 +45,9 @@ class com.foo.ClassName#testMethodToSkip
 */
 
 object Shard {
+    // When a test does not have previous results to reference, fall back to this run time.
+    @VisibleForTesting
+    const val DEFAULT_TEST_TIME_SEC: Double = 120.0
 
     private fun JUnitTestCase.androidKey(): String {
         return "class $classname#$name"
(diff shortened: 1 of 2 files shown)


bootstraponline:
Thank you for the detailed pull request and issue! I will take a look when I have some time.

bootstraponline:
re: #632 

> Option 1 - Make the 10 second default significantly higher (2-5m)
Having an over-estimated default means that we err on the side of too many shards rather than too few. This effectively trades cost for stability. Enabling smart flank will mitigate this cost.

I prefer option 1 over introducing another flag. 10s was randomly chosen since it seemed to work well. Updating the default to something like 2m probably makes the most sense.

Flank already has a ton of config options. I am hesitant to add more unless we truly have a need.

RainNapper:
> re: #632
> 
> > Option 1 - Make the 10 second default significantly higher (2-5m)
> > Having an over-estimated default means that we err on the side of too many shards rather than too few. This effectively trades cost for stability. Enabling smart flank will mitigate this cost.
> 
> I prefer option 1 over introducing another flag. 10s was randomly chosen since it seemed to work well. Updating the default to something like 2m probably makes the most sense.
> 
> Flank already has a ton of config options. I am hesitant to add more unless we truly have a need.

That sounds good to me. Simplifies the change too. I went with 2m.

bootstraponline:
Thanks! That looks great.

I have to setup the bitrise jobs + CLA bot since we moved to the new Flank org. After that I'll merge this PR.

bootstraponline:
Hi, sorry for the delay! We now have 2 full time resources working on Flank. 😄 

I've merged the pull request.

PR #641, opened 2020-02-25
Make Deps.kt the source of truth for dependencies

doodla:
Versions haven't been modified, so if this builds and passes tests, we should be good to go.

### DIFF
--- a/test_runner/build.gradle.kts
+++ b/test_runner/build.gradle.kts
@@ -128,79 +128,49 @@ tasks.withType<Test> {
 }
 
 dependencies {
+    implementation(Libs.BUGSNAG)
+
+    implementation(Libs.DD_PLIST)
+    implementation(Libs.DEX_TEST_PARSER)
+
+    implementation(Libs.GSON)
+
+    implementation(Libs.JACKSON_DATABIND)
+    implementation(Libs.JACKSON_KOTLIN)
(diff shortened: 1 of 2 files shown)


pawelpasterz:
Hey @doodla could you change commits message to be more explanatory? Thanks :)

doodla:

doodla:
> Hey @doodla could you change commits message to be more explanatory? Thanks :)

It's too much work, imo. The PR title should be sufficient. I usually don't bother with detailed commit messages in repositories where the pull requests are merged by squashing.

pawelpasterz:
> > Hey @doodla could you change commits message to be more explanatory? Thanks :)
> 
> It's too much work, imo. The PR title should be sufficient. I usually don't bother with detailed commit messages in repositories where the pull requests are merged by squashing.

Got it, thanks for answer

doodla:
> Got it, thanks for answer

For some more context, my thinking is that Pull requests are atomic commits ( **squash merge** ) into the `master` branch. Because of that, it makes little sense to document exactly how you arrived there if the PR author doesn't feel the need to. In my case, the majority of the change is likely the first commit with all others being modifications based on what CI spits out. `f` is an abbreviation of `fix` for me.

If a pull request can be divided into multiple discrete chunks, each chunk can be a separate pull request to decrease the size and make the reviewers' lives easy.

PR #643, opened 2020-02-27
Add --dry option to android run & ios run

### LINKED ISSUE #447: Dry run mode for flank
Enable toggling the mock server to enable a "dry run."

`flank android run --dry`

jan-gogo:
Fixes #447

## Checklist

- [x] Documented
- [x] Unit tested


jan-gogo:
> LGTM -- thanks for contribution!
> BTW, did you check if server mock paths are not outdated? Or should we do it in another issue?

If server pathes are outdated it wouldn't work, also tests shouldn't pass. So it's ok I guess.

PR #644, opened 2020-02-28
Add default flank settings to optimize performance

### LINKED ISSUE #455: Fast/stable by default
Flank should use the most optimal config by default. Currently a few options (such as Google account login) are on by default which slows down testing.

* [ ] Nexuslowres can make executions up to 4x faster due to smaller screen size.
* [ ] Disable video (results bucket significant smaller, skips processing time of generating video)
* [ ] Disable auto login (saves 30s - 1m)
* [ ] Disable performance metrics. (saves processing time)

pawelpasterz:
Fixes #455 

# Chnages
* Add `FlankDefaults` class to preserve defaults Flank settings
* Update unit tests
* Update documentation

## Checklist

- [x] Documented
- [x] Unit tested


test_runner/src/main/kotlin/ftl/config/FlankDefaults.kt
```diff
@@ -0,0 +1,12 @@
+package ftl.config
+
+/**
+ * Flank specific default settings.
+ * Values should assure best (quickest) test performance.
+ * Each of value can be overwritten by config *.yml file
+ */
+object FlankDefaults {
```
jan-gogo: I'm not sure if hiding default values behind constants was necessary. Any of those constant are used only once and they are not telling more than members which they are assigned. But we have one file more to check, if we want to know defaults.

pawelpasterz: In general case I agree. In this particular one I wanted to point explicitly that those are not default, `gcloud` values, and what those values mean (ex when I see `DISABLE_VIDEO_RECORDING` I would conclude that Flank's default setting is to have this disabled not being particularly interested if this is false, -1 or sth else). Probably `enum` based would be better but this will require bigger refactor that's why I decided to proceed with this solution.

 I could also just change `true -> false` and leave additional comment why it is implemented that way.

TBH both solutions are fine for me. Let me know what you think, thanks!

jan-gogo: That was only suggestion so you can leave it as is. Your arguments are reasonable, it's ok for me.

PR #648, opened 2020-03-03
Flank should not skip tests annotated with @Ignored

### LINKED ISSUE #629: [Feature request] Include tests marked with @org.junit.Ignore in JUnitReport.xml
We often use `@org.junit.Ignore` annotation to mark tests we plan to fix later. When we are doing multiple changes in the specific area and we predict that some other changes will happen very soon and will break the tests again, for us it's better to mark the test as ignored for some time and go back to it when all the changes are introduced.

Such tests are (correctly) not taken into account when running test suite on FTL. However, we'd like to include ignored tests in `JUnitReport.xml` so that our CI environment (when the XML report gets parsed) can present ignored tests. In such way we can track which tests were ignored in the given test run.

The end result could be similar to how regular Android unit tests report looks like - sometimes we also have some unit tests marked with `@Ignore`, they land in the XML report and are correctly parsed by the CI.

pawelpasterz:
Fixes #629 

## Changes
* tests annotated with `@Ignore` are included in test matrix by default (TL skips them anyway)
* user still can filter tests with `@Ignore` with `notAnnotation org.junit.Ignore`
* Ignored tests now have time = 0 (regardless previous results)
* update and refactor unit tests

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/test_runner/src/main/kotlin/ftl/args/AndroidTestShard.kt
+++ b/test_runner/src/main/kotlin/ftl/args/AndroidTestShard.kt
@@ -3,8 +3,10 @@ package ftl.args
 import com.linkedin.dex.parser.DexParser
 import com.linkedin.dex.parser.TestMethod
 import ftl.config.FtlConstants
+import ftl.filter.TestFilter
 import ftl.filter.TestFilters
 import ftl.gc.GcStorage
+import ftl.util.FlankTestMethod
 import ftl.util.Utils
 import kotlinx.coroutines.runBlocking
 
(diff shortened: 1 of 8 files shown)


bootstraponline:
If we know a test is ignored, we probably want to assign it a runtime of 0 when packing shards by time.

pawelpasterz:
> If we know a test is ignored, we probably want to assign it a runtime of 0 when packing shards by time.

Nice catch! Thanks for pointing this out

PR #657, opened 2020-03-08
Fix execution hangs

### LINKED ISSUE #649: Execution hangs when exceptions are thrown
When passing in a test filter that is improperly formed, Flank will throw an exception but will not terminate. When flank is run in CI, this causes unnecessary timeouts while the scripts are waiting for the command to finish.

Example error
`java.lang.IllegalArgumentException: Invalid argument: com.package.Class#testMethod
	at ftl.filter.TestFilters.parseSingleFilter(TestFilters.kt:100)
	at ftl.filter.TestFilters.access$parseSingleFilter(TestFilters.kt:31)
	at ftl.filter.TestFilters$fromTestTargets$parsedFilters$2.invoke(TestFilters.kt:85)
	at ftl.filter.TestFilters$fromTestTargets$parsedFilters$2.invoke(TestFilters.kt:31)
	at kotlin.sequences.TransformingSequence$iterator$1.next(Sequences.kt:172)
	at kotlin.sequences.SequencesKt___SequencesKt.toCollection(_Sequences.kt:702)
	at kotlin.sequences.SequencesKt___SequencesKt.toMutableList(_Sequences.kt:732)
	at kotlin.sequences.SequencesKt___SequencesKt.toList(_Sequences.kt:723)
	at ftl.filter.TestFilters.fromTestTargets(TestFilters.kt:86)
	at ftl.args.AndroidTestShard.getTestMethods(AndroidTestShard.kt:32)
	at ftl.args.AndroidTestShard.getTestShardChunks(AndroidTestShard.kt:25)
	at ftl.run.AndroidTestRunner$runTests$2.invokeSuspend(AndroidTestRunner.kt:38)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.ResumeModeKt.resumeUninterceptedMode(ResumeMode.kt:45)
	at kotlinx.coroutines.internal.ScopeCoroutine.afterCompletionInternal(Scopes.kt:32)
	at kotlinx.coroutines.JobSupport.completeStateFinalization(JobSupport.kt:311)
	at kotlinx.coroutines.JobSupport.tryFinalizeFinishingState(JobSupport.kt:237)
	at kotlinx.coroutines.JobSupport.tryMakeCompletingSlowPath(JobSupport.kt:851)
	at kotlinx.coroutines.JobSupport.tryMakeCompleting(JobSupport.kt:813)
	at kotlinx.coroutines.JobSupport.makeCompletingOnce$kotlinx_coroutines_core(JobSupport.kt:789)
	at kotlinx.coroutines.AbstractCoroutine.resumeWith(AbstractCoroutine.kt:111)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:46)
	at kotlinx.coroutines.DispatchedTask.run(Dispatched.kt:241)
	at kotlinx.coroutines.EventLoopImplBase.processNextEvent(EventLoop.kt:116)
	at kotlinx.coroutines.BlockingCoroutine.joinBlocking(Builders.kt:79)
	at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking(Builders.kt:54)
	at kotlinx.coroutines.BuildersKt.runBlocking(Unknown Source)
	at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking$default(Builders.kt:36)
	at kotlinx.coroutines.BuildersKt.runBlocking$default(Unknown Source)
	at ftl.cli.firebase.test.android.AndroidRunCommand.run(AndroidRunCommand.kt:49)
	at picocli.CommandLine.executeUserObject(CommandLine.java:1687)
	at picocli.CommandLine.access$900(CommandLine.java:146)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2059)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2026)
	at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:1893)
	at picocli.CommandLine.execute(CommandLine.java:1822)
	at ftl.Main$Companion.main(Main.kt:41)
	at ftl.Main.main(Main.kt)`

pawelpasterz:
Fixes #649 

This PR contains commit [2b92fec](https://github.com/Flank/flank/commit/2b92fec60259256975a58757ecf927ffb4e91a38) which is only small refactor which touches quite big number of files. Please, skip this commit when reviewing.

## Changes:
* wrap `CommandLine(Main()).execute(*args)` with `exitProcess(...)` which closes jvm (this is workaround for bugsnag bug related to thread pool executor"
* add unit tests

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/test_runner/src/main/kotlin/ftl/Main.kt
+++ b/test_runner/src/main/kotlin/ftl/Main.kt
@@ -9,8 +9,8 @@ import ftl.cli.firebase.test.IosCommand
 import ftl.log.setDebugLogging
 import ftl.util.readRevision
 import ftl.util.readVersion
+import ftl.util.jvmHangingSafe
 import picocli.CommandLine
-import kotlin.system.exitProcess
 
 @CommandLine.Command(
     name = "flank.jar\n",
@@ -49,12 +49,7 @@ class Main : Runnable {
             // BugSnag opens a non-daemon thread which will keep the JVM process alive.
             // Flank must invoke exitProcess to exit cleanly.
             // https://github.com/bugsnag/bugsnag-java/issues/151
-            try {
-                exitProcess(CommandLine(Main()).execute(*args))
-            } catch (t: Throwable) {
-                t.printStackTrace()
-                exitProcess(CommandLine.ExitCode.SOFTWARE)
-            }
+            jvmHangingSafe { CommandLine(Main()).execute(*args) }
         }
     }
 }
--- a/test_runner/src/main/kotlin/ftl/Main.kt
+++ b/test_runner/src/main/kotlin/ftl/Main.kt
@@ -7,8 +7,8 @@ import ftl.cli.firebase.RefreshCommand
 import ftl.cli.firebase.test.AndroidCommand
 import ftl.cli.firebase.test.IosCommand
 import ftl.log.setDebugLogging
-import ftl.util.Utils.readRevision
-import ftl.util.Utils.readVersion
+import ftl.util.readRevision
+import ftl.util.readVersion
 import picocli.CommandLine
 
 @CommandLine.Command(
--- a/test_runner/src/main/kotlin/ftl/Main.kt
+++ b/test_runner/src/main/kotlin/ftl/Main.kt
@@ -46,6 +46,9 @@ class Main : Runnable {
     companion object {
         @JvmStatic
         fun main(args: Array<String>) {
+            // BugSnag opens a non-daemon thread which will keep the JVM process alive.
+            // Flank must invoke exitProcess to exit cleanly.
+            // https://github.com/bugsnag/bugsnag-java/issues/151
             exitProcess(CommandLine(Main()).execute(*args))
         }
     }
--- a/test_runner/src/main/kotlin/ftl/Main.kt
+++ b/test_runner/src/main/kotlin/ftl/Main.kt
@@ -10,6 +10,7 @@ import ftl.log.setDebugLogging
 import ftl.util.readRevision
 import ftl.util.readVersion
 import picocli.CommandLine
+import kotlin.system.exitProcess
 
 @CommandLine.Command(
     name = "flank.jar\n",
@@ -45,7 +46,7 @@ class Main : Runnable {
     companion object {
         @JvmStatic
         fun main(args: Array<String>) {
-            CommandLine(Main()).execute(*args)
+            exitProcess(CommandLine(Main()).execute(*args))
         }
     }
 }
--- a/test_runner/src/main/kotlin/ftl/Main.kt
+++ b/test_runner/src/main/kotlin/ftl/Main.kt
@@ -49,7 +49,12 @@ class Main : Runnable {
             // BugSnag opens a non-daemon thread which will keep the JVM process alive.
             // Flank must invoke exitProcess to exit cleanly.
             // https://github.com/bugsnag/bugsnag-java/issues/151
-            exitProcess(CommandLine(Main()).execute(*args))
+            try {
+                exitProcess(CommandLine(Main()).execute(*args))
+            } catch (t: Throwable) {
+                t.printStackTrace()
+                exitProcess(CommandLine.ExitCode.SOFTWARE)
+            }
         }
     }
 }
--- a/release_notes.md
+++ b/release_notes.md
@@ -1,7 +1,8 @@
 ## next (unreleased)
 
-- [#654](https://github.com/Flank/flank/pull/654) Fix test filters when using both notPackage and notClass. ([jan-gogo](https://github.com/jan-gogo)) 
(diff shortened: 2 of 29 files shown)


test_runner/src/main/kotlin/ftl/Main.kt
```diff
@@ -49,7 +49,12 @@ class Main : Runnable {
             // BugSnag opens a non-daemon thread which will keep the JVM process alive.
             // Flank must invoke exitProcess to exit cleanly.
             // https://github.com/bugsnag/bugsnag-java/issues/151
-            exitProcess(CommandLine(Main()).execute(*args))
+            try {
```
bootstraponline: Let's document in the code why we need `try/catch` since this isn't obvious.

bootstraponline: Also is there any way for us to add a JUnit test that will fail if we delete try/catch? I'm worried about "simplifying" this logic in the future and the test suite will still pass.

remkop: To test, create a "malicious" command that starts a non-daemon thread, and then throws an Error.

The only issue is that this test will hang if the try/catch is removed. Instead of hanging, you could run the test itself in a separate thread, with a timer that ensures that the test completes within some timespan or else the test fails.

Something like this?

```java
import org.junit.jupiter.api.Test;
import picocli.CommandLine;
import picocli.CommandLine.Command;
import picocli.CommandLine.Unmatched;

import java.util.List;
import java.util.concurrent.CountDownLatch;
import java.util.concurrent.atomic.AtomicBoolean;

import static org.junit.jupiter.api.Assertions.assertEquals;
import static org.junit.jupiter.api.Assertions.assertTrue;

public class TimeoutTest {
    @Command(name = "whosbad")
    static class Malicious implements Runnable {
        static CountDownLatch throwingError = new CountDownLatch(1);
        @Unmatched
        List<String> ignoredArgs;

        @Override
        public void run() {
            CountDownLatch isThreadRunning = new CountDownLatch(1);
            Thread forever = new Thread("forever") {
                public void run() {
                    try {
                        isThreadRunning.countDown();
                        Thread.sleep(Long.MAX_VALUE);
                    } catch (InterruptedException ignored) { }
                }
            };
            forever.setDaemon(false);
            forever.start();
            try {
                isThreadRunning.await();
            } catch (InterruptedException ignored) { }

            throwingError.countDown();
            throw new Error("Killing the calling thread...");
        }
    }

    static Integer exitedStatus = null;
    static void simulatedSystemExit(int status) {
        exitedStatus = status;
    }

    static class MyApp {
        public static void main(String... args) {
            try {
                simulatedSystemExit(new CommandLine(new Malicious()).execute(args));
            } catch (Throwable t) {
                t.printStackTrace();
                simulatedSystemExit(CommandLine.ExitCode.SOFTWARE);
            }
        }
    }

    @Test
    public void testSafeAgainstMaliciousActors() throws Exception {
        AtomicBoolean completed = new AtomicBoolean(false);
        Thread simulatedMain = new Thread("simulated-main") {
            public void run() {
                MyApp.main("some", "args");
                completed.set(true);
            }
        };
        simulatedMain.start();

        Malicious.throwingError.await();
        long TIMEOUT = 3 * 1000L;
        simulatedMain.join(TIMEOUT); // in this test, this returns immediately because the simulated-main thread was killed by the Error

        assertTrue(completed.get(), "Our simulated main thread should have completed but instead it hung...");
        assertEquals(CommandLine.ExitCode.SOFTWARE, exitedStatus);
    }
}
```

bootstraponline: That's awesome. Thanks for all your help with this issue!

remkop: Updated to use a separate process instead of the same JVM as the test.

pawelpasterz:
@bootstraponline @remkop 
Thanks for comments, I've added changes you requested, please verify.
Thanks :)

bootstraponline:
Looks great! Approved. Thanks for updating and adding the test.

remkop:
The classpath in `UtilsTest` still has the dummy value `classpath...picocli-4.2.0.jar`. So that process will die immediately. It won’t hang. 

You can get a better classpath from the value of system property `"java.class.path"`. 
Sorry if the example was misleading.

bootstraponline:
Thanks for the follow up clarification! I opened https://github.com/Flank/flank/issues/662 to track this work.

pawelpasterz:
@remkop Omg, I totally missed that ... shame on me. Thanks for catching this!
Here is fix #663 I also added additional verification logic if process was started.
cc @bootstraponline

Issue #557, opened 2019-05-18
Time info incomplete when using orchestrator


bootstraponline:
Shard time is calculated by summing all test durations together. FTL doesn't report orchestrator setup/teardown. That means individual test times and overall shard times used by flank are incorrect.

- Work with FTL to get valid time data from the API
- Update sharding algorithm to use real time values

Example JUnit XML result with 3s per test overhead:

```xml
<testsuite name="" tests="2" failures="0" errors="0" skipped="0" time="8s" timestamp="2019-05-13T16:03:42" hostname="localhost">
    <testcase name="test1" classname="com.example.ExampleTest" time="1.0"/>
    <testcase name="test2" classname="com.example.ExampleTest" time="1.0"/>
</testsuite>
```

Total time in the JUnit XML is valid `<testsuite time="8s"`. Individual test times don't include overhead.

PR #666, opened 2020-03-13
Use API instead of XML for result parsing for android

### LINKED ISSUE #545: Use API instead of XML for result parsing
https://github.com/Flank/flank/blob/master/test_runner/src/test/kotlin/Tmp.kt

### LINKED ISSUE #618: jUnitReport.xml contains wrong data with smart sharding
I am using smart sharding and i recognized the the jUnitReport-xml file contains wrong data compared to real results of the matrixes. I can provide you some xml files for debugging if needed

jan-gogo:
Fixes #618, Fixes #545

## Checklist

- [x] Documented
- [x] Unit tested


test_runner/src/main/kotlin/ftl/reports/api/CreateTestExecutionData.kt
```diff
@@ -0,0 +1,26 @@
...
+import kotlinx.coroutines.runBlocking
+
+internal fun List<TestExecution>.createTestExecutionDataListAsync(): List<TestExecutionData> = runBlocking {
+    map { testExecution ->
+        async(Dispatchers.IO) {
+            // calling this function in async block, speeds up execution because
+            // it is making api calls under the hood
+            testExecution.createTestExecutionData()
+        }
+    }.map { deferred ->
```
pawelpasterz: I think we can use `.awaitAll()` instead, just for simplification

test_runner/src/main/kotlin/ftl/reports/api/ProcessFromApi.kt
```diff
@@ -0,0 +1,15 @@
...
+import ftl.gc.GcTestMatrix
+import ftl.json.MatrixMap
+import ftl.reports.xml.model.JUnitTestResult
+
+fun processXmlFromApi(
+    matrices: MatrixMap,
+    args: IArgs
+): JUnitTestResult = GcTestMatrix.refresh(
+    // Android uses server side sharding, it results only one matrix, so we can simply get first.
+    matrices.map.values.first().matrixId,
```
bootstraponline: iOS doesn't use server side sharding right now. Is it possible to use the API for both Android & iOS?

test_runner/src/test/kotlin/TmpV2.kt
```diff
@@ -0,0 +1,64 @@
...
+import ftl.gc.GcTestMatrix
+import ftl.reports.api.createJUnitTestResult
+import ftl.reports.api.createTestExecutionDataListAsync
+import ftl.reports.api.data.TestExecutionData
+import ftl.reports.api.filterForJUnitResult
+import ftl.reports.xml.xmlToString
+import java.io.File
+import kotlin.system.exitProcess
+
+object TmpV2 {
```
bootstraponline: we should probably delete `TmpV2` and turn this into a proper JUnit test. What do you think?

bootstraponline: Also probably delete the original tmp

test_runner/src/main/kotlin/ftl/reports/api/Utils.kt
```diff
@@ -0,0 +1,19 @@
+package ftl.reports.api
+
+import com.google.api.services.toolresults.model.Duration
+import java.util.*
```
bootstraponline: I think we have a lint rule about no wildcard imports.

test_runner/src/test/kotlin/TmpV2.kt
```diff
@@ -0,0 +1,64 @@
...
+        print("writing api results to file")
+        File(API_JSON_FILE).writeText(apiResult.toPrettyString())
+        println(" - OK")
+
+        print("generating junit report")
+        val jUnitTestResult = matrix.createJUnitTestResult()
+        println(" - OK")
+
+        print("writing api results to file")
+        File(JUNIT_REPORT_FILE).writeText(jUnitTestResult.xmlToString())
```
bootstraponline: To test this, run an Android app with multiple tests using server side sharding that has flaky tests. Use `num-flaky-test-attempts` set to a non-zero value such as 2. Then compare the JUnit XML file that FTL is creating vs the one we create via the API.

bootstraponline:
When using Android Test Orchestrator, we'll want to verify the API test times match the times returned in the FTL XML file.

https://github.com/Flank/flank/issues/557

test_runner/src/test/kotlin/TmpV2.kt
jan-gogo: Of Course I will do that. This is only draft of implementation that I tested manually and looks correct. I now need to update unit tests, cleanup a little, and It will be ready for review.

test_runner/src/test/kotlin/TmpV2.kt
jan-gogo: I was testing exactly that case. I know there is a difference between old and new JUnitReport.xml. 
Old: 
```
<?xml version='1.0' encoding='UTF-8' ?>
<testsuites>
  <testsuite name="NexusLowRes-28-en-portrait#" tests="3" failures="0" errors="0" skipped="0" time="3.103" timestamp="2020-03-10T17:12:24" hostname="localhost">
    <testcase name="test0" classname="com.example.test_app.InstrumentedTest" time="1.161">
      <webLink></webLink>
    </testcase>
    <testcase name="test1" classname="com.example.test_app.InstrumentedTest" time="1.034">
      <webLink></webLink>
    </testcase>
    <testcase name="test2" classname="com.example.test_app.InstrumentedTest" time="0.908">
      <webLink></webLink>
    </testcase>
  </testsuite>
</testsuites>
```
New:
```
<?xml version='1.0' encoding='UTF-8' ?>
<testsuites>
  <testsuite name="NexusLowRes-28-en-portrait" tests="1" failures="0" errors="0" skipped="0" time="1.161" timestamp="2020-03-10T17:07:54.422Z">
    <testcase name="test0" classname="com.example.test_app.InstrumentedTest" time="1.161"/>
  </testsuite>
  <testsuite name="NexusLowRes-28-en-portrait" tests="2" failures="0" errors="0" skipped="0" time="1.942" timestamp="2020-03-10T17:07:54.422Z">
    <testcase name="test1" classname="com.example.test_app.InstrumentedTest" time="1.034"/>
    <testcase name="test2" classname="com.example.test_app.InstrumentedTest" time="0.908"/>
  </testsuite>
</testsuites>
```

1. In new JUnitReport suites reflects shards. I can fix it easily, but that was first what I achieved. I guess that maybe is better to expose some info about shards in JUnitReport. If you don't like that I will fix it.achieved 

2. The `timestamp` is different yes. Old timestamp is simply timestamp of first of downloaded reports. New one is not correct I will fix it.

3. I need to also fix weblinks.

bootstraponline: makes sense!

test_runner/src/main/kotlin/ftl/reports/api/CreateJUnitTestCase.kt
```diff
@@ -0,0 +1,46 @@
...
+        name = testCase.testCaseReference.name,
+        classname = testCase.testCaseReference.className,
+        time = testCase.elapsedTime.format(),
+        failures = stackTraces["failed"],
+        errors = stackTraces["error"],
+        // skipped = true is represented by null. skipped = false is "absent"
+        skipped = if (testCase.status == "skipped") null else "absent"
+    ).apply {
+        if (errors != null || failures != null) {
+            webLink = getWebLink(toolResultsStep, testCase.testCaseId)
```
pawelpasterz: 👍

test_runner/src/main/kotlin/ftl/reports/api/CreateTestExecutionData.kt
```diff
@@ -0,0 +1,39 @@
...
+            // calling this function in async block, speeds up execution because
+            // it is making api calls under the hood
+            testExecution.createTestExecutionData()
+        }
+    }.awaitAll()
+}
+
+private fun TestExecution.createTestExecutionData(): TestExecutionData {
+    val response: ListTestCasesResponse = GcToolResults.listTestCases(toolResultsStep)
+    val step: Step = GcToolResults.getStepResult(toolResultsStep)
```
pawelpasterz: Both `GcToolResults.listTestCases` and `GcToolResults.getStepResult` are making api calls and seems that they don't depend on each other. Do you think we could make async call for each?

test_runner/src/main/kotlin/ftl/reports/api/PrepareForJUnitResult.kt
```diff
@@ -0,0 +1,45 @@
...
+}
+
+private fun List<TestExecutionData>.reduceTestCases() = map(TestExecutionData::reduceTestCases)
+
+private fun TestExecutionData.reduceTestCases() = copy(
+    testCases = testCases.groupBy(TestCase::getTestCaseId).map { (_, testCases) ->
+        testCases.sortedBy { testCase: TestCase ->
+            testCase.startTime.asUnixTimestamp()
+        }.run {
+            firstOrNull { it.stackTraces != null } ?: first()
```
pawelpasterz: It can be just me (and probably is) but this lines looks bit strange for me (at least for the first glance). 
```
.run {
    firstOrNull { it.stackTraces != null } ?: first()
}
```
I am wondering if it is doable and we could add private extension function like `.firstWithStackTrace` or sth similar. 
Just thinking

test_runner/src/main/kotlin/ftl/reports/api/Utils.kt
```diff
@@ -0,0 +1,34 @@
...
+
+private val utcDateFormat = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss").apply {
+    timeZone = TimeZone.getTimeZone("UTC")
+}
+
+fun Timestamp.asUnixTimestamp() = (seconds ?: 0) * 1_000 + (nanos ?: 0) / 1_000_000
+
+fun Long.formatUtcDate() = utcDateFormat.format(this)!!
+
+var TestCase.flaky: Boolean by mutableMapProperty { false }
```
pawelpasterz: 👍

test_runner/src/main/kotlin/ftl/reports/api/Utils.kt
```diff
@@ -0,0 +1,34 @@
+package ftl.reports.api
```
pawelpasterz: We already have `Utils.kt` file, can we rename this one to sth like `ApiUtils` or somehow different? Just for sake of clarity.

test_runner/src/main/kotlin/ftl/reports/xml/JUnitXml.kt
```diff
@@ -15,7 +15,7 @@ private val xmlMapper = XmlMapper(xmlModule)
     .registerModules(KotlinModule())
     .configure(FAIL_ON_UNKNOWN_PROPERTIES, false)
 
-private val xmlPrettyWriter = xmlMapper.writerWithDefaultPrettyPrinter()
+internal val xmlPrettyWriter = xmlMapper.writerWithDefaultPrettyPrinter()
 
 private fun xmlBytes(path: Path): ByteArray {
     if (!path.toFile().exists()) RuntimeException("$path doesn't exist!")
```
pawelpasterz: I am totally aware it is not part of your PR (so feel free to skip this comment). I think there is `throw` missing

test_runner/src/main/kotlin/ftl/reports/xml/model/JUnitTestSuite.kt
```diff
@@ -29,11 +32,11 @@ data class JUnitTestSuite(
     val timestamp: String?, // String. Android only
 
     @JacksonXmlProperty(isAttribute = true)
-    val hostname: String, // String.
+    val hostname: String? = null, // String.
```
pawelpasterz: Just for my understanding, are we ok with possible `null` here? Previous implementation did not predicted such value. Thanks!

test_runner/src/main/kotlin/ftl/run/platform/common/AfterRunTests.kt
```diff
@@ -11,37 +11,32 @@ import java.nio.file.Files
...
+private fun List<TestMatrix>.toSavedMatrixMap() = this
+    .map { matrix -> matrix.testMatrixId to SavedMatrix(matrix) }
+    .toMap()
+
 private fun saveConfigFile(matrixMap: MatrixMap, args: IArgs) {
-    val configFilePath = if (args.useLocalResultDir()) {
-        Paths.get(args.localResultDir, FtlConstants.configFileName(args))
-    } else {
+    val configFilePath = if (args.useLocalResultDir())
+        Paths.get(args.localResultDir, FtlConstants.configFileName(args)) else
```
pawelpasterz: TBH I don't like it when I need to read whole line to find out there is else at the end. But it is matter of personal preferences  :)

test_runner/src/main/kotlin/ftl/reports/api/CreateJUnitTestSuite.kt
```diff
@@ -0,0 +1,43 @@
...
+    testcases = testCases.toMutableList(),
+    time = testCases.sumTime().format() // FIXME include also setup and teardown duration https://github.com/Flank/flank/issues/557
+)
+
+private fun Step.testSuiteName(): String {
+    val map = dimensionValue.map { it.key to it.value }.toMap()
+    return listOf(map["Model"], map["Version"], map["Locale"], map["Orientation"]).joinToString("-")
+}
+
+private fun List<JUnitTestCase>.sumTime() = this
```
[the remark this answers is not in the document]
jan-gogo: That's right, intentionally for readability

test_runner/src/main/kotlin/ftl/reports/api/CreateTestExecutionData.kt
jan-gogo: Right, I will fix that.

test_runner/src/main/kotlin/ftl/reports/api/Utils.kt
jan-gogo: Looks like you don't care about packages :). I guess adding part of package as prefix for class is redundant.

test_runner/src/main/kotlin/ftl/reports/xml/JUnitXml.kt
jan-gogo: Good checkup. I can fix it in this PR, no problem.

test_runner/src/main/kotlin/ftl/reports/xml/model/JUnitTestSuite.kt
jan-gogo: I have asked @bootstraponline about that, he said https://github.com/Flank/flank/issues/545#issuecomment-597885474. So I decide to change it to nullable here, because I didn't find hostname in API

test_runner/src/main/kotlin/ftl/run/platform/common/AfterRunTests.kt
jan-gogo: TBH I don't like `if else` and prefer using `when` but our `detect` config forces to use `if else` in cases like this. Anyway I like that kind of formatting cause of clear indentation same for two parts.

test_runner/src/main/kotlin/ftl/reports/api/PrepareForJUnitResult.kt
jan-gogo: As you wish, code should be readable for everyone who contribute

jan-gogo: I guess `firstWithStackTrace` can be little bit misleading. Should be `firstWithStackTraceOrFirst`, but is to long, and I avoid names that contain condition. So I hope this construction will solve the problem

```
.run {
    if (!isFlaky()) first()
    else first { it.stackTraces != null }.apply { flaky = true }
}
```

test_runner/src/main/kotlin/ftl/reports/api/Utils.kt
pawelpasterz: Yeah, true. Sometimes I am kind of bitchy :D ¯\_(ツ)_/¯

test_runner/src/main/kotlin/ftl/reports/xml/model/JUnitTestSuite.kt
pawelpasterz: Got it, thanks for explanation

test_runner/src/main/kotlin/ftl/reports/api/Utils.kt
jan-gogo: Easy. You are doing your work. It is always better to express your opinion than to do nothing. There are no ultimate ways, everything we do is based on consensus.

Issue #532, opened 2019-03-28
Run timeout


bootstraponline:
# Use Case

Sometimes tests hang forever due to FTL outages or an issue in the test. Currently Flank only supports a shard level timeout. Run timeouts are enforced at the CI job level. Ideally Flank would be able to timeout a stuck run without relying on the CI timeout.

# Feature

```yaml
flank:
  # The max time this test run can execute before it is cancelled (default: unlimited).
  run-timeout: 60m
```

Support the same duration format as `--timeout`.
- `1h`, `5m`, `200s`
- `100` is 100 seconds

After exceeding the the run timeout, Flank should:

- Print a message that says the run timeout has been exceeded
- Cancel any in progress test executions and print a link to those executions on FTL.

PR #671, opened 2020-03-15
Refactor coroutines

pawelpasterz:
Small refactor that prepares background for future work (especially #532 )

### DIFF
--- a/test_runner/src/main/kotlin/ftl/gc/GcTestMatrix.kt
+++ b/test_runner/src/main/kotlin/ftl/gc/GcTestMatrix.kt
@@ -2,7 +2,6 @@ package ftl.gc
 
 import com.google.api.services.testing.model.CancelTestMatrixResponse
 import com.google.api.services.testing.model.TestMatrix
-import ftl.args.IArgs
 import ftl.http.executeWithRetry
 import kotlinx.coroutines.delay
 import java.time.Duration.ofHours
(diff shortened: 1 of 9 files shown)


test_runner/src/main/kotlin/ftl/gc/GcTestMatrix.kt
```diff
@@ -33,7 +33,7 @@ object GcTestMatrix {
     //        "message" : "The service is currently unavailable.",
     //        "status" : "UNAVAILABLE"
     //    }
-    fun refresh(testMatrixId: String, args: IArgs): TestMatrix {
+    suspend fun refresh(testMatrixId: String, args: IArgs): TestMatrix {
```
jan-gogo: Simplifying is right way, I like it. You have changed signature of `cancel` function but you forgot to do the same think here.

test_runner/src/main/kotlin/ftl/gc/GcTestMatrix.kt
```diff
@@ -42,16 +42,16 @@ object GcTestMatrix {
...
                 failed += 1
             }
         }
 
         throw RuntimeException("Failed to refresh matrix")
     }
 
-    fun cancel(testMatrixId: String, args: IArgs): CancelTestMatrixResponse {
-        val cancelMatrix = GcTesting.get.projects().testMatrices().cancel(args.project, testMatrixId)
+    suspend fun cancel(testMatrixId: String, project: String): CancelTestMatrixResponse {
```
jan-gogo: Maybe `projectId` will be more meaningful.

pawelpasterz:
@jan-gogo 
Thanks for comments, I've implemented changes.

PR #677, opened 2020-03-19
Add CommonRunCommand

pawelpasterz:
## Changes
* introduce `CommonRunCommand` which keeps duplicated parameters for both android and ios run commands

### DIFF
--- a/test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
+++ b/test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
@@ -71,7 +71,7 @@ class AndroidArgs(
     override val filesToDownload = cli?.filesToDownload ?: flank.filesToDownload
     override val disableSharding = cli?.disableSharding ?: flank.disableSharding
     override val project = cli?.project ?: flank.project
-    override val localResultDir = cli?.localResultDir ?: flank.localResultDir
+    override val localResultDir = cli?.localResultsDir ?: flank.localResultDir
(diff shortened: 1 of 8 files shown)


test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
```diff
@@ -71,7 +71,7 @@ class AndroidArgs(
     override val filesToDownload = cli?.filesToDownload ?: flank.filesToDownload
     override val disableSharding = cli?.disableSharding ?: flank.disableSharding
     override val project = cli?.project ?: flank.project
-    override val localResultDir = cli?.localResultDir ?: flank.localResultDir
+    override val localResultDir = cli?.localResultsDir ?: flank.localResultDir
```
bootstraponline: Do you think we should rename `flank.localResultDir` as well if `cli?.localResultDir` is changing?

jan-gogo: I am curious about using plural/singular form in naming convention. @bootstraponline you are native, which one is more proper in this case `localResultDir` or `localResultsDir`?

bootstraponline: The folder on disk is called `results/` so `localResultsDir` makes sense to me.

pawelpasterz: Well, TBH I changed from `result` to `results` just to be on the same page with `IosArgs` and move it to `CommonRunCommad`. I will fix it then.

PR #678, opened 2020-03-20
Skip Bugsnag initialisation if user disabled gcloud analytics

### LINKED ISSUE #581: Respect gcloud analytics opt-out
If a user has opted out of gcloud analytics, disable reporting bugs to bugsnag.

```python
_UUID_FILE_PATH = os.path.expanduser(os.path.join('~',
                                                  '.gsutil/analytics-uuid'))
# If this string is written to analytics-uuid, that means that the user said
# 'no' to analytics, and it should thus be disabled.
_DISABLED_TEXT = 'DISABLED'
```

https://github.com/bootstraponline/gcloud_cli/blob/41b4d3438d5a8e215912897d2aa615fd8d509e0d/google-cloud-sdk/platform/gsutil/gslib/metrics.py#L55

pawelpasterz:
Fixes #581 

## Changes
* Introduce `BugsnagInitHelper` class to handle Bugsnag initialisation logic
* if `useMock` is equal true -- Bugsnag is not enabled
* if user had disabled `gcloud` analytics -- Bugsnag is not enabled
* in other cases -- Bugsnag enabled
* Add `updateFlank` gradle task to update local flank jar
* Add `mockk` library as POC (will replace `mockito` in the future)

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/test_runner/src/main/kotlin/ftl/config/FtlConstants.kt
+++ b/test_runner/src/main/kotlin/ftl/config/FtlConstants.kt
@@ -1,6 +1,5 @@
 package ftl.config
 
-import com.bugsnag.Bugsnag
 import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport
 import com.google.api.client.googleapis.util.Utils
 import com.google.api.client.http.GoogleApiLogger
@@ -13,6 +12,7 @@ import com.google.auth.oauth2.ServiceAccountCredentials
 import ftl.args.AndroidArgs
 import ftl.args.IArgs
 import ftl.args.IosArgs
+import ftl.config.BugsnagInitHelper.initBugsnag
 import ftl.gc.UserAuth
 import ftl.http.HttpTimeoutIncrease
 import ftl.util.fatalError
@@ -38,7 +38,6 @@ object FtlConstants {
     }
 
     const val localhost = "http://localhost:8080"
-
     const val defaultLocale = "en"
     const val defaultOrientation = "portrait"
     const val defaultIosModel = "iphone8"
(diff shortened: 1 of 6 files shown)


test_runner/src/test/kotlin/ftl/config/BugsnagInitHelperTest.kt
```diff
@@ -0,0 +1,70 @@
...
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.rules.TemporaryFolder
+import java.io.File
+import java.util.UUID
+
+private const val GSUTIL_FOLDER = ".gsutil"
+private const val ANALYTICS_FILE = "analytics-uuid"
+private const val DISABLED = "DISABLED\n"
```
doodla: Why is there a new line at the end?

pawelpasterz: This file can be added manually by the user. Since some of text editors add new line char at the end of text file automatically I wanted to simulate that case.

doodla:
Gotcha. Makes sense.

test_runner/src/main/kotlin/ftl/config/FtlConstants.kt
```diff
@@ -52,12 +51,13 @@ object FtlConstants {
     const val applicationName = "Flank"
     const val GCS_PREFIX = "gs://"
     const val runTimeout = "-1"
+
     val JSON_FACTORY: JsonFactory by lazy { Utils.getDefaultJsonFactory() }
 
-    val bugsnag = Bugsnag(if (useMock) null else "3d5f8ba4ee847d6bb51cb9c347eda74f")
+    val bugsnag = initBugsnag(useMock)
```
jan-gogo: Maybe will be better to get rid of bugsnag reference and initialization in `FtlConstants` and move it to somewhere else. Also I am not sure if we need to keep reference to bugsnag because it's used only to setting app version. Additionally make sure that it will be initialized after `MockServer` start because it changes `useMock` flag. BTW I don't know why useMock flag belongs to FtlConstants, it is not a constant :), but I guess it's another story.

PR #683, opened 2020-03-25
Print web link

### LINKED ISSUE #645: Print test results URL
Flank should now print a link to the FTL test results URL to the console after starting the run.

bootstraponline:
Fixes #645

## Checklist

- [x] Unit tested

```
  4 matrix ids created in 0m 22s
  https://console.developers.google.com/storage/browser/test-lab-v9cn46bb990nx-kz69ymd4nm9aq/2020-04-01_07-25-03.132000_AQWv/

Matrices webLink
  matrix-26tm1277rg3gr: https://console.firebase.google.com/project/flank-open-source/testlab/histories/bh.da0c237aaa33732/matrices/8849499259998194241
  matrix-2veeo7xug1f8k: https://console.firebase.google.com/project/flank-open-source/testlab/histories/bh.da0c237aaa33732/matrices/5106500884348802256
  matrix-2z6o0hu8tt5bf: https://console.firebase.google.com/project/flank-open-source/testlab/histories/bh.da0c237aaa33732/matrices/4681223845649672500
  matrix-2jl1g9zptcs9m: https://console.firebase.google.com/project/flank-open-source/testlab/histories/bh.da0c237aaa33732/matrices/6721963817674217321/executions/bs.faacc05c891cc43c

PollMatrices
```

### DIFF
--- a/test_runner/src/main/kotlin/ftl/run/common/PollMatrices.kt
+++ b/test_runner/src/main/kotlin/ftl/run/common/PollMatrices.kt
@@ -8,6 +8,7 @@ import ftl.util.MatrixState
 import ftl.util.StopWatch
 import ftl.util.StopWatchMatrix
 import ftl.util.completed
+import ftl.util.webLink
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.delay
 
@@ -36,8 +37,15 @@ private suspend fun pollMatrix(matrixId: String, stopwatch: StopWatch, args: IAr
     var refreshedMatrix = GcTestMatrix.refresh(matrixId, args.project)
     val watch = StopWatchMatrix(stopwatch, matrixId)
     val runningDevices = RunningDevices(stopwatch, refreshedMatrix.testExecutions)
+    var printWebLink = true
 
     while (true) {
+        val webLink = refreshedMatrix.webLink()
+        if (printWebLink && webLink.isNotBlank()) {
+            printWebLink = false
+            println(webLink)
+        }
+
         if (matrices.map[matrixId]?.update(refreshedMatrix) == true) updateMatrixFile(
             matrices,
             args
--- a/test_runner/src/main/kotlin/ftl/run/common/PollMatrices.kt
+++ b/test_runner/src/main/kotlin/ftl/run/common/PollMatrices.kt
@@ -8,7 +8,6 @@ import ftl.util.MatrixState
 import ftl.util.StopWatch
 import ftl.util.StopWatchMatrix
 import ftl.util.completed
-import ftl.util.webLink
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.delay
 
@@ -37,15 +36,8 @@ private suspend fun pollMatrix(matrixId: String, stopwatch: StopWatch, args: IAr
     var refreshedMatrix = GcTestMatrix.refresh(matrixId, args.project)
     val watch = StopWatchMatrix(stopwatch, matrixId)
     val runningDevices = RunningDevices(stopwatch, refreshedMatrix.testExecutions)
-    var printWebLink = true
 
     while (true) {
-        val webLink = refreshedMatrix.webLink()
-        if (printWebLink && webLink.isNotBlank()) {
-            printWebLink = false
-            println(webLink)
-        }
-
         if (matrices.map[matrixId]?.update(refreshedMatrix) == true) updateMatrixFile(
             matrices,
             args
--- a/test_runner/src/main/kotlin/ftl/run/platform/common/AfterRunTests.kt
+++ b/test_runner/src/main/kotlin/ftl/run/platform/common/AfterRunTests.kt
@@ -3,14 +3,19 @@ package ftl.run.platform.common
 import com.google.api.services.testing.model.TestMatrix
 import ftl.args.IArgs
 import ftl.config.FtlConstants
+import ftl.gc.GcTestMatrix
 import ftl.json.MatrixMap
 import ftl.json.SavedMatrix
 import ftl.run.common.updateMatrixFile
 import ftl.util.StopWatch
+import ftl.util.webLink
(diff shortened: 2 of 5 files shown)


test_runner/src/main/kotlin/ftl/run/common/PollMatrices.kt
```diff
@@ -36,8 +37,15 @@ private suspend fun pollMatrix(matrixId: String, stopwatch: StopWatch, args: IAr
...
     val watch = StopWatchMatrix(stopwatch, matrixId)
     val runningDevices = RunningDevices(stopwatch, refreshedMatrix.testExecutions)
+    var printWebLink = true
 
     while (true) {
+        val webLink = refreshedMatrix.webLink()
+        if (printWebLink && webLink.isNotBlank()) {
+            printWebLink = false
+            println(webLink)
+        }
```
bootstraponline: Is there a cleaner way to print the web link once per matrix?

pawelpasterz: I am wondering if we could print this link using this logic in `SavedMatrix`
https://github.com/Flank/flank/blob/d91f7784dadfafb9359a9e487d53507fda57e432/test_runner/src/main/kotlin/ftl/json/SavedMatrix.kt#L65-L67
AFAIK link is updated only once for each matrix (please confirm that, I've tested it with couple of examples but might be some corner cases about which I am not aware of).
Also, I think we could print it in more user friendly way --  with indention, message saying what exactly this is etc.
I am just thinking, let me know what you think.

bootstraponline: I like that idea!

My understanding is the matrix link is only updated once. I'll update the pull request.

jan-gogo: If we need to print web link ASAP "after starting run", I guess that the cleaner way is to do first refresh of matrices somewhere [here](https://github.com/Flank/flank/blob/f114c83789d95525b35d7daa6f200d7feebd23d6/test_runner/src/main/kotlin/ftl/run/platform/common/AfterRunTests.kt#L29). So the web link can be properly generated and printed before "PollMatrices" message will appears. But it's not good idea if number of additional pulls matters.

test_runner/src/main/kotlin/ftl/run/platform/common/AfterRunTests.kt
```diff
@@ -41,3 +48,16 @@ private fun saveConfigFile(matrixMap: MatrixMap, args: IArgs) {
...
+    map.values.map { launch { it.printWebLink(project) } }.joinAll()
+    println()
+}
+
+private suspend inline fun SavedMatrix.printWebLink(project: String) =
+    println("${FtlConstants.indent}$matrixId: ${getOrUpdateWebLink(webLink, project, matrixId)}")
+
+private tailrec suspend fun getOrUpdateWebLink(link: String, project: String, matrixId: String): String =
+    if (link.isNotBlank()) link
+    else getOrUpdateWebLink(GcTestMatrix.refresh(matrixId, project).webLink(), project, matrixId)
```
bootstraponline, line 63: A matrix is not guaranteed to have a webLink. A matrix can be invalid.
https://github.com/Flank/flank/blob/master/test_runner/src/main/kotlin/ftl/util/MatrixState.kt

Here are the possible states:

```
        "enumDescriptions": [
            "Do not use.  For proto versioning only.",
            "The execution or matrix is being validated.",
            "The execution or matrix is waiting for resources to become available.",
            "The execution is currently being processed.\n\nCan only be set on an execution.",
            "The execution or matrix has terminated normally.\n\nOn a matrix this means that the matrix level processing completed normally,\nbut individual executions may be in an ERROR state.",
            "The execution or matrix has stopped because it encountered an\ninfrastructure failure.",
            "The execution was not run because it corresponds to a unsupported\nenvironment.\n\nCan only be set on an execution.",
            "The execution was not run because the provided inputs are incompatible with\nthe requested environment.\n\nExample: requested AndroidVersion is lower than APK's minSdkVersion\n\nCan only be set on an execution.",
            "The execution was not run because the provided inputs are incompatible with\nthe requested architecture.\n\nExample: requested device does not support running the native code in\nthe supplied APK\n\nCan only be set on an execution.",
            "The user cancelled the execution.\n\nCan only be set on an execution.",
            "The execution or matrix was not run because the provided inputs are not\nvalid.\n\nExamples: input file is not of the expected type, is malformed\/corrupt, or\nwas flagged as malware"
          ],
          "type": "string",
          "enum": [
            "TEST_STATE_UNSPECIFIED",
            "VALIDATING",
            "PENDING",
            "RUNNING",
            "FINISHED",
            "ERROR",
            "UNSUPPORTED_ENVIRONMENT",
            "INCOMPATIBLE_ENVIRONMENT",
            "INCOMPATIBLE_ARCHITECTURE",
            "CANCELLED",
            "INVALID"
          ]
```

- [MatrixState.kt](https://github.com/Flank/flank/blob/master/test_runner/src/main/kotlin/ftl/util/MatrixState.kt)
- [testing_v1.json#L171](https://github.com/Flank/gcloud_cli/blob/master/google-cloud-sdk/lib/googlecloudsdk/third_party/apis/testing_v1.json#L171)

I think only:
```json
"VALIDATING",
"PENDING",
"RUNNING",
"FINISHED",
```
will may have a webLink that we can poll for.

```json
"TEST_STATE_UNSPECIFIED",
"UNSUPPORTED_ENVIRONMENT",
"INCOMPATIBLE_ENVIRONMENT",
"INCOMPATIBLE_ARCHITECTURE",
"CANCELLED",
"INVALID"
```

I don't think we'll get a web link from these states. In the current implementation, will we poll indefinitely until a web link is provided?

pawelpasterz: Thanks for input. I will implement additional logic to prevent this.

PR #684, opened 2020-03-27
Add overhead time to junit test case report

### LINKED ISSUE #557: Time info incomplete when using orchestrator

jan-gogo:
Fixes #557 

## Checklist

- [x] Unit tested

### DIFF
--- a/test_runner/src/main/kotlin/ftl/reports/api/CreateTestSuitOverviewData.kt
+++ b/test_runner/src/main/kotlin/ftl/reports/api/CreateTestSuitOverviewData.kt
@@ -1,20 +1,35 @@
 package ftl.reports.api
 
 import com.google.api.services.toolresults.model.TestCase
+import com.google.api.services.toolresults.model.TestSuiteOverview
 import ftl.reports.api.data.TestExecutionData
 import ftl.reports.api.data.TestSuiteOverviewData
 
 internal fun TestExecutionData.createTestSuitOverviewData(): TestSuiteOverviewData {
-    val skipped = step.testExecutionStep.testSuiteOverviews.first().skippedCount ?: 0
+    val overview: TestSuiteOverview = step.testExecutionStep.testSuiteOverviews.first()
+    val skipped: Int = overview.skippedCount ?: 0
     return TestSuiteOverviewData(
         total = testCases.size + skipped,
         errors = testCases.countErrors(),
         failures = testCases.countFailures(),
         flakes = testCases.countFlakes(),
-        skipped = skipped
+        skipped = skipped,
+        elapsedTime = overview.elapsedTime.millis(),
+        overheadTime = getOverheadTime(overview, testCases)
     )
 }
 
 private fun List<TestCase>.countErrors() = count { !it.flaky && it.status == "error" }
 private fun List<TestCase>.countFailures() = count { !it.flaky && it.status == "failed" }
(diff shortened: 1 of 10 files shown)


bootstraponline:
awesome!

jan-gogo:
Hey @bootstraponline there is one think to consider here. 
Our new junit report is uploaded to bucket only when option `smart-flank-gcs-path` is used. That means to use smart sharding correctly we need to run flank twice with `smart-flank-gcs-path`. First run will generate and upload correct report so second run can use it to properly calculate shard times. I it correct approach? Or maybe new junit report should be uploaded always, no matter if `smart-flank-gcs-path` used or not.

jan-gogo:
Second thing is that test suites names in junit report may not be unique so method `mergeTestTimes` [here](https://github.com/Flank/flank/blob/f114c83789d95525b35d7daa6f200d7feebd23d6/test_runner/src/main/kotlin/ftl/reports/util/ReportManager.kt#L179) may not work correctly. For now I have not idea how to specify unique names to identify same suites between two test runs. But I guess that we can adjust implementation to omit suite names matching and base only on testcase names (and class names) from all suites. But them also may not be unique if user will add additional test apks where both contains same test name.  Do you have any suggestions?

test_runner/src/main/kotlin/ftl/reports/util/ReportManager.kt
```diff
@@ -176,7 +176,9 @@ object ReportManager {
 
         val oldTestResult = GcStorage.downloadJunitXml(args)
 
-        newTestResult.mergeTestTimes(oldTestResult)
+        if (args.useLegacyJUnitResult) {
+            newTestResult.mergeTestTimes(oldTestResult)
```
jan-gogo: I've decided to omit this call cause new junit result (from api) does not favor success over failure nor error.

test_runner/src/main/kotlin/ftl/reports/api/CreateTestSuitOverviewData.kt
```diff
@@ -1,20 +1,35 @@
...
+private fun getOverheadTime(
+    overview: TestSuiteOverview,
+    testCases: List<TestCase>
+) = if (testCases.isEmpty())
+    0.0 else
+    (overview.elapsedTime.millis() - testCases.sumTime()) / testCases.size
+
+private fun List<TestCase>.sumTime(): Double = this
+    .map { it.elapsedTime.millis() }
+    .reduce { acc, d -> acc + d }
```
pawelpasterz: Hey! I think we can simplify this with `this.sumByDouble { it.elapsedTime.millis() }`. Let me know what do you think. Thanks!

PR #687, opened 2020-03-29
Fix logging level for both root and Bugsnag not being initialized

### LINKED ISSUE #682: Debug message printed after every command
> 21:57:34.445 [Thread-2] DEBUG com.bugsnag.Bugsnag - Closing connection to Bugsnag

Debug should only print if we have toggled debug mode.

pawelpasterz:
Fixes #682 

Due to lazy initialization of Kotlin's objects logging level was not set (default was used).

## Checklist

### DIFF
--- a/test_runner/src/main/kotlin/ftl/Main.kt
+++ b/test_runner/src/main/kotlin/ftl/Main.kt
@@ -39,7 +39,8 @@ class Main : Runnable {
 
     @CommandLine.Option(
         names = ["--debug"],
-        description = ["Enables debug logging"]
+        description = ["Enables debug logging"],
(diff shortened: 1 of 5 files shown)


test_runner/src/main/kotlin/ftl/Main.kt
```diff
@@ -39,7 +39,8 @@ class Main : Runnable {
 
     @CommandLine.Option(
         names = ["--debug"],
-        description = ["Enables debug logging"]
+        description = ["Enables debug logging"],
+        defaultValue = "false"
```
jan-gogo: 👍

PR #696, opened 2020-04-03
Introduce global exception handler

pawelpasterz:
### Refactor: 
* throw custom exceptions instead of multiple `exitProcess(...)` invocation
* introduce common exception handler

### DIFF
--- a/test_runner/src/main/kotlin/ftl/util/Utils.kt
+++ b/test_runner/src/main/kotlin/ftl/util/Utils.kt
@@ -6,7 +6,6 @@ import ftl.config.FtlConstants
 import ftl.json.SavedMatrix
 import ftl.run.cancelMatrices
 import kotlinx.coroutines.runBlocking
-import picocli.CommandLine
 import java.io.InputStream
 import java.io.StringWriter
 import java.nio.file.Files
@@ -47,10 +46,7 @@ inline fun fatalError(e: Exception, message: String? = null) {
 }
 
 @Suppress("NOTHING_TO_INLINE")
-inline fun flankFatalError(e: String): String {
-    if (FtlConstants.useMock) throw RuntimeException(e)
-    else throw FlankFatalError(e)
-}
+inline fun flankFatalError(e: String): Nothing = throw FlankFatalError(e)
 
 fun assertNotEmpty(str: String, e: String) {
     if (str.isEmpty()) {
@@ -138,8 +134,6 @@ fun copyBinaryResource(name: String) {
     destinationFile.setExecutable(true)
 }
 
-// We need to cover the case where some component in the call stack starts a non-daemon
-// thread, and then throws an Error that kills the main thread. This is extra safe implementation
 fun withGlobalExceptionHandling(block: () -> Int) {
     try {
         exitProcess(block())
@@ -161,15 +155,22 @@ fun withGlobalExceptionHandling(block: () -> Int) {
             }
             is FTLError -> {
                 t.matrix.logError("not finished")
-                exitProcess(2)
+                exitProcess(3)
             }
             is FlankFatalError -> {
                 System.err.println(t.message)
-                exitProcess(3)
+                exitProcess(2)
             }
+            // We need to cover the case where some component in the call stack starts a non-daemon
+            // thread, and then throws an Error that kills the main thread. This is extra safe implementation
             else -> {
+                // this is workaround for Bugsnag initialization bug, should be removed when resolved
+                // https://github.com/Flank/flank/issues/699
+                if (FtlConstants.useMock.not()) {
+                    FtlConstants.bugsnag?.notify(t)
+                }
                 t.printStackTrace()
-                exitProcess(CommandLine.ExitCode.SOFTWARE)
+                exitProcess(3)
             }
         }
     }
--- a/test_runner/src/main/kotlin/ftl/util/Utils.kt
+++ b/test_runner/src/main/kotlin/ftl/util/Utils.kt
@@ -39,18 +39,9 @@ fun join(first: String, vararg more: String): String {
         .replace(regex = Regex("/+"), replacement = "/")
 }
 
-// marked as inline to make JaCoCo happy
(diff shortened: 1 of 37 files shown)


test_runner/src/main/kotlin/ftl/util/Utils.kt
```diff
@@ -149,15 +140,45 @@ fun copyBinaryResource(name: String) {
...
 // We need to cover the case where some component in the call stack starts a non-daemon
 // thread, and then throws an Error that kills the main thread. This is extra safe implementation
-fun jvmHangingSafe(block: () -> Int) {
+fun withGlobalExceptionHandling(block: () -> Int) {
     try {
         exitProcess(block())
     } catch (t: Throwable) {
-        t.printStackTrace()
-        exitProcess(CommandLine.ExitCode.SOFTWARE)
+        when (t) {
```
bootstraponline, line 147: Let's improve the documentation, both in code & the readme about exit status.

This was the old documentation for reference.
```
 * exit code 0 - all tests passed
  * exit code 1 - at least one test failed/inconclusive & all matrices finished
  * exit code 2 - at least one matrix not finished (usually FTL error)
```

bootstraponline:
The refactor looks awesome. I think this cleans up the logic. Requested changes about documenting the new exit codes and what they mean. Downstream tools will read the exit codes.

test_runner/src/main/kotlin/ftl/util/Utils.kt
pawelpasterz: @bootstraponline thanks! Here is my proposal:
* `0` -- everything is fine
* `1` --  failed matrix, yml validation failed or flank cancelled the run
* `2` -- Wrong usage of flags, incorrect values, errors in config files
* `3` --  FTL internal errors or other 'unexpected' exceptions (here I'd like also add line to notify bugsnag if 'unexpected' error occurred)

Let me know what do you think

bootstraponline: In this proposal, what exit code represents `at least one test failed/inconclusive & all matrices finished`?

It seems like we have a few categories of outcomes:

0 - everything is fine
1 - a test failed
2 - FTL server side issue/error
3 - Flank client side issue

pawelpasterz: > In this proposal, what exit code represents `at least one test failed/inconclusive & all matrices finished`?
It will be `1`

I think we are on the same page.

bootstraponline: Perfect. Thanks.

test_runner/src/main/kotlin/ftl/util/Utils.kt
```diff
@@ -38,32 +40,21 @@ fun join(first: String, vararg more: String): String {
...
-        Thread.sleep(ofSeconds(seconds).toMillis())
-    } catch (e: Exception) {
-        System.err.println(e)
-    }
-}
-
 // marked as inline to make JaCoCo happy
 @Suppress("NOTHING_TO_INLINE")
 inline fun fatalError(e: Exception, message: String? = null) {
     throw RuntimeException(message ?: e.toString(), e)
```
jan-gogo, line 46: Is was intentional to leave that name unchanged, in opposite to function below?

test_runner/src/main/kotlin/ftl/util/Utils.kt
```diff
@@ -38,32 +40,21 @@ fun join(first: String, vararg more: String): String {
...
 
 @Suppress("NOTHING_TO_INLINE")
-inline fun fatalError(e: String): String {
-    if (FtlConstants.useMock) {
-        throw RuntimeException(e)
-    }
-    System.err.println(e)
-    exitProcess(3)
+inline fun flankFatalError(e: String): String {
+    if (FtlConstants.useMock) throw RuntimeException(e)
```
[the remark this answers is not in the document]
pawelpasterz: Good catch, thanks!

test_runner/src/main/kotlin/ftl/util/Utils.kt
pawelpasterz: Yes it was, firstly I wanted to get rid of overloaded methods. And secondly I noticed that this method is used for FTL/gcloud related exceptions and renamed one for flank related errors. I wanted to keep it separate.

pawelpasterz:
@bootstraponline @jan-gogo 
I've implemented requested changes , please review. Thanks!

test_runner/src/main/kotlin/ftl/util/Utils.kt
bootstraponline: should we rename this to `firebaseFatalError` then to differentiate from `flankFatalError`?

jan-gogo: I appreciate dedicated exceptions in FlankException.kt this is clean and meaningful but `fatalError` is legacy and now looks kinda odd for me. Maybe te best solution will be getting rid of that helper functions and replace them with direct throws?

bootstraponline: The only issue I have with direct throws is it's not clear why we're throwing. A helper may communicate better?

jan-gogo: Hmm, I guess that the type of exception plus additional message can communicate as well. Helper adds one additional step to stack trace which holds source of throw.

bootstraponline: > the type of exception plus additional message can communicate as well. Helper adds one additional step to stack trace which holds source of throw.

That makes sense. I'm fine with dropping the helper.

pawelpasterz: @bootstraponline @jan-gogo 
Hey! As suggested (and I agree with your comments) I removed helper methods

PR #708, opened 2020-04-08
Add ignore failed tests option to Flank

### LINKED ISSUE #602: Add ignoreFailures to Flank
re: https://github.com/runningcode/fladle/issues/48

It seems there's a use case for ignoring the flank exit code. fladle failing will fail the build. In Jenkins, some users want to separate build failures from test failures. To do this, they rely on a Jenkins plugin to inspect the JUnit XML results.

image

Gradle supports providing control to the user over if test failures should break the build or not.

pawelpasterz:
Fixes #602 

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,6 +170,9 @@ flank:
 
   ## The max time this test run can execute before it is cancelled (default: unlimited).
   # run-timeout: 60m
+
+  ## Set true to end with exit code 0 despite of failed matrices. False by default.
+  # ignore-failed-tests: true
 ```
 
 ### Android example
@@ -329,6 +332,9 @@ flank:
 
   ## The max time this test run can execute before it is cancelled (default: unlimited).
   # run-timeout: 60m
+  
+  ## Set true to end with exit code 0 despite of failed matrices. False by default.
+  # ignore-failed-tests: true
 ```
 
 ### Android code coverage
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -170,6 +170,9 @@ flank:
 
   ## The max time this test run can execute before it is cancelled (default: unlimited).
   # run-timeout: 60m
+
+  ## Set true to end with exit code 0 despite of failed matrices. False by default.
```
bootstraponline, line 174: > ## Terminate with exit code 0 when there are failed tests. Useful for Fladle and other gradle plugins that don't expect the process to have a non-zero exit code. The JUnit XML is used to determine failure. (default: false)

What do you think of this more verbose description?

test_runner/src/main/kotlin/ftl/args/yml/FlankYml.kt
```diff
@@ -56,7 +59,8 @@ class FlankYmlParams(
             "project",
             "local-result-dir",
             "run-timeout",
-            "legacy-junit-result"
+            "legacy-junit-result",
+            "ignoreFailedTests"
```
bootstraponline, line 63: I think the key is `ignore-failed-tests`

README.md
pawelpasterz: Perfect, updated!

PR #731, opened 2020-04-22
730 - Refactor jUnit HTML report

Writhe:
This PR is a refactor of `/junit_html_report`. It replaces CRA with plain TS and simplifies the shape of data expected from Kotlin app (hence the WIP status).

~~**TO DO:**~~ **DONE:**
- change the output from Kotlin app that gets injected into the resulting HTML, to the following shape:
```
IResult {
  label: string
  url: string
}

IResultGroup {
  label: string
  items: IResult[]
}
```
- make it so it gets JSONified and injected into HTML, replacing "INJECT-DATA-HERE", preferably with quotes escaped.


junit_html_report/inline.js
```diff
@@ -1,14 +0,0 @@
...
-var inputFile = './build/index.html';
-var inputString = fs.readFileSync(inputFile, "utf8")
-
-inline.html({
-  fileContent: inputString,
-  relativeTo: "./build/"
-},
-function(err, result) {
-  if (err) { console.log(err); }
-  var kotlinResource = '../test_runner/src/main/resources/inline.html';
```
jan-gogo, line 12: I can't find replacement for this line. @Writhe are you sure that new implementation will update `inline.html` in `resources` automatically? I guess we don't want to do copy paste manually after update.

Writhe: Ah, good catch - it won't.
I don't think we need this file anywhere other than `resources`, so I can just tell Webpack to target that directory when building the thing. Would that be ok?

bootstraponline: > I can just tell Webpack to target that directory when building the thing. Would that be ok?

Yes, that will work. We don't use `inline.html` anywhere else.

Issue #739, opened 2020-04-24
Skip apks with 0 tests instead of failing


bootstraponline:
Issue reported via Slack. Giorgio will submit a pull request. Related issue in Fladle: https://github.com/runningcode/fladle/issues/96

> Giorgio Antonioli 
> Hi, currently Flank doesn’t support test APKs with zero tests on Android so currently we are detecting the test count in the bytecode before including the test APK with Flank. Would it be ok if I open a PR to implement this inside Flank? This would be easier since you are using DexParser already.
> 
> When you have a project with several modules and flank is configured automatically on each Android module. If you have several modules, not configuring Flank automatically for each module is dangerous since you usually want to run all your tests with your CI.
> 
> Imagine this scenario:
> 
> Module A has zero tests currently and Flank is not applied since there aren’t tests to run. A dev adds a test on module A. He should remember to add Flank manually if that is the first instrumented test of the module A. 99% of the time he will forget to do it and the tests won’t run on the CI for that given module

Issue #606, opened 2019-10-10
Android Hypershard example


bootstraponline:
Provide an example of using Hypershard for companies that are unable to use dex parsing on Android.

- https://github.com/dropbox/hypershard-android

The example should be described step by step in the markdown document.

bootstraponline:
Here's an example of how ktfmt parses kotlin code:

- https://github.com/facebookincubator/ktfmt/blob/59f7ad8d1fde08f3402a013571c9997316083ebf/core/src/main/java/com/facebook/ktfmt/Parser.kt#L37

jan-gogo:
https://github.com/dropbox/hypershard-android/issues/17 is this related issue?

bootstraponline:
> [dropbox/hypershard-android#17](https://github.com/dropbox/hypershard-android/issues/17) is this related issue?

yes

PR #741, opened 2020-04-24
Allow APKs with zero tests without throwing an exception

fondesa:
#### Description
This PR is a follow-up of our discussion on Slack about supporting APKs with zero tests.
Issue: #739

#### Motivation
This behaviour can be useful for projects with several Android modules.
Usually, in big projects, Flank is configured automatically on each Android module. Not configuring it automatically would be dangerous since you usually want to run all your tests on the CI without forgetting some of them.

#### Important notes

When the sharding is disabled, Flank already supported zero tests APK, since it supported invalid APKs. To be coherent with it, I reproduced the same behaviour when the sharding is enabled.

However, IMHO, it would be useful to distinguish an invalid APK from an APK with zero tests, regardless of whether the sharding is enabled or not. What do you think about this? In case, I could work on that too in another PR.


bootstraponline:
> However, IMHO, it would be useful to distinguish an invalid APK from an APK with zero tests, regardless of whether the sharding is enabled or not. What do you think about this? In case, I could work on that too in another PR.

I think invalid APKs should throw an exception. APKs with zero tests should be filtered out entirely and not uploaded to gcloud.

bootstraponline:
apk validation is done server side by FTL, we don't have a great way to detect if an APK is valid on the client. We can detect the test counts client side though.

fondesa:
> apk validation is done server side by FTL

Didn't know that, nice to know.

> we don't have a great way to detect if an APK is valid on the client.

Agree, in case we can still do a light validation as optimization counting the dex files through `DexParser.readDexFiles`. In this way, if an APK doesn't contain dex files, we can avoid to upload it on gcloud since we already know it's an invalid APK.

bootstraponline:
> Agree, in case we can still do a light validation as optimization counting the dex files through `DexParser.readDexFiles`. In this way, if an APK doesn't contain dex files, we can avoid to upload it on gcloud since we already know it's an invalid APK.

I can't trust DexParser in the future though, that's why disable sharding was added. That code path completely bypasses dex parser.
https://github.com/linkedin/dex-test-parser/issues/31

The Android bytecode evolves. I removed the APK validation check because when dex parser breaks, we need a release valve that doesn't involve being able to read the bytecode. The hypershard example we're working on is related. 

- https://github.com/dropbox/hypershard-android/issues/8
- https://github.com/Flank/flank/issues/606

fondesa:
Understood, thanks for the explanation.

BTW, let me know if you want me to make some changes to this PR.

bootstraponline:
Thanks! I will let @pawelpasterz and @jan-gogo review the code. We're trying to make sure there's at least two engineers looking at incoming pull requests.

The code looks reasonable.

test_runner/src/main/kotlin/ftl/run/platform/RunAndroidTests.kt
```diff
@@ -38,9 +39,15 @@ internal suspend fun runAndroidTests(args: AndroidArgs): TestResult = coroutineS
 
     args.resolveApks().forEach { apks: ResolvedApks ->
         val testShards = apks.test?.let { test ->
-            AndroidTestShard.getTestShardChunks(args, test).also {
-                allTestShardChunks += it
+            AndroidTestShard.getTestShardChunks(args, test)
+        }
+        // We can't return if testShards is null since it can be a robo test.
+        if (testShards != null) {
```
pawelpasterz, line 45: Thanks for implementing logic to skip empty apks!

If this is not a problem for you, could you consider rewriting this with (example) `testShards?.let {...}`
Regardless, I like the solution

test_runner/src/main/kotlin/ftl/run/platform/RunAndroidTests.kt
```diff
@@ -38,9 +39,15 @@ internal suspend fun runAndroidTests(args: AndroidArgs): TestResult = coroutineS
...
+            if (!testShards.haveAtLeastOneTest) {
+                // No tests to run, skipping the execution.
+                return@forEach
             }
+            allTestShardChunks += testShards
```
pawelpasterz, line 50: I am wondering if we could add only non-empty shards here, let me know what you think

fondesa: Actually, it will add only non-empty shards since otherwise `testShards.haveAtLeastOneTest` would return false and `return@forEach` will return before adding those shards

test_runner/src/main/kotlin/ftl/run/platform/RunAndroidTests.kt
fondesa: Absolutely, refactoring it!

test_runner/src/main/kotlin/ftl/run/platform/RunAndroidTests.kt
pawelpasterz: `haveAtLeastOneTest` uses `any {...}` which returns true if at least one item matches predicate

fondesa: I've got you, my bad, I didn't think about that case. I'm refactoring it thanks!

PR #757, opened 2020-04-30
Print version and revision before each command

### LINKED ISSUE #740: Print the version of Flank on each test run
Users often have no idea what version of Flank they're using. It'd be helpful if each Flank execution printed the current version of flank to stdout before running the tests.

Update Flank to always print a version (instead of the current unknown for local builds)

jan-gogo:
Fixes #740 

## Checklist

### DIFF
--- a/test_runner/build.gradle.kts
+++ b/test_runner/build.gradle.kts
@@ -5,6 +5,7 @@ import org.gradle.api.tasks.testing.logging.TestExceptionFormat
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 import java.util.Date
 import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
+import java.io.ByteArrayOutputStream
 
 plugins {
     application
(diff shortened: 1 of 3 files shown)


test_runner/build.gradle.kts
```diff
@@ -258,3 +258,17 @@ tasks.assemble {
     dependsOn(generateManpageAsciiDoc)
 }
 // end --- ASCII doc generation ---
+
+val updateVersion by tasks.registering {
+    shouldRunAfter(tasks.processResources)
+    doLast {
+        File("${project.buildDir}/resources/main/version.txt").writeText("local_snapshot")
+        File("${project.buildDir}/resources/main/revision.txt").writeText(
```
bootstraponline, line 266: The bitrise job does this:

```
echo "$BITRISE_GIT_TAG" > ./test_runner/src/main/resources/version.txt
echo "$GIT_CLONE_COMMIT_HASH" > ./test_runner/src/main/resources/revision.txt
```

and then the gradle commands are run. I think this may break the bitrise release job.

jan-gogo: Thanks for info. So I will use `BITRISE_IO` env for turning off this task when running on bitrise.

test_runner/build.gradle.kts
```diff
@@ -258,3 +259,17 @@ tasks.assemble {
     dependsOn(generateManpageAsciiDoc)
 }
 // end --- ASCII doc generation ---
+
+val updateVersion by tasks.registering {
+    shouldRunAfter(tasks.processResources)
+    if (!runningOnBitrise) doLast {
+        File("${project.buildDir}/resources/main/version.txt").writeText("local_snapshot")
+        File("${project.buildDir}/resources/main/revision.txt").writeText(
+            String(Runtime.getRuntime().exec("git rev-parse HEAD").inputStream.readBytes())
```
bootstraponline, line 268: should we exec on a background thread? here's an example:
https://github.com/bootstraponline/gradle_enterprise_example/blob/master/app/enterprise.gradle.kts#L152

also probably better to use `exec { ... }` instead of Runtime directly.

jan-gogo: Running on background from your example requires GradleEnterprisePlugin, TBH I have no experience with build scan. Any way, adding GradleEnterprisePlugin looks like standalone tasks, not a part of another. Also Updating those two files is pretty fast so it don't much impact on build time. 

Ok, I will use `exec { ... }`.

PR #761, opened 2020-05-03
Update api client

pawelpasterz:
Update Google API client


pawelpasterz:
> the json files should be in firebase_apis/json/ and fetched via update_api_json.sh

Hm...I did it via script but I had to do something wrong anyway :)

bootstraponline:
> Hm...I did it via script but I had to do something wrong anyway :)

> firebase_apis$ ./json/update_api_json.sh

I think if you invoke the script like that it'll save the files in the wrong folder. 😂 

I fixed this bug here: https://github.com/Flank/flank/commit/78e8f811a8e153b02191218d0e9f57c8f84a1ca7

PR #771, opened 2020-05-07
#764 Fix crash on parse some control chars

### LINKED ISSUE #764: Crash on XML parsing
Posted by Andrey Mischenko on Slack. Flank version unknown.

https://firebase-community.slack.com/archives/C1MTSQ5QT/p1588569925404500

We got invalid XML on one of our tests from test lab, it’s because Firebase converts some characters in failure message to html entities, which is invalid XML, which cause fails of our CI when try to get results
Invalid XML looks like:

```
<testcase name="test_name" classname="some.class.Name" time="23.194">
    <failure>java.net.ConnectException: Failed to connect to ... at &#8;&#8;&#8;(Coroutine boundary.&#8;(&#8;)
```

In original message we do not have html entities, just UTF character for Backspace https://unicode-table.com/en/0008/

I think it will be solved if Firebase will not encode html entities, or encode it using xml encoding using <!Entity> tag, or just replace &  with &amp;  it will not be perfect, but better than broken xml parsing (edited)

adamfilipow92:
Fixes #764

Fix crash on parse some control chars


test_runner/src/test/kotlin/ftl/reports/xml/JUnitXmlTest.kt
```diff
@@ -437,4 +440,32 @@ junit.framework.Assert.fail(Assert.java:50)</failure>
...
+                <testcase name="a()" classname="a" time="1.0">
+                <failure> java.net.ConnectException: Failed to connect to ... at &#8;&#8;&#8;(Coroutine boundary.&#8;(&#8;)</failure>
+                </testcase>
+                <testcase name="b()" classname="b" time="2.0"/>
+                <testcase name="c()" classname="c" time="7.0"/>
+              </testsuite>
+        """.trimIndent()
+
+        parseAllSuitesXml(crashingAllSuitesMessage)
+        parseOneSuiteXml(crashingOneSuiteMessage)
```
bootstraponline, line 469: should we also assert that the parsed value matches the expected value? Currently this test will pass as long as parsing doesn't fail. We probably want to assert ` <failure>...` is decoded correctly as well.

adamfilipow92: You have right, Im add asserts and split tests to two. One for one suite and one for all suites

test_runner/src/main/kotlin/ftl/reports/xml/JUnitXml.kt
```diff
@@ -6,20 +6,22 @@ import com.fasterxml.jackson.module.kotlin.KotlinModule
...
     .registerModules(KotlinModule())
     .configure(FAIL_ON_UNKNOWN_PROPERTIES, false)
 
 internal val xmlPrettyWriter = xmlMapper.writerWithDefaultPrettyPrinter()
 
-private fun xmlBytes(path: Path): ByteArray {
+private fun xmlText(path: Path): String {
     if (!path.toFile().exists()) throw RuntimeException("$path doesn't exist!")
-    return Files.readAllBytes(path)
+    return String(Files.readAllBytes(path))
```
bootstraponline, line 24: I'd like to merge https://github.com/Flank/flank/pull/767 before merging this pull request.

test_runner/src/main/kotlin/ftl/reports/xml/preprocesor/XmlPreprocessor.kt
```diff
@@ -0,0 +1,14 @@
+package ftl.reports.xml.preprocesor
+
+import org.apache.commons.text.StringEscapeUtils
+
+fun fixHtmlCodes(data: String): String {
+    val isoHtmlCodesToReplace = listOf(0x00..0x1F).union(listOf(0x7F..0x9F)).flatten()
```
bootstraponline, line 6: Is there some document we can reference in a comment about where these magic numbers come from?

test_runner/src/test/kotlin/ftl/reports/xml/JUnitXmlTest.kt
```diff
@@ -437,4 +441,64 @@ junit.framework.Assert.fail(Assert.java:50)</failure>
...
+                <testcase name="a()" classname="a" time="1.0">
+                  <failure> java.net.ConnectException: Failed to connect to ... at (Coroutine boundary.()</failure>
+                </testcase>
+                <testcase name="b()" classname="b" time="2.0"/>
+                <testcase name="c()" classname="c" time="7.0"/>
+              </testsuite>
+            </testsuites>
+        """.trimIndent()
+        val oneSuiteXml = parseOneSuiteXml(crashingOneSuiteMessage).xmlToString().trimIndent()
+        Assert.assertEquals("One Suite Messages should be the same!", expectedOneSuiteMessage, oneSuiteXml)
```
bootstraponline, line 502: The tests look great! Thanks for updating these.

test_runner/src/main/kotlin/ftl/reports/xml/JUnitXml.kt
jan-gogo: > I'd like to merge #767 before merging this pull request.

Done

test_runner/src/main/kotlin/ftl/reports/xml/preprocesor/XmlPreprocessor.kt
bootstraponline: The other thought I had is, are we sure there's no helper already defined in a library that does what we need? Maybe this is a special case so the custom code is justified.

adamfilipow92: I this case we need to get list of control chars and convert them to html codes with avoid white chars. I cant find method doing this. 
About magic numbers: Im added enum with I hope intuitive names. Enum on the top has comment about destination and link to utf-8 table to prove my idea.

test_runner/src/main/kotlin/ftl/reports/xml/preprocesor/XmlPreprocessor.kt
```diff
@@ -0,0 +1,17 @@
+package ftl.reports.xml.preprocesor
+
+import org.apache.commons.text.StringEscapeUtils
+
+fun fixHtmlCodes(data: String): String {
```
jan-gogo, line 5: IMO expressions > blocks ;)
```
fun fixHtmlCodes(data: String): String = listOf(
    UtfControlCharRanges.CONTROL_TOP_START.charValue..UtfControlCharRanges.CONTROL_TOP_END.charValue,
    UtfControlCharRanges.CONTROL_BOTTOM_START.charValue..UtfControlCharRanges.CONTROL_BOTTOM_END.charValue
).flatten()
    .map { StringEscapeUtils.escapeXml11(it.toChar().toString()) }
    .filter { it.startsWith("&#") }
    .fold(data) { fixedStr: String, isoControlCode: String -> fixedStr.replace(isoControlCode, "") }
```
unions are not mandatory for those values

test_runner/buildSrc/src/main/kotlin/Deps.kt
```diff
@@ -74,6 +74,8 @@ object Versions {
 
     // https://github.com/mockk/mockk
     const val MOCKK = "1.9.3"
+
+    const val COMMON_TEXT = "1.7"
```
pawelpasterz, line 78: Please, add here link to repo or equivalent, thanks.

test_runner/src/main/kotlin/ftl/reports/xml/preprocesor/XmlPreprocessor.kt
pawelpasterz: I am also for @jan-gogo version 👍

adamfilipow92: Thanks! Changed to better version :)

test_runner/buildSrc/src/main/kotlin/Deps.kt
adamfilipow92: Comment added

adamfilipow92:
> Please remove `should_exists.txt` file from PR.

Removed :)

pawelpasterz:
@adamfilipow92 
Nice work! Remember to squash commits before you merge, thanks!

PR #783, opened 2020-05-14
Use legacy results for iOS by default

### LINKED ISSUE #652: iOS: JUnitReport.xml contains data from previously used device
case:
* run test with iphone x on ios 12
* change device to iphone 11 with ios 13
* run tests
-> JUnitReport.xml contains results from iphone x, which was not used in last execution

pawelpasterz:
Fixes #652

## Checklist

- [x] Unit tested


test_runner/src/main/kotlin/ftl/args/IosArgs.kt
```diff
@@ -69,6 +69,8 @@ class IosArgs(
     override val networkProfile = cli?.networkProfile ?: gcloud.networkProfile
     override val ignoreFailedTests = cli?.ignoreFailedTests ?: flank.ignoreFailedTests
     override val keepFilePath = cli?.keepFilePath ?: flank.keepFilePath
+    // currently, FTL does not provide API based results for iOS
+    override val useLegacyJUnitResult = true
```
jan-gogo, line 73: Thanks for this fix. Shame on me :/

PR #784, opened 2020-05-15
Add output-style option

jan-gogo:
Fixes #779 

Adds option for output style 

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/test_runner/docs/ascii/flank.jar_-android-run.adoc
+++ b/test_runner/docs/ascii/flank.jar_-android-run.adoc
@@ -26,7 +26,8 @@ flank.jar
                        [*--num-flaky-test-attempts*=_<flakyTestAttempts>_]
                        [*--num-test-runs*=_<repeatTests>_]
                        [*--num-uniform-shards*=_<numUniformShards>_]
-                       [*--project*=_<project>_] [*--results-bucket*=_<resultsBucket>_]
+                       [*--output-style*=_<outputStyle>_] [*--project*=_<project>_]
+                       [*--results-bucket*=_<resultsBucket>_]
                        [*--results-dir*=_<resultsDir>_]
                        [*--results-history-name*=_<resultsHistoryName>_]
                        [*--robo-script*=_<roboScript>_]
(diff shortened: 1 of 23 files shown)


test_runner/docs/ascii/flank.jar_-android-run.adoc
```diff
@@ -142,6 +143,9 @@ Configuration is read from flank.yml
 *--keep-file-path*::
   Keeps the full path of downloaded files. Required when file names are not unique.
 
+*--output-style*=_<outputStyle>_::
+  Output style of execution status. Maybe be one of [verbose, multi, single]For runs with only one test execution default value is 'verbose', in other cases'multi' is used as default. Output style 'multi' is not displayed correctly on consoleswhich not supports ansi codes, to avoid corrupted output use `single` or `verbose`.
```
bootstraponline, line 147: > Output style of execution status. May be one of [verbose, multi, single] For runs with only one test execution the default value is 'verbose', in other cases 'multi' is used as the default. The output style 'multi' is not displayed correctly on consoles which don't support ansi codes, to avoid corrupted output use `single` or `verbose`.

bootstraponline:
I love the output style ported from gcloud. 🙂

bootstraponline:
> #### Conflicting files
> `test_runner/src/main/kotlin/ftl/args/IArgs.kt`

test_runner/src/main/kotlin/ftl/args/IArgs.kt
```diff
@@ -40,7 +40,11 @@ interface IArgs {
     val useLegacyJUnitResult: Boolean get() = false
     val ignoreFailedTests: Boolean
     val keepFilePath: Boolean
-    val outputStyle: OutputStyle? get() = null
+    val outputStyle: OutputStyle
+    val defaultOutputStyle
+        get() = if (flakyTestAttempts > 0 || (!disableSharding && maxTestShards > 0))
```
bootstraponline, line 45: what about additional test apks? If we have additional test apks, then that'll involve multiple matrices.

PR #786, opened 2020-05-15
[Do not merge] | Minimal setup required to run cucumber

dbarwacz:
Reproduces crash when running cucumber tests on flank. 
Tests run on firebase directly through AS picks up cucumber tests correctly and executes them to completion.


bootstraponline:
Thanks! This will help us investigate the problem.

bootstraponline:
Feedback on slack is that disable sharding & specifying the correct test runner custom resolves the issue. https://github.com/linkedin/dex-test-parser isn't able to discover cucumber tests. I think this is the best approach for now.

```yaml
gcloud:
    test-runner-class: cucumber.api.android.CucumberAndroidJUnitRunner

flank:
    disable-sharding: true
```

dbarwacz:
I checked out my changes branch here again and: 
* I used custom runner (without it the tests would just pass because they would be unaware of cucumber tests)
* I disabled sharding (which crashed at the setup phase without firebase being even reached). 

The only notable difference was that I configured junit runner at gradle level, while you suggest using it at flank.yml level (which to be honest should be the same, as flank docs specify that default value of `test-runner-class` falls back to Manifest value, which my gradle config sets up).

I wanted to inquire more to see how slack feedback differed because I believe https://github.com/Flank/flank/pull/786#issuecomment-630376993 was what I was using.

bootstraponline:
> The only notable difference was that I configured junit runner at gradle level, while you suggest using it at flank.yml level (which to be honest should be the same, as flank docs specify that default value of `test-runner-class` falls back to Manifest value, which my gradle config sets up).

I'd try it with the flank.yml level. The flank.yml is the only config that Flank controls. FTL server has behavior changes over time. If you're still seeing problems then I'm happy to reopen.

dbarwacz:
Just to rule out the possibility that I'm missing something obvious, this is flank.yml I'm using: 
```
AndroidArgs
    gcloud:
      results-bucket: test-lab-4nv5xxtjty0h6-isys7vj9jhnm2
      results-dir: null
      record-video: false
      timeout: 15m
      async: false
      client-details: 
      network-profile: null
      results-history-name: null
      # Android gcloud
      app: /Users/dombar/flank/test_app/app/build/outputs/apk/singleSuccess/debug/app-single-success-debug.apk
      test: /Users/dombar/flank/test_app/app/build/outputs/apk/androidTest/singleSuccess/debug/app-single-success-debug-androidTest.apk
      additional-apks: 
      auto-google-login: false
      use-orchestrator: false
      directories-to-pull:
        - /sdcard/
      other-files:
      performance-metrics: false
      num-uniform-shards: null
      test-runner-class: com.example.test_app.CucumberAndroidJUnitRunner
      test-targets:
      robo-directives:
      robo-script: null
      device:
        - model: NexusLowRes
          version: 28
          locale: en
          orientation: portrait
      num-flaky-test-attempts: 0

    flank:
      max-test-shards: 1
      shard-time: -1
      num-test-runs: 1
      smart-flank-gcs-path: 
      smart-flank-disable-upload: false
      files-to-download:
        - .*/sdcard/[^/]+\.ec$
      test-targets-always-run:
      disable-sharding: true
      project: PROBABLY_NOT_PUBLIC
      local-result-dir: results
      # Android Flank Yml
      keep-file-path: false
      additional-app-test-apks:
      run-timeout: -1
      legacy-junit-result: false
      ignore-failed-tests: false
```

It still fails at the same step with the same error message as previously.

bootstraponline:
Thanks for testing again. I've reopened the PR so the team can take a look.

bootstraponline:
Added `android.useAndroidX=true` to `gradle.properties`

```
$ gradle assemble assembleAndroidTest
```

```
gcloud:
  app:  ./app/build/outputs/apk/singleSuccess/debug/app-single-success-debug.apk
  test: ./app/build/outputs/apk/androidTest/singleSuccess/debug/app-single-success-debug-androidTest.apk
```

```
$ gcloud firebase test android run flank.yml:gcloud
```

>  Passed  │ walleye-26-en-portrait │ 1 test cases passed

Confirmed that the test passes using gcloud. A single test is run called `Launch new screens`

Gcloud:

- Client gcloud
- TestRunnerClass cucumber.api.android.CucumberAndroidJUnitRunner
- RanWithOrchestrator false

Confirmed that the test fails using flank.

Flank:

- Client Flank
- TestRunnerClass cucumber.api.android.CucumberAndroidJUnitRunner
- TestTargets class com.example.test_app.InstrumentedTest#test
- RanWithOrchestrator true

The bug is we are setting test targets when disable-sharding is true. I've opened https://github.com/Flank/flank/issues/815

bootstraponline:
https://github.com/bootstraponline/flank_bug problem still reproduces on master.

bootstraponline:
I confirmed this issue is fixed on flank master and will be resolved in the next release.

Issue #718, opened 2020-04-14
Flank Refactor


bootstraponline:
Refactor proposal from yaneq (copied from Slack)

1. ~~Building simple config data class from cli and config file which will be a context for next step.~~
1. ~~Validating config data class~~
1. ~~Building domain structures which should explicitly specify how to create test matrices.~~
1. ~~Uploading all required files asynchronously and updating file references in domain structures.~~
1. ~~Building and running gc test matrices from domain structures.~~
1. ~~Pooling results.~~
1. ~~Generating reports.~~ (Partially finished)

Of course flank now fulfills those steps but there are not fully visible at first look, and some of responsibilities are mixed.

Additionally:
- We should avoid operations on mutable data.
- All gc operations can be hidden behind interfaces
- We should use Displatchers.IO only in gc api layer.
- Because flank's run has relatively simple lifecycle, we should avoid OOP, and rely on functions and structures. As Edsger Dijkstra said: "Object-oriented programming is an exceptionally bad idea", I am fully agreed with this :).

runningcode:
Can we maintain the current API and outputs of Flank? I worry a refactor would change these and therefore leave certain projects behind which rely on the API and outputs of flank and are unable to migrate.

bootstraponline:
> Can we maintain the current API and outputs of Flank?

I agree, I see this as an internal refactor. We should make sure to maintain the current outputs and not break [fladle](https://github.com/runningcode/fladle)

PR #799, opened 2020-05-19
#707 Refactor Shard object by splitting it into smaller functions

### LINKED ISSUE #707: Refactor shard object to make it easier to understand
yaneq mentioned the legacy code is difficult to reason about. Let's refactor!

https://github.com/Flank/flank/blob/master/test_runner/src/main/kotlin/ftl/shard/Shard.kt

piotradamczyk5:
Fixes #707 Refactor Shared object by splitting it into smaller functions

## Checklist

- [x] Documented
- [x] Unit tested


jan-gogo:
Also `shardCountByTime` function needs more work, there are 4 returns and 1 hidden throw + some assignments (value assignments are good for writing but not for reading). 

I was thinking about something like that but without magic numbers:
```
fun shardCountByTime(
    testsToRun: List<FlankTestMethod>,
    oldTestResult: JUnitTestResult,
    args: IArgs
): Int = when {
    args.shardTime == -1 -> -1
    args.shardTime < -1 || args.shardTime == 0 -> throw FlankFatalError("Invalid shard time ${args.shardTime}")
    else -> testsToRun.testTotalTime(oldTestResult, args).let { testsTotalTime ->
        if (testsTotalTime <= args.shardTime) -1
        else min(ceil(testsTotalTime / args.shardTime).toInt(), args.maxTestShards)
    }
}

private fun List<FlankTestMethod>.testTotalTime(
    oldTestResult: JUnitTestResult,
    args: IArgs
) = sumByDouble {
    if (it.ignored) Shard.IGNORE_TEST_TIME
    else createTestMethodDurationMap(oldTestResult, args)[it.testName] ?: Shard.DEFAULT_TEST_TIME_SEC
}
```
Additionally flank now is using ftl api to run shards, and this api has restriction on maximum test shards up to 50. So below code is not valid now:
```
        // If there is no limit, use the calculated amount
        if (args.maxTestShards == -1) {
            return shardsByTime
        }
```

test_runner/src/main/kotlin/ftl/shard/Shard.kt
```diff
@@ -100,68 +111,87 @@ object Shard {
...
+    private fun addTestMethodToMostEmptyShard(shards: List<TestShard>, testMethod: TestMethod) {
+        val mostEmptyShard = shards.first()
+        mostEmptyShard.testMethods.add(testMethod)
+        mostEmptyShard.time += testMethod.time
+    }
+
+    private val IArgs.platformName
+        get() = if (this is IosArgs) "ios" else "android"
+
+    private fun assertNotTrue(condition: Boolean, lazyException: () -> Throwable) {
```
jan-gogo, line 175: I am not sure if this abstraction is really needed.

jan-gogo:
`createShardsByShardCount` is still breaks SRP. As I see, it makes at least 3 things. 
1. It creates `testsCases` and `cacheMiss`. 
2. It creates shards from testCases 
3. prints info.

bootstraponline:
> Additionally flank now is using ftl api to run shards, and this api has restriction on maximum test shards up to 50.

FYI the API restriction is changing soon to 250. I don't want to hardcode the API restriction in Flank as Firebase can change this anytime.

jan-gogo:
> > Additionally flank now is using ftl api to run shards, and this api has restriction on maximum test shards up to 50.
> 
> FYI the API restriction is changing soon to 250. I don't want to hardcode the API restriction in Flank as Firebase can change this anytime.

We have `IArgs.AVAILABLE_SHARD_COUNT_RANGE` value, it is used to validate possible maximum shards on config level. Without hardcoded max value we are losing ability to check it early so It will be done later on ftl side.

bootstraponline:
> We have `IArgs.AVAILABLE_SHARD_COUNT_RANGE` value, it is used to validate possible maximum shards on config level.

We'll likely have to revisit this design. 50 is currently the global limit for both physical and virtual devices. In the future 250 will be the limit for virtual devices. Physical devices will still be limited to 50.

> Without hardcoded max value we are losing ability to check it early so It will be done later on ftl side.

That's a good point. Let's keep the hard coded values.

Ideally FTL would expose an API for us to fetch the shard limits so we don't have to push a new Flank release every time. Probably the numbers change infrequently enough this won't be a big deal. If they did keep changing it then possibly https://firebase.google.com/docs/remote-config would be helpful.

test_runner/src/main/kotlin/ftl/shard/Shard.kt
piotradamczyk5: I will remove this

test_runner/src/main/kotlin/ftl/shard/Shard.kt
```diff
@@ -100,68 +111,87 @@ object Shard {
         args: IArgs,
         forcedShardCount: Int = -1
     ): List<TestShard> {
-        if (forcedShardCount < -1 || forcedShardCount == 0) throw FlankFatalError("Invalid forcedShardCount value $forcedShardCount")
-
+        assertNotTrue((forcedShardCount < -1 || forcedShardCount == 0)) {
```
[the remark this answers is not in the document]
piotradamczyk5: removed

piotradamczyk5:
> Additionally flank now is using ftl api to run shards, and this api has restriction on maximum test shards up to 50. So below code is not valid now:
> 
> ```
>         // If there is no limit, use the calculated amount
>         if (args.maxTestShards == -1) {
>             return shardsByTime
>         }
> ```

so what should be done here? remove this condition?

bootstraponline:
> so what should be done here? remove this condition?

I think we should update to respect the limit defined in AVAILABLE_SHARD_COUNT_RANGE, then in a future PR we'll have to solve for the 250 limit once FTL makes this change.

piotradamczyk5:
> AVAILABLE_SHARD_COUNT_RANGE



> > so what should be done here? remove this condition?
> 
> I think we should update to respect the limit defined in AVAILABLE_SHARD_COUNT_RANGE, then in a future PR we'll have to solve for the 250 limit once FTL makes this change.

so new condition will look 
        // If there is no limit, use the calculated amount
        if (args.maxTestShards == NO_LIMIT) {
            return AVAILABLE_SHARD_COUNT_RANGE.last
        }

test_runner/src/main/kotlin/ftl/shard/CacheReport.kt
```diff
@@ -0,0 +1,22 @@
+package ftl.shard
+
+import ftl.util.FlankTestMethod
+import kotlin.math.roundToInt
+
+object CacheReport {
```
jan-gogo, line 6: No object needed

test_runner/src/main/kotlin/ftl/shard/Shard.kt
```diff
@@ -46,52 +43,6 @@ class com.foo.ClassName#testMethodToSkip
 */
 
 object Shard {
```
jan-gogo, line 40: No object needed

test_runner/src/main/kotlin/ftl/shard/ShardCount.kt
```diff
@@ -0,0 +1,53 @@
...
+import ftl.shard.TestMethodDuration.testTotalTime
+import ftl.util.FlankFatalError
+import ftl.util.FlankTestMethod
+import kotlin.math.ceil
+import kotlin.math.min
+
+private const val SINGLE_SHARD = 1
+private const val NO_LIMIT = -1
+
+object ShardCount {
```
jan-gogo, line 16: No object needed

test_runner/src/main/kotlin/ftl/shard/TestCasesCreator.kt
```diff
@@ -0,0 +1,24 @@
+package ftl.shard
+
+import ftl.util.FlankTestMethod
+
+data class TestMethod(
+    val name: String,
+    val time: Double
+)
+
+object TestCasesCreator {
```
jan-gogo, line 10: No object needed

test_runner/src/main/kotlin/ftl/shard/TestMethodDuration.kt
```diff
@@ -0,0 +1,45 @@
+package ftl.shard
+
+import ftl.args.AndroidArgs
+import ftl.args.IArgs
+import ftl.reports.xml.model.JUnitTestCase
+import ftl.reports.xml.model.JUnitTestResult
+import ftl.util.FlankTestMethod
+
+object TestMethodDuration {
```
jan-gogo, line 9: No object needed

test_runner/src/main/kotlin/ftl/shard/TestMethodDuration.kt
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    private fun JUnitTestCase.iosKey(): String {
+        // FTL iOS XML appends `()` to each test name. ex: `testBasicSelection()`
+        // xctestrun file requires classname/name with no `()`
+        val testName = name?.substringBefore('(')
+        return "$classname/$testName"
+    }
+
+    fun createTestMethodDurationMap(junitResult: JUnitTestResult, args: IArgs): Map<String, Double> {
```
jan-gogo, line 22: Its good idea to sort functions from most to less important. `createTestMethodDurationMap` is more important then `androidKey` and `iosKey` cause is public.

test_runner/src/test/kotlin/ftl/shard/TestMethodDurationTest.kt
```diff
@@ -0,0 +1,57 @@
...
+        val previousMapDuration = testMethods.associate { it.testName to previousTestTimeSeconds }
+        val expectedTestTime = testMethods.size * previousTestTimeSeconds
+
+        // when
+        val calculatedTotalTestTime = TestMethodDuration.testTotalTime(testMethods, previousMapDuration)
+
+        // then
+        assertThat(calculatedTotalTestTime).isEqualTo(expectedTestTime)
+    }
+}
```
jan-gogo: In Intellij check `Settings -> Editor -> General -> Ensure an empty line at the end of file on Save`

test_runner/src/test/kotlin/ftl/shard/TestCasesCreatorTest.kt
```diff
@@ -0,0 +1,62 @@
...
+
+        // when
+        val createdTestCases = TestCasesCreator.createTestCases(testMethods, previousMapDuration)
+
+        // then
+        createdTestCases.forEach {
+            assertThat(it.time).isEqualTo(previousTestTimeSeconds)
+        }
+    }
+}
```
jan-gogo: add empty line

test_runner/src/test/kotlin/ftl/shard/ShardTestCommon.kt
```diff
@@ -0,0 +1,37 @@
...
+}
+
+internal fun listOfFlankTestMethod(vararg args: String) = listOf(*args).map { FlankTestMethod(it) }
+
+internal fun mockArgs(maxTestShards: Int, shardTime: Int = 0): IArgs {
+    val mockArgs = mockk<IosArgs>()
+    every { mockArgs.maxTestShards } returns maxTestShards
+    every { mockArgs.shardTime } returns shardTime
+    return mockArgs
+}
```
jan-gogo: missing empty line

test_runner/src/test/kotlin/ftl/shard/ShardCountTest.kt
```diff
@@ -0,0 +1,74 @@
...
+        val testsToRun = listOfFlankTestMethod("i/i", "j/j")
+        val suite = sample()
+        val result = ShardCount.shardCountByTime(
+            testsToRun,
+            suite,
+            mockArgs(maxTestShards = -1, shardTime = (DEFAULT_TEST_TIME_SEC * 2).toInt()))
+
+        assertThat(result).isEqualTo(1)
+    }
+}
```
jan-gogo: missing empty line

test_runner/src/main/kotlin/ftl/shard/CacheReport.kt
```diff
@@ -0,0 +1,22 @@
...
+    }
+
+    private fun calculateCacheMiss(testsToRun: List<FlankTestMethod>, previousMethodDurations: Map<String, Double>): Int {
+        return testsToRun.count { !previousMethodDurations.containsKey(it.testName) }
+    }
+    
+    private fun cacheHit(allTestCount: Int, cacheMiss: Int) = allTestCount - cacheMiss
+
+    private fun cachePercent(allTestCount: Int, cacheHit: Int): Double = if (allTestCount == 0) 0.0 else cacheHit.toDouble() / allTestCount * 100.0
+}
```
jan-gogo: https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-a-source-file

test_runner/src/main/kotlin/ftl/shard/ShardCount.kt
```diff
@@ -0,0 +1,53 @@
...
+            return shardsByTime
+        }
+
+        // We need to respect the maxTestShards
+        val shardCount = min(shardsByTime, args.maxTestShards)
+
+        if (shardCount <= 0) throw FlankFatalError("Invalid shard count $shardCount")
+
+        return shardCount
+    }
```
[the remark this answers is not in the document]
piotradamczyk5: fixed

test_runner/src/main/kotlin/ftl/shard/Shard.kt
```diff
@@ -45,137 +34,77 @@ class com.foo.ClassName#testMethodToSkip
...
-            shard.time += testMethod.time
+    printCacheInfo(testsToRun, previousMethodDurations)
+    printShardsInfo(shards)
+    return shards
+}
 
-            // Sort the shards to keep the most empty shard first
-            shards = shards.sortedBy { it.time }
-        }
+private fun createShardsForTestCases(testCases: List<TestMethod>, shardsCount: Int): List<TestShard> {
```
jan-gogo, line 75: I guess this one can be reduced to
```
private fun createShardsForTestCases(
    testCases: List<TestMethod>,
    shardsCount: Int
): List<TestShard> = testCases.fold(
    initial = List(shardsCount) { TestShard(0.0, mutableListOf()) }
) { shards, testMethod ->
    shards.sortedBy(TestShard::time).apply {
        first().apply {
            testMethods += testMethod
            time += testMethod.time
        }
    }
}
```

test_runner/src/main/kotlin/ftl/shard/TestMethodDuration.kt
```diff
@@ -0,0 +1,35 @@
...
+            testsuite.testcases
+                ?.asSequence()
+                ?.filter { testcase -> !testcase.empty() && testcase.time != null }
+                ?.map { testcase ->
+                    val key = if (args is AndroidArgs) testcase.androidKey() else testcase.iosKey()
+                    val time = testcase.time!!.toDouble()
+                    key to time
+                }
+                ?.filter { (_, time) -> time >= 0 }
+                ?.forEach { (key, time) -> this[key] = time }
```
jan-gogo, line 21: This new version adds 5 additional unnecessary null checks and 5 lines of code more. Also I am not sure if this filters and maps are necessary if at last we still have forEach and additions to map. IMO it is not simpler than previous version. Pls rollback this change, and I will approve pull request

test_runner/src/main/kotlin/ftl/shard/Shard.kt
```diff
@@ -45,137 +34,74 @@ class com.foo.ClassName#testMethodToSkip
...
+) { shards, testMethod ->
+    shards.sortedBy(TestShard::time).apply {
+        first().apply {
+            testMethods += testMethod
+            time += testMethod.time
+        }
     }
 }
+
+private fun createListOfShards(shardsCount: Int) = List(shardsCount) { TestShard(0.0, mutableListOf()) }
```
jan-gogo, line 107: I am not sure if this is necessary

test_runner/src/main/kotlin/ftl/shard/ShardCount.kt
```diff
@@ -0,0 +1,53 @@
...
+    else -> calculateShardCount(testsToRun, oldTestResult, args)
+}
+
+private fun calculateShardCount(
+    testsToRun: List<FlankTestMethod>,
+    oldTestResult: JUnitTestResult,
+    args: IArgs
+): Int = calculateShardCount(
+    args = args,
+    testsTotalTime = testTotalTime(testsToRun, createTestMethodDurationMap(oldTestResult, args)),
```
jan-gogo, line 31: 👍

test_runner/src/main/kotlin/ftl/shard/TestMethodDuration.kt
piotradamczyk5: reverted. 
What I have in my mind is that previously there was also null checks and filters and maps change the code sequential way

test_runner/src/main/kotlin/ftl/shard/Shard.kt
piotradamczyk5: removed

release_notes.md
```diff
@@ -2,10 +2,12 @@
 
 - [#784](https://github.com/Flank/flank/pull/784) Add output-style option. ([jan-gogo](https://github.com/jan-gogo))
 - [#779](https://github.com/Flank/flank/pull/779) Print retries & display additional info. ([jan-gogo](https://github.com/jan-gogo))
-- [#793](https://github.com/Flank/flank/issues/793) Better error message on file not found. ([adamfilipow92](https://github.com/adamfilipow92))
-- [#808](https://github.com/Flank/flank/issues/808) Fixed dry run crashes. ([piotradamczyk5](https://github.com/piotradamczyk5))
-- [#807](https://github.com/Flank/flank/issues/807) Fix Bugsnag being initialized during tests. ([piotradamczyk5](https://github.com/piotradamczyk5))
+- [#793](https://github.com/Flank/flank/issues/793) Better error message on file not found. ([adamfilipow92](https://github.com/adamfilipow92)) 
+- [#808](https://github.com/Flank/flank/issues/808) Fixed dry run crashes. ([piotradamczyk5](https://github.com/piotradamczyk5))
```
bootstraponline, line 6: trailing spaces?

test_runner/src/main/kotlin/ftl/shard/TestMethodDuration.kt
```diff
@@ -0,0 +1,35 @@
...
+        testsuite.testcases?.forEach { testcase ->
+            if (!testcase.empty() && testcase.time != null) {
+                val key = if (args is AndroidArgs) testcase.androidKey() else testcase.iosKey()
+                val time = testcase.time.toDouble()
+                if (time >= 0) junitMap[key] = time
+            }
+        }
+    }
+
+    return junitMap
```
pawelpasterz, line 23: I think we could return here instance of read only map to avoid potential, unwanted operations on returned object.

test_runner/src/main/kotlin/ftl/shard/TestMethodDuration.kt
```diff
@@ -0,0 +1,35 @@
...
+import ftl.args.IArgs
+import ftl.reports.xml.model.JUnitTestCase
+import ftl.reports.xml.model.JUnitTestResult
+
+fun createTestMethodDurationMap(junitResult: JUnitTestResult, args: IArgs): Map<String, Double> {
+    val junitMap = mutableMapOf<String, Double>()
+
+    // Create a map with information from previous junit run
+
+    junitResult.testsuites?.forEach { testsuite ->
```
pawelpasterz, line 13: How about rewriting this to more functional style? Something like
```
return junitResult.testsuites?.asSequence()
        ?.mapNotNull { it.testcases }?.flatten()
        ?.filterNot { it.empty() || it.time == null }
        ?.mapNotNull { it.time?.run { (if (args is AndroidArgs) it.androidKey() else it.iosKey()) to it.time.toDouble() } }
        ?.filter { it.second >= 0 }
        ?.toMap() ?: emptyMap()
```

It's not change request, just thinking.

jan-gogo: No, no, no!!! xD

pawelpasterz: > How about (...)

...or not :)

piotradamczyk5: so no :)

test_runner/src/test/kotlin/ftl/shard/ShardCountTest.kt
```diff
@@ -0,0 +1,74 @@
+package ftl.shard
+
+import com.google.common.truth.Truth.assertThat
+import ftl.test.util.FlankTestRunner
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(FlankTestRunner::class)
+class ShardCountTest {
```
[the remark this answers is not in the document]
piotradamczyk5: fixed

test_runner/src/main/kotlin/ftl/shard/Shard.kt
piotradamczyk5: fixed

release_notes.md
piotradamczyk5: fixed

bootstraponline:
> 1). IMO we should continue our effort with second part, maybe not within this issue but followup one

Agreed. We can consider converting the Flank refactor (https://github.com/Flank/flank/issues/718) ticket into an Epic composed of multiple tickets. I'd like to get the remaining features/bug fixes out and then focus on the refactor. What do you think?

Issue #792, opened 2020-05-16
Test results from shards with overlapping names in different matrices are lost


andrey-bolduzev:
_The below is a speculation based on the data I've been able to collect and interpret so far._

I'm trying to upgrade a fairly sized multi-module project to the latest flank version (via fladle) with minimal config changes but so far I'm unsuccessful at that, not sure if I'm doing something wrong.

The fladle plugin is applied to the top-level `build.gradle` and extra test apk's from the submodules are pulled in with `additionalTestApks`. After all the tests are run, we download sdcard contents of all shards from the returned combined test bucket. Specifically in our case, we are interested in `coverage.ec` files from all the sdcard, which we merge together with the help of the jacoco gradle plugin.

After upgrade to the new flank version, the coverage dropped suddenly and significantly. What I think is happening is files from different gradle modules are overwriting each other in the combined results bucket. So for instance for `:module1` 3 shards are created and collected in `Pixel2-29-en-portrait-shard_0/`, `Pixel2-29-en-portrait-shard_1/`, `Pixel2-29-en-portrait-shard_2/` directories in the combined bucket. Then say :module2 is large enough to have 2 shards created for its tests: `Pixel2-29-en-portrait-shard_0/`, `Pixel2-29-en-portrait-shard_1/`. All the directories with the overlapping names get copied over to the same bucket and the last one wins, so the coverage execution data from the "losing" shards is lost. There's also a new `Pixel2-29-en-portrait-test_results_merged.xml` file in the results bucket, which also doesn't contain all tests being run which confirms my suspicion that there are multiple `Pixel2-29-en-portrait-test_results_merged.xml` files coming from different matrices and this is just one survivor of them.

pawelpasterz:
@bootstraponline 
[...]

PR #801, opened 2020-05-20
Omit missing app apk if additional-app-test-apks specified

### LINKED ISSUE #800: Ignore missing app & test if additional-app-test-apks are specified
https://github.com/runningcode/fladle/issues/96#issuecomment-629495470

jan-gogo:
Fixes #800 

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
+++ b/test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
@@ -32,6 +32,7 @@ import ftl.config.FtlConstants
 import ftl.config.parseRoboDirectives
 import ftl.run.status.asOutputStyle
 import ftl.util.FlankFatalError
+import java.io.File
 import java.io.Reader
 import java.nio.file.Path
 
@@ -104,9 +105,12 @@ class AndroidArgs(
     override val hasMultipleExecutions: Boolean get() = super.hasMultipleExecutions || additionalAppTestApks.isNotEmpty()
 
     init {
-        if (appApk == null) additionalAppTestApks.filter { (app, _) -> app == null }.let { missingApks ->
-            if (missingApks.isNotEmpty()) throw FlankFatalError("Cannot resolve app apk pair for ${missingApks.map { it.test }}")
-        }
+        if (appApk == null) additionalAppTestApks
+            .filter { (app, _) -> app == null }
+            .map { File(it.test).name }
+            .run {
+                if (isNotEmpty()) throw FlankFatalError("Cannot resolve app apk pair for $this")
+            }
 
         resultsBucket = createGcsBucket(project, cli?.resultsBucket ?: gcloud.resultsBucket)
         createJunitBucket(project, flank.smartFlankGcsPath)
--- a/test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
+++ b/test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
@@ -55,8 +55,8 @@ class AndroidArgs(
     override val flakyTestAttempts = cli?.flakyTestAttempts ?: gcloud.flakyTestAttempts
 
     private val androidGcloud = androidGcloudYml.gcloud
-    var appApk = (cli?.app ?: androidGcloud.app ?: throw FlankFatalError("app is not set")).processFilePath("from app")
-    var testApk = (cli?.test ?: androidGcloud.test)?.processFilePath("from test")
+    var appApk: String? = (cli?.app ?: androidGcloud.app)?.processFilePath("from app")
+    var testApk: String? = (cli?.test ?: androidGcloud.test)?.processFilePath("from test")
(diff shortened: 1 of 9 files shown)


bootstraponline:
Should we also omit missing test apk when using additiona-app-test-apks? I think Fladle wanted to specify all the app/test apks via additional-app-test-apks.

runningcode:
Hey, thank you very much for doing this. Yes @bootstraponline is correct that we want to specify all app/test apks via additional app test apks.

That being said, and I really appreciate this change, after having implemented [the plugin](https://runningcode.github.io/fladle/) I think I am unlikely to take advantage of this API for now.  Many of the largers companies who require automatically specifying their additional test apks will be stuck on flank 8.1.0 for the foreseeable future until firebase test lab supports more than 50 shards. If Fladle were to take advantage of this new API, it would not be backward compatible anymore.

That being said, I really hope google will change their mind about the 50 shard limitation and this api will greatly simplify this part of the code: https://github.com/runningcode/fladle/blob/master/buildSrc/src/main/java/com/osacky/flank/gradle/FulladlePlugin.kt#L42

bootstraponline:
@runningcode I already persuaded FTL to increase the limit, it will be raised to 250 in a few weeks. We will probably be breaking backwards compatibility in the next release. https://github.com/Flank/flank/issues/792

https://firebase-community.slack.com/archives/C1MTSQ5QT/p1589242467494100?thread_ts=1589239891.491600

jan-gogo:
> Should we also omit missing test apk when using additiona-app-test-apks? I think Fladle wanted to specify all the app/test apks via additional-app-test-apks.
> 
> The PR title and release note only mention the app apk. Maybe the test apk is already optional?

Yes, test apk was already optional because robo tests requires only app apk

test_runner/src/main/kotlin/ftl/run/platform/android/ResolveApks.kt
```diff
@@ -3,16 +3,18 @@ package ftl.run.platform.android
...
+            app = appApk,
+            test = testApk,
+            additionalApks = additionalApks
+        )
+    }
 ).plus(
     elements = additionalAppTestApks.map {
         ResolvedApks(
-            app = it.app ?: appApk,
+            app = it.app ?: appApk!!,
```
pawelpasterz, line 17: I think should throw here custom exception, not nasty NPE (which should not happen in kotlin app)

test_runner/src/test/kotlin/ftl/run/platform/android/ResolveApksKtTest.kt
```diff
@@ -0,0 +1,68 @@
...
+                    AppTestPair(
+                        app = "app",
+                        test = "test"
+                    )
+                )
+            }.resolveApks().toTypedArray()
+        )
+    }
+
+    @Test(expected = Exception::class)
```
pawelpasterz, line 54: In addition to previous comment about NPE. We should check here for this particular custom exception.

test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
```diff
@@ -104,13 +104,17 @@ class AndroidArgs(
     override val hasMultipleExecutions: Boolean get() = super.hasMultipleExecutions || additionalAppTestApks.isNotEmpty()
 
     init {
+        if (appApk == null) additionalAppTestApks.filter { (app, _) -> app == null }.let { missingApks ->
+            if (missingApks.isNotEmpty()) throw FlankFatalError("Cannot resolve app apk pair for ${missingApks.map { it.test }}")
```
pawelpasterz, line 108: I am wondering if it wasn't be better to print just apk file name not full path to file. For many (potential) incorrect pairs it would be hard to read something.

bootstraponline: I agree a file name is easier to read. I think the full path will be useful when debugging problems in CI where the paths aren't obvious though. Maybe we could print each missing path on a new line to improve readability?

test_runner/src/main/kotlin/ftl/run/platform/android/ResolveApks.kt
bootstraponline: agreed, I don't like force unwraps in the production code.

test_runner/src/main/kotlin/ftl/cli/firebase/test/android/AndroidRunCommand.kt
```diff
@@ -243,7 +243,7 @@ class AndroidRunCommand : CommonRunCommand(), Runnable {
     // AndroidFlankYml
 
     @Option(
-        names = ["--additional-app-test-apks"],
+        names = ["--app-test-apks", "--additional-app-test-apks"],
```
bootstraponline, line 246: we should probably keep the flags 1:1 mapping with the yaml. `--app-test-apks` isn't a valid key name in the yaml.

PR #825, opened 2020-05-29
#824 Automatically convert -1 in maximum-test-shards to the maximum shard amount

### LINKED ISSUE #824: Fix max-test-shards: -1 to use the max allowed shards
**Describe the bug**
Test lab no longer accepts -1 as the max-shard number 
**To Reproduce**
1. Set    max-test-shards: -1 in flank.yml
2. Error: Caused by: com.google.api.client.googleapis.json.GoogleJsonResponseException: 400 Bad Request
 {
   "code" : 400,
   "errors" : [ {
     "domain" : "global",
     "message" : "num_shards must be >= 1 and <= 50.",
     "reason" : "badRequest"
   } ],
   "message" : "num_shards must be >= 1 and <= 50.",
   "status" : "INVALID_ARGUMENT"
 }

**Expected behavior**
GIve a flank error or remove the readme reference to -1

 Invoke tests on the builds in FTL
 version: v20.05.2
 revision: 7618ebdb5a74db4b448de0b93357221de3ab10c2
 AndroidArgs
     gcloud:
       results-bucket: cloud-test-sofi-test
       results-dir: null
       record-video: true
       timeout: 15m
       async: false
       client-details: 
       network-profile: null
       results-history-name: null
       # Android gcloud
       app: /builds/sofiinc/mobile/wealth-android/app/build/outputs/apk/debug/app-debug.apk
       test: /builds/sofiinc/mobile/wealth-android/app/build/outputs/apk/androidTest/debug/app-debug-androidTest.apk
       additional-apks: 
       auto-google-login: false
       use-orchestrator: false
       directories-to-pull:
       other-files:
       performance-metrics: false
       num-uniform-shards: null
       test-runner-class: null
       test-targets:
         - package com.sofi.mobile.tests.money
         -  notAnnotation com.sofi.mobile.Quarantine
       robo-directives:
       robo-script: null
       device:
         - model: Nexus5X
           version: 26
           locale: en
           orientation: portrait
       num-flaky-test-attempts: 1
     flank:
       max-test-shards: -1
       shard-time: -1
       num-test-runs: 1
       smart-flank-gcs-path: 
       smart-flank-disable-upload: false
       files-to-download:
       test-targets-always-run:
       disable-sharding: false
       project: sofi-test
       local-result-dir: ../app/build/test-results/money-tests/
       # Android Flank Yml
       keep-file-path: false
       additional-app-test-apks:
       run-timeout: -1
       legacy-junit-result: false
       ignore-failed-tests: false

**Additional context**
`set to -1 to use one shard per test. default: 1` from readme

adamfilipow92:
Fixes #824

## Checklist

- [X] Unit tested

### DIFF
--- a/test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
+++ b/test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
@@ -66,7 +66,7 @@ object ArgsHelper {
         )
 
         if (args.maxTestShards !in AVAILABLE_SHARD_COUNT_RANGE && args.maxTestShards != -1)
-            throw FlankFatalError("max-test-shards must be >= ${AVAILABLE_SHARD_COUNT_RANGE.first} and <= ${AVAILABLE_SHARD_COUNT_RANGE.last}, or -1. But current is ${args.maxTestShards}")
+            throw FlankFatalError("max-test-shards must be >= ${AVAILABLE_SHARD_COUNT_RANGE.first} and <= ${AVAILABLE_SHARD_COUNT_RANGE.last}. But current is ${args.maxTestShards}")
 
         if (args.shardTime <= 0 && args.shardTime != -1) throw FlankFatalError("shard-time must be >= 1 or -1")
         if (args.repeatTests < 1) throw FlankFatalError("num-test-runs must be >= 1")
--- a/test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
+++ b/test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
@@ -10,6 +10,7 @@ import com.google.cloud.storage.BucketInfo
 import com.google.cloud.storage.Storage
 import com.google.cloud.storage.StorageClass
 import com.google.cloud.storage.StorageOptions
+import ftl.args.IArgs.Companion.AVAILABLE_SHARD_COUNT_RANGE
 import ftl.args.yml.IYmlMap
 import ftl.args.yml.YamlObjectMapper
 import ftl.config.FtlConstants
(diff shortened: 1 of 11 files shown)


test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
```diff
@@ -64,8 +65,8 @@ object ArgsHelper {
                     " See https://github.com/GoogleCloudPlatform/google-cloud-java#specifying-a-project-id"
         )
 
-        if (args.maxTestShards !in IArgs.AVAILABLE_SHARD_COUNT_RANGE && args.maxTestShards != -1)
-            throw FlankFatalError("max-test-shards must be >= 1 and <= 50, or -1. But current is ${args.maxTestShards}")
+        if (args.maxTestShards !in AVAILABLE_SHARD_COUNT_RANGE && args.maxTestShards != -1)
+            throw FlankFatalError("max-test-shards must be >= ${AVAILABLE_SHARD_COUNT_RANGE.first} and <= ${AVAILABLE_SHARD_COUNT_RANGE.last}, or -1. But current is ${args.maxTestShards}")
```
pawelpasterz, line 69: We should also update message and remove `or -1` since it's not supported.

test_runner/src/main/kotlin/ftl/args/IArgs.kt
```diff
@@ -51,6 +51,12 @@ interface IArgs {
 
     fun useLocalResultDir() = localResultDir != FlankYmlParams.defaultLocalResultsDir
 
+    fun fixMaxTestShardsValue(inputValue: Int): Int = if (inputValue == -1) {
```
pawelpasterz, line 54: I am not convinced with method name, it indicates some problems and does not tell what is wrong. First that came up to my mind is `checkIfInRangeOrDefault` but I am sure there is a better name

jan-gogo: Or maybe `convertToShardCount`?

bootstraponline:
> @bootstraponline I don't know if this is desired behavior but I think we should either 'fix' all incorrect input values or throw an error for all.
> With this implementation we treat `-1` with special care but how about `0`, `< -1` and `> 50`.
> We could change input to the closest acceptable value and maybe print message to inform use that incorrect input was provided and how flank process it...

- `-1` is a special value that means use the maximum amount of shards. This is an important feature for backwards compatibility and also forwards compatibility when FTL increases the maximum shard count.
- `0` is technically a valid shard amount (equal to disabling sharding). I'm fine if this throws an exception, no strong opinions here.
- `< -1` is invalid and should throw an exception
- `> 50` is invalid and should throw an exception

pawelpasterz:
I understand, thanks for response!

test_runner/src/main/kotlin/ftl/args/IArgs.kt
adamfilipow92: Thanks, changed 😄

PR #830, opened 2020-06-01
Add dependency check plugin

pawelpasterz:
Add simple dependency check plugin [gradle-versions-plugin](https://github.com/ben-manes/gradle-versions-plugin)
It adds `gradle dependencyUpdates` task which produces dependencies version report. With given configuration, only latest stable versions are proposed as update candidate

Screenshot 2020-06-01 at 19 22 21


bootstraponline:
This is awesome. I love updating dependencies. 🙂

PR #831, opened 2020-06-02
Refactor config entities and arguments

jan-gogo:
This is a part of #718 

Goals:
1. Unification of yaml entities and cli options as both do the same there was no reason to keep them separately. 
2. Reduction of redundant code between AndroidArgs and IosArgs.
3. Convertion of AndroidArgs and IosArgs to data classes. 
4. Extraction of factory methods and validators.

This refactor of cli options, yaml config, AndroidArgs and IosArgs. It doesn't bring any features or changes so the existing unit tests should cover possible regression.

## Checklist

- [x] Documented
- [x] Unit tested


test_runner/src/main/kotlin/ftl/args/ValidateCommonArgs.kt
```diff
@@ -0,0 +1,52 @@
...
+
+private fun CommonArgs.assertSmartFlankGcsPath() {
+    if (smartFlankGcsPath.isNotEmpty()) {
+        if (!smartFlankGcsPath.startsWith(FtlConstants.GCS_PREFIX)) {
+            throw FlankFatalError("smart-flank-gcs-path must start with gs://")
+        }
+        if (smartFlankGcsPath.count { it == '/' } <= 2 || !smartFlankGcsPath.endsWith(".xml")) {
+            throw FlankFatalError("smart-flank-gcs-path must be in the format gs://bucket/foo.xml")
+        }
+    }
```
piotradamczyk5, line 51: Personally I prefer when instead of nested ifs, but I am not sure if this case it will be suitable, because no-op when empty

test_runner/src/main/kotlin/ftl/cli/firebase/test/CommonRunCommand.kt
```diff
@@ -4,205 +4,22 @@ import picocli.CommandLine
...
 
-    @CommandLine.Option(
-        names = ["--output-style"],
-        description = ["Output style of execution status. May be one of [verbose, multi, single]. " +
-                "For runs with only one test execution the default value is 'verbose', in other cases " +
-                "'multi' is used as the default. The output style 'multi' is not displayed correctly on consoles " +
-                "which don't support ansi codes, to avoid corrupted output use `single` or `verbose`."]
-    )
-    var outputStyle: String? = null
+    // Flank specific
```
piotradamczyk5, line 19: empty line not needed :)

test_runner/src/main/kotlin/ftl/config/common/CommonGcloudConfig.kt
```diff
@@ -0,0 +1,147 @@
...
+            "record-video",
+            "timeout",
+            "async",
+            "results-history-name",
+            "num-flaky-test-attempts"
+        )
+
+        override val map = mapOf(
+            "gcloud" to keys
+        )
```
piotradamczyk5, line 131: could be single line

test_runner/src/main/kotlin/ftl/config/android/AndroidGcloudConfig.kt
```diff
@@ -0,0 +1,263 @@
...
+        description = ["A list of DIMENSION=VALUE pairs which specify a target " +
+                "device to test against. This flag may be repeated to specify multiple devices. The four device dimensions are: " +
+                "model, version, locale, and orientation. If any dimensions are omitted, they will use a default value. Omitting " +
+                "all of the preceding dimension-related flags will run tests against a single device using defaults for all four " +
+                "device dimensions."]
+    )
+    fun deviceMap(map: Map<String, String>?) {
+        if (map.isNullOrEmpty()) return
+        val androidDevice = Device(
+            model = map.getOrDefault("model", FtlConstants.defaultAndroidModel),
```
adamfilipow92, line 207: We use "model","version" etc in IosArgs and Android args. Maybe we should move keys to const/enums?

test_runner/src/main/kotlin/ftl/args/CreateIosArgs.kt
```diff
@@ -0,0 +1,19 @@
...
+
+import ftl.config.IosConfig
+import ftl.config.ios.IosFlankConfig
+import ftl.config.ios.IosGcloudConfig
+
+fun createIosArgs(
+    config: IosConfig? = null,
+    gcloud: IosGcloudConfig = config!!.platform.gcloud,
+    flank: IosFlankConfig = config!!.platform.flank,
+    commonArgs: CommonArgs = config!!.common.createCommonArgs(config.data)
```
piotradamczyk5, line 11: this is a little bit unclear for me, `config` is nullable, but later on we declare that it is not null. So probably IosConfig is not nullable type.

test_runner/src/main/kotlin/ftl/args/IosArgs.kt
```diff
@@ -159,47 +58,22 @@ IosArgs
...
+    listOf(emptyList()) else
+    ArgsHelper.calculateShards(
+        filterTests(
+            Xctestrun.findTestNames(xctestrunFile),
+            testTargets
+        ).distinct().map {
+            FlankTestMethod(it, ignored = false)
+        },
+        this
+    )
```
piotradamczyk5, line 73: I would format it
```
private fun IosArgs.calculateShardChunks() =
    if (disableSharding) listOf(emptyList())
    else ArgsHelper.calculateShards(
        filteredTests = filterTests(Xctestrun.findTestNames(xctestrunFile), testTargets)
            .distinct()
            .map { FlankTestMethod(it, ignored = false) },
        args = this
    )
```

test_runner/src/main/kotlin/ftl/config/common/CommonGcloudConfig.kt
jan-gogo: Rights, I was trying to keep one convention for maps and keys in all configs. Anyway this probably shouldn't be map, rather simple pair, same about another configs.

test_runner/src/main/kotlin/ftl/config/android/AndroidGcloudConfig.kt
jan-gogo: yeah, why not

test_runner/src/main/kotlin/ftl/args/CreateIosArgs.kt
jan-gogo: Right, it doesn't looks well and should be explained, `config` is nullable cause when `gcloud`, `flank`, `commonArgs` are set explicit the `config` is redundant from logical perspective. Possible solution to get rid of null is:
```
fun createIosArgs(
    config: IosConfig
) = createIosArgs(
    gcloud = config.platform.gcloud,
    flank = config.platform.flank,
    commonArgs = config.common.createCommonArgs(config.data)
)

private fun createIosArgs(
    gcloud: IosGcloudConfig,
    flank: IosFlankConfig,
    commonArgs: CommonArgs
) = IosArgs(
    commonArgs = commonArgs,
    xctestrunZip = gcloud.test!!.processFilePath("from test"),
    xctestrunFile = gcloud.xctestrunFile!!.processFilePath("from xctestrun-file"),
    xcodeVersion = gcloud.xcodeVersion,
    devices = gcloud.device!!,
    testTargets = flank.testTargets!!.filterNotNull()
)
```
But it gives 2 functions instead of 1 and more lines. I'm am not sure which is better.

test_runner/src/main/kotlin/ftl/args/ValidateCommonArgs.kt
jan-gogo: TBH me to. I will change this.

jan-gogo:
```
private fun CommonArgs.assertSmartFlankGcsPath(): Unit = with(smartFlankGcsPath) {
    when {
        isEmpty() -> Unit

        startsWith(FtlConstants.GCS_PREFIX).not() -> throw FlankFatalError(
            "smart-flank-gcs-path must start with gs://"
        )

        count { it == '/' } <= 2 || endsWith(".xml").not() -> throw FlankFatalError(
            "smart-flank-gcs-path must be in the format gs://bucket/foo.xml"
        )
    }
}
```

test_runner/src/main/kotlin/ftl/config/android/AndroidGcloudConfig.kt
jan-gogo: Heh, if both android and ios gcloud config has common `device` option so probably it should be in CommonGcloudConfig.

test_runner/src/main/kotlin/ftl/args/ValidateCommonArgs.kt
```diff
@@ -0,0 +1,55 @@
...
+                " See https://github.com/GoogleCloudPlatform/google-cloud-java#specifying-a-project-id"
+    )
+}
+
+private fun CommonArgs.assertMaxTestShards() {
+    if (
+        maxTestShards !in IArgs.AVAILABLE_SHARD_COUNT_RANGE &&
+        maxTestShards != -1
+    ) throw FlankFatalError(
+        "max-test-shards must be >= 1 and <= 50, or -1. But current is $maxTestShards"
```
adamfilipow92, line 27: I think we should use IArgs.AVAILABLE_SHARD_COUNT_RANGE.first / IArgs.AVAILABLE_SHARD_COUNT_RANGE.last in exception message.

test_runner/src/main/kotlin/ftl/config/Device.kt
```diff
@@ -19,3 +23,19 @@ data class Device(
           orientation: $orientation""".trimStartLine()
     }
```
adamfilipow92, line 24: I know you don't change this line but:
```
override fun toString(): String {
        return """
        - model: $model
          version: $version
          locale: $locale
          orientation: $orientation""".trimStartLine()
    }
```
vs

```
override fun toString() = """
        - model: $model
          version: $version
          locale: $locale
          orientation: $orientation""".trimStartLine()
```
This only for discussion to keep one style. Personally I prefer second.

PR #855, opened 2020-06-16
Null no tests found

### LINKED ISSUE #818: Shards failed - no test cases inside
> Mr.uno 
>
> One or more shards failed because there are no test cases inside. Please check your sharding configuration. I've encountered this a couple times with our submodule tests which use a lot of additional test apks. Maybe some error somewhere in that logic - I've only ever seen it with the submodule tests. This is from the latest Flank version.

image

https://firebase-community.slack.com/archives/C72V6UW8M/p1590537597298100

jan-gogo:
Fixes #818 

We don't know what is the reason and how to reproduce it in same way as reported #818. 
We need to collect all information as we can about this issue.
This issue has dedicated [doc](https://github.com/Flank/flank/blob/818-null-no-tests-found/docs/bugs/null_no_tests_found.md).

estimation - dragon

## Checklist

- [ ] Documented
- [ ] Unit tested

### DIFF
--- a/docs/bugs/null_no_tests_found.md
+++ b/docs/bugs/null_no_tests_found.md
@@ -1,0 +1,114 @@
+# The problem
+One or more shards failed because there are no test cases inside. This problem encounter a couple times on submodule tests which use a lot of additional test apks. It was noticed only with the submodule tests.
+
+## Stack trace
+```
+java.lang.ClassNotFoundException: Invalid name: no tests found
+at java.lang.Class.classForName(Native Method)
+at java.lang.Class.forName(Class.java:324)
+at androidx.test.internal.runner.TestLoader.doCreateRunner(TestLoader.java:72)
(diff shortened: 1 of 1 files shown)


docs/bugs/null_no_tests_found.md
```diff
@@ -0,0 +1,114 @@
+# The problem
+One or more shards failed because there are no test cases inside. This problem encounter a couple times on submodule tests which use a lot of additional test apks. It was noticed only with the submodule tests.
```
bootstraponline, line 2: submodule means something different: https://git-scm.com/book/en/v2/Git-Tools-Submodules

I think you mean multi-module apps.

jan-gogo: Right, that is slightly redacted copy paste from #818 description. I will fix.

PR #865, opened 2020-06-23
Flank needs to respect the timeout value as that's a cap for billing purposes.

### LINKED ISSUE #836: Fix billable minutes
> pauldavis 
>
> Flank is over-reporting billable minutes for tests that hit their timeout limit.  FTL caps the billed time at the timeout value, but Flank seems to report total time (e.g. a Robo test has a 1 minute timeout, but actually runs 75sec before the cancellation happens. FTL bills for 1m, but Flank says 2m.

See https://firebase-community.slack.com/archives/C1MTSQ5QT/p1591218666237500

adamfilipow92:
Fixes #836 

#### Flank billing calculation rules:

1. test execution time is rounded up to next minute and used in calculations

| Execution time  | Rounded value |
| ------------- | ------------- |
| `3m 22s` | `4m` |
| `15m 01s`  | `16m`  |
| `5m 00s` | `5m` |
2. if test execution time is smaller then test timeout (default 15m) round up execution time and use it in billing calculations
3. if test execution time is bigger then test timeout use timeout value in billing calculations
4. if test execution state is `ERROR` flank should skip billing calculation and assume `$0.00`

Examples:

1. execution time is ` 13m 24s` and timeout is default (`15m`), execution time is smaller then timeout therefor is rounded up to `14m` and used in billing calculations
2. execution time is `20m 13s` and timeout is set to `18m`, execution time is bigger then timeout therefor timeout value (`18m`) is used in billing calculations

Flank needs to respect the timeout value as that's a cap for billing purposes, and round time to nearest minutes. 

1. Run flank with yml

```

gcloud:
  robo-script: ./src/test/kotlin/ftl/fixtures/test_app_cases/MainActivity_robo_script.json
  environment-variables:
    coverage: true
    coverageFilePath: /sdcard/
    clearPackageData: true

```
flank should return cost report with 2 min

```

CostReport
  Virtual devices
    $0.03 for 2m

```

2. Add in gcloud section timeout for 60s

```

gcloud:
  robo-script: ./src/test/kotlin/ftl/fixtures/test_app_cases/MainActivity_robo_script.json
  environment-variables:
    coverage: true
    coverageFilePath: /sdcard/
    clearPackageData: true
  timeout: 60s

```

flank should return cost report with 1 min
```

CostReport
  Virtual devices
    $0.02 for 1m

```

3. To check timeout is respected only when is lower than execution time set timeout to 5m and run

```

gcloud:
  robo-script: ./src/test/kotlin/ftl/fixtures/test_app_cases/MainActivity_robo_script.json
  environment-variables:
    coverage: true
    coverageFilePath: /sdcard/
    clearPackageData: true
  timeout: 5m

```

flank should return cost report with 2 min

```

CostReport
  Virtual devices
    $0.03 for 2m

```

## Checklist

- [X] Documented
- [X] Unit tested


bootstraponline:
> Cost report should contains seconds not rounded up minutes.

Hey, this is not a correct description of the problem. FTL does round to the nearest minute and that's how they bill.

The problem is that Flank needs to respect the timeout value as that's a cap for billing purposes. 

> Flank is over-reporting billable minutes for tests that hit their timeout limit. FTL caps the billed time at the timeout value, but Flank seems to report total time (e.g. a Robo test has a 1 minute timeout, but actually runs 75sec before the cancellation happens. FTL bills for 1m, but Flank says 2m.

adamfilipow92:
> > Cost report should contains seconds not rounded up minutes.
> 
> Hey, this is not a correct description of the problem. FTL does round to the nearest minute and that's how they bill.
> 
> The problem is that Flank needs to respect the timeout value as that's a cap for billing purposes.
> 
> > Flank is over-reporting billable minutes for tests that hit their timeout limit. FTL caps the billed time at the timeout value, but Flank seems to report total time (e.g. a Robo test has a 1 minute timeout, but actually runs 75sec before the cancellation happens. FTL bills for 1m, but Flank says 2m.

Thanks for fast feedback! I check rounding time strategy and flank rounding time always up so if we have 1m 7s flank round it to 2 mins. I think this is problem too because flank should round 1m 7s to 1m right?

bootstraponline:
> Thanks for fast feedback! I check rounding time strategy and flank rounding time always up. Now if we have 1m 7s flank round it to 2 mins. I think this is problem too because flank should round 1m 7s to 1m right?

Firebase will bill us for 2m if a shard runs for 1m 7s. The only exception is if there's a timeout for 1m, then we'll only ever be billed up to the timeout.

pawelpasterz:
> > Thanks for fast feedback! I check rounding time strategy and flank rounding time always up. Now if we have 1m 7s flank round it to 2 mins. I think this is problem too because flank should round 1m 7s to 1m right?
> 
> Firebase will bill us for 2m if a shard runs for 1m 7s. The only exception is if there's a timeout for 1m, then we'll only ever be billed up to the timeout.

agreed, flank needs to take timeout value under consideration.
This implementation will be improved :)

pawelpasterz:
Description updated

bootstraponline:
Feedback from Paul Davis via Slack:

> That looks correct, as long as you also report zero billing for all infrastructure errors. Unfortunately, infra errors can be marked on either the last attempt within an execution, or on the execution itself.
> In FTL billing logic, that looks like

```java
if (!execution.attempts().isEmpty()) {
    Attempt lastAttempt = execution.attempts().get(execution.attempts().size() - 1);
    if (lastAttempt.getAttemptStatus() != AttemptStatus.ERROR && execution.state() != TestState.ERROR) {
        // do rounding/timeout calculations as described in the pull request
    }
}
```

pawelpasterz:
> Feedback from Paul Davis via Slack:
> 
> > That looks correct, as long as you also report zero billing for all infrastructure errors. Unfortunately, infra errors can be marked on either the last attempt within an execution, or on the execution itself.
> > In FTL billing logic, that looks like
> 
> ```java
> if (!execution.attempts().isEmpty()) {
>     Attempt lastAttempt = execution.attempts().get(execution.attempts().size() - 1);
>     if (lastAttempt.getAttemptStatus() != AttemptStatus.ERROR && execution.state() != TestState.ERROR) {
>         // do rounding/timeout calculations as described in the pull request
>     }
> }
> ```

@bootstraponline 
That's valuable information, thanks! (I was not able to find such info in FTL docs)

bootstraponline:
> I was not able to find such info in FTL docs

we should probably document clearly in the Kotlin file what the business rules for billing are. 🙂

test_runner/src/main/kotlin/ftl/json/SavedMatrix.kt
```diff
@@ -27,7 +29,7 @@ class SavedMatrix(matrix: TestMatrix) {
 
     var billableVirtualMinutes: Long = 0
         private set
-    var billablePhysicalMinutes: Long = 0
+    var billablePhysicalSecondsMinutes: Long = 0
```
piotradamczyk5, line 32: could we think about other name, because I am confused if we use seconds or minutes

test_runner/src/test/kotlin/ftl/json/MatrixMapTest.kt
```diff
@@ -25,15 +26,16 @@ class MatrixMapTest {
...
                 .setResultStorage(createResultsStorage())
                 .setState(MatrixState.FINISHED)
                 .setTestMatrixId("123")
                 .setTestExecutions(
                     listOf(
                         createStepExecution(executionId, "")
                     )
-                )
+                ),
+            testTimeout = defaultTestTimeout
```
piotradamczyk5, line 38: maybe make it as a default parameter value

pawelpasterz: If you are thinking about default value for `SavedMatrix` constructor I don't think it's good idea. That would lead us to 2 places where default value (common args and `SavedMatrix` constructor) are hold, easy to forget about one when changing/updating second

test_runner/src/main/kotlin/ftl/json/SavedMatrix.kt
pawelpasterz: Yeah, definitely. We have yet another round of refactor ahead of us

pawelpasterz:
According to Paul Davis, we can check just `TestExecution` for state and if it's `ERROR` skip billing
> Just check the TestExecution state, and report $0.00 if the state is ERROR.
Description updated

cc @bootstraponline @adamfilipow92

bootstraponline:
> In theory we could use `deviceUsageDuration` to exact billing calculation and assume `$0.00` in case when it's `null` -- this needs to be verified

I'd like to support both Spark plan and the paid plan. It's useful to know what Flank runs will cost.

We probably need to setup a paid FTL account to verify `deviceUsageDuration` works as expected.

pawelpasterz:
> > In theory we could use `deviceUsageDuration` to exact billing calculation and assume `$0.00` in case when it's `null` -- this needs to be verified
> 
> I'd like to support both Spark plan and the paid plan. It's useful to know what Flank runs will cost.

Therefor `deviceUsageDuration` would be best for this.

> We probably need to setup a paid FTL account to verify deviceUsageDuration works as expected.

I might have access to project with paid plan -- will check it

test_runner/src/main/kotlin/ftl/json/SavedMatrix.kt
adamfilipow92: Thanks, its my fail. Its should be billablePhysicalMinutes

test_runner/src/test/kotlin/ftl/json/MatrixMapTest.kt
adamfilipow92: Now we get testTimeout from api response so we don't need pass them by constructor.

pawelpasterz:
@bootstraponline 
I've tested fetching value from `deviceUsageDuration` on project with `Blaze` without any result -- `null` is all I got. I think for now we should go with 'error check' approach and come back to this once API would be more...reliable? 

With 'error check' we would have correct numbers as well

cc @adamfilipow92

adamfilipow92:
So this is plan:
1. Skip billing if testExecution.state == ERROR.
2. Calculate cost even if we on spark plan.

bootstraponline:
> I've tested fetching value from `deviceUsageDuration` on project with `Blaze` without any result -- `null` is all I got. I think for now we should go with 'error check' approach and come back to this once API would be more...reliable?
> 
> With 'error check' we would have correct numbers as well

That sounds good. The FTL API is often completely broken so it's good that we always double check. 🙂

PR #876, opened 2020-07-02
#870 Added --directories-to-pull validation and avoid making request with empty toolStepResult

### LINKED ISSUE #870: IllegalStateException: toolResultsStep must not be null
I'm using Fladle 0.10.1 which depends on Flank 20.05.2.

When I add the following to my fladle block:
```
fladle {
  ...
  directoriesToPull = [
      "/data/data/com.squareup.cash.beta.debug/test-attachments"
  ]
```

I get the following error:
```
RunTests

  Smart Flank cache hit: 0% (0 / 3)
  Shard times: 360s

  Uploading app-internal-debug-androidTest.apk .  Uploading app-internal-debug.apk .
  ...

FetchArtifacts
  
java.lang.IllegalStateException: toolResultsStep must not be null
	at ftl.reports.api.CreateTestExecutionDataKt.createTestExecutionData(CreateTestExecutionData.kt:31)
	at ftl.reports.api.CreateTestExecutionDataKt$createTestExecutionDataListAsync$1$1$1.invokeSuspend(CreateTestExecutionData.kt:22)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(Dispatched.kt:241)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:594)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.access$runSafely(CoroutineScheduler.kt:60)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:740)
```

piotradamczyk5:
Fixes #870 

NPE with missing `toolStepResult` was caused because we try to get results for an invalid matrix. Matrix was invalid because FTL rejects the request with invalid `--directories-to-pull` value.
To avoid this issue correct validation of this parameter was introduced, as well as filtering getting results for empty `toolStepResult`.
[Gcloud](https://cloud.google.com/sdk/gcloud/reference/firebase/test/android/run#--directories-to-pull) policies were applied to `--directories-to-pull`:

- the path must start with `/sdcard` or `/data/local/tmp`
- the path must contain only characters `a-zA-Z0-9_-./+`

In order to test this issue, please use `--directories-to-pull` value which does not meet this 2 criteria.

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/test_runner/src/main/kotlin/ftl/reports/api/CreateJUnitTestResult.kt
+++ b/test_runner/src/main/kotlin/ftl/reports/api/CreateJUnitTestResult.kt
@@ -7,6 +7,7 @@ internal fun List<TestExecution>.createJUnitTestResult(
     withStackTraces: Boolean = false
 ) = JUnitTestResult(
     testsuites = this
+        .filterNullToolResultsStep()
         .createTestExecutionDataListAsync()
         .prepareForJUnitResult()
         .let { executionDataList ->
@@ -16,3 +17,5 @@ internal fun List<TestExecution>.createJUnitTestResult(
         .createJUnitTestSuites()
         .toMutableList()
 )
+
+private fun List<TestExecution>.filterNullToolResultsStep() = filter { it.toolResultsStep != null }
--- a/test_runner/src/main/kotlin/ftl/util/Utils.kt
+++ b/test_runner/src/main/kotlin/ftl/util/Utils.kt
@@ -169,8 +169,7 @@ fun withGlobalExceptionHandling(block: () -> Int) {
 }
 
 private fun SavedMatrix.logError() {
-    println("More details are available at [${this.webLink}]")
-    println(this.asPrintableTable())
+    println("Matrix is $state")
 }
 
 fun <R : MutableMap<String, Any>, T> mutableMapProperty(
--- a/test_runner/src/test/kotlin/ftl/util/UtilsTest.kt
+++ b/test_runner/src/test/kotlin/ftl/util/UtilsTest.kt
@@ -197,8 +197,7 @@ class UtilsTest {
         // will
         exit.expectSystemExitWithStatus(3)
         exit.checkAssertionAfterwards {
-            assertTrue(output.log.contains("More details are available at [${testMatrix1.webLink}]"))
-            assertOutputIsEqualReferenceTable(output.log, testMatrix1.matrixId)
+            assertTrue(output.log.contains("Matrix is ${testMatrix1.state}"))
         }
 
         // when
@@ -274,20 +273,6 @@ class UtilsTest {
 
         // when
         withGlobalExceptionHandling(block)
-    }
-
-    private fun assertOutputIsEqualReferenceTable(outputLog: String, matrixId: String) {
-        val referenceTable =
-            "┌─────────┬───────────┬────────────────────┐\n" +
-            "│ OUTCOME │ MATRIX ID │    TEST DETAILS    │\n" +
-            "├─────────┼───────────┼────────────────────┤\n" +
-            "│ Failed  │     $matrixId     │ Test failed to run │\n" +
-            "└─────────┴───────────┴────────────────────┘"
-        referenceTable.split("\n")
-            .forEach {
-                println(it)
-                assertTrue(outputLog.contains(it))
-            }
     }
 
     @Test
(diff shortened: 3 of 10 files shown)


adamfilipow92:
I noticed some differnces on master and your branche.
If you run multiple success on master its working and not crashing with npe
Check yaml config
```
gcloud:
  test: ./src/test/kotlin/ftl/fixtures/tmp/apk/app-multiple-success-debug-androidTest.apk
  environment-variables: 
    coverage: true
    coverageFilePath: /sdcard/
    clearPackageData: true
  directories-to-pull: 
    - /sdcard/
  use-orchestrator: false

flank:
  disable-sharding: false
  max-test-shards: 2
  files-to-download:
    - .*/sdcard/[^/]+\.ec$


```

With your changes its fast fail. I think the problem is when last char is ``` / ```

I think if ``` /sdcard/ ``` not crashing its should't fast fail. 

Tell me what you think.

test_runner/src/main/kotlin/ftl/reports/api/CreateJUnitTestResult.kt
```diff
@@ -7,6 +7,7 @@ internal fun List<TestExecution>.createJUnitTestResult(
     withStackTraces: Boolean = false
 ) = JUnitTestResult(
     testsuites = this
+        .filterNullToolResultsStep()
```
jan-gogo, line 10: 1. I wonder why FTL is returning `null` for `toolResultsStep`? 
2. Can we reproduce this error? 
3. Its safe to skip some tests execution and get correct results?
4. Maybe flank should display proper warning about some unexpected behaviour?

piotradamczyk5: AD 1. If matrix is invalid FTL returns `null` as `toolResultsStep`
AD 2. This is more to avoid crashing flank in the future, so far, for now, I do not know how to reproduce this issue
AD 3. This is safe because tests results are based on a request made with `toolResultsStep` which will not succeed without `toolResultsStep` 
AD 4. Ideally, we should not trigger INVALID matrix request, but for now, we do not know which settings will cause matrix invalid (see point 2)

> 1. I wonder why FTL is returning `null` for `toolResultsStep`?
> 2. Can we reproduce this error?
> 3. Its safe to skip some tests execution and get correct results?
> 4. Maybe flank should display proper warning about some unexpected behaviour?

AD 1. If matrix is invalid FTL returns `null` as `toolResultsStep`
AD 2. This is more to avoid crashing flank in the future, so far, for now, I do not know how to reproduce this issue
AD 3. This is safe because tests results are based on a request made with `toolResultsStep` which will not succeed without `toolResultsStep` 
AD 4. Ideally, we should not trigger INVALID matrix request, but for now, we do not know which settings will cause matrix invalid (see point 2)

test_runner/src/test/kotlin/ftl/util/UtilsTest.kt
```diff
@@ -197,8 +197,7 @@ class UtilsTest {
         // will
         exit.expectSystemExitWithStatus(3)
         exit.checkAssertionAfterwards {
-            assertTrue(output.log.contains("More details are available at [${testMatrix1.webLink}]"))
-            assertOutputIsEqualReferenceTable(output.log, testMatrix1.matrixId)
+            assertTrue(output.log.contains("Matrix is ${testMatrix1.state}"))
```
jan-gogo, line 200: There will be no table output if flank throw `FTLError` before `MatrixResultsReport`?

test_runner/src/main/kotlin/ftl/util/Utils.kt
```diff
@@ -169,8 +169,7 @@ fun withGlobalExceptionHandling(block: () -> Int) {
 }
 
 private fun SavedMatrix.logError() {
-    println("More details are available at [${this.webLink}]")
-    println(this.asPrintableTable())
+    println("Matrix is $state")
```
jan-gogo, line 172: There will be no table output if flank throw FTLError before MatrixResultsReport?

piotradamczyk5: `FTLError` is throw after printing table
```

listOf(
    CostReport,
    MatrixResultsReport
).map {
    it.run(matrices, testSuite, printToStdout = true, args = args)  >>>>>>>>>>>>> PRINTING TABLE
}
if (!matrices.allSuccessful()) {
    listOf(
        HtmlErrorReport
    ).map { it.run(matrices, testSuite, printToStdout = false, args = args) }
}
JUnitReport.run(matrices, testSuite?.apply {
    if (ignoredTestCases.isNotEmpty()) {
        testsuites?.add(ignoredTestCases.toJunitTestsResults())
    }
}, printToStdout = false, args = args)
when {
    args.fullJUnitResult -> processFullJunitResult(args, matrices, testShardChunks)
    args.useLegacyJUnitResult -> processJunitXml(testSuite, args, testShardChunks)
}
matrices.validateMatrices(args.ignoreFailedTests) >>>>>>>>>>> FTLError
```

Issue #872, opened 2020-06-26
Non-deterministic issue when parsing JUnitReport.xml generated by Flank 20.06.2


plnice:
**Describe the bug**

After updating from Flank 8.0.1 to 20.06.2 and switching to server-side sharding sometimes on our CI (TeamCity) we see following error when parsing `JUnitReport.xml`:
```
[17:00:00]Ant JUnit report watcher
[17:00:00][Ant JUnit report watcher] 1 report found for paths:
[17:00:00][Ant JUnit report watcher] results/**/JUnitReport.xml
[17:00:00][Ant JUnit report watcher] Parsing errors
[17:00:00][Parsing errors] Failed to parse 1 report
[17:00:00][Parsing errors] results/2020-06-25_16-47-33.675000_QnCW/JUnitReport.xml: Content is not allowed in prolog.
[17:00:00][Parsing errors] jetbrains.buildServer.util.XmlXppAbstractParser$3: Content is not allowed in prolog.
	at jetbrains.buildServer.util.XmlXppAbstractParser.parse(XmlXppAbstractParser.java:39)
	at jetbrains.buildServer.util.XmlXppAbstractParser.parse(XmlXppAbstractParser.java:31)
	at jetbrains.buildServer.xmlReportPlugin.parsers.antJUnit.AntJUnitReportParser.parse(AntJUnitReportParser.java:179)
	at jetbrains.buildServer.xmlReportPlugin.ParseReportCommand.run(ParseReportCommand.java:62)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: org.xml.sax.SAXParseException; lineNumber: 1; columnNumber: 1; Content is not allowed in prolog.
	at org.apache.xerces.parsers.AbstractSAXParser.parse(Unknown Source)
	at org.apache.xerces.jaxp.SAXParserImpl$JAXPSAXParser.parse(Unknown Source)
	at jetbrains.buildServer.util.XmlXppAbstractParser.parseWithSAX(XmlXppAbstractParser.java:240)
[...]

PR #877, opened 2020-07-06
#872 Added documentation about investigating #872 issue

piotradamczyk5:
Contains investigation results of #872 

This is investigation results of checking #872 issues

## Checklist

- [x] Documented


bootstraponline:
Is the issue possibly the result of running a test using Flank 8 and then running a test with Flank 20 and the format has diverged? Maybe there's some caching of the smart test XML.

piotradamczyk5:
> Is the issue possibly the result of running a test using Flank 8 and then running a test with Flank 20 and the format has diverged? Maybe there's some caching of the smart test XML.

I think this is not that case, because most of time report is parsed properly. 
The exception which occurs on TeamCity CI is about bad characters at the first line. From I could see the code of generating a report is not slight different than in previous versions, so I am considering raising an issue to TeamCity CI. 
@adamfilipow92  has some experience with this CI and he will ask the user about configuration

adamfilipow92:
I asked user about agents count and configuration but all agents are docker containers so it's always the same. So the problem isn't here. User have 2 types of tests 
1. Using MockWebServer and have ~140 tests executed on multiple shards
2. Using real backend, not many tests executed on single shards - this type of tests sometime falling

Between versions user changed config from: 
``` num-uniform-shards: 1 ```
to
``` num-test-runs: 1 ```

But when user back to old configuration, he can still reproduce problem.

PR #890, opened 2020-07-15
#768 Convert bitrise ubuntu workflow into GitHub actions

### LINKED ISSUE #768: Convert Flank | Ubuntu Public to GitHub Actions
Convert the Bitrise `Flank | Ubuntu Public` build to run on GitHub actions.

Use `ubuntu-latest`

https://app.bitrise.io/app/fbdede5ce50cbc21#/builds

piotradamczyk5:
Fixes #768 

The ubuntu-workflow runs in each PR

## Checklist


bootstraponline:
I think these binaries should be auto downloaded? We don't want to add large binaries to the repo if possible.

piotradamczyk5:
> I think these binaries should be auto downloaded? We don't want to add large binaries to the repo if possible.

I think that you are right, the same is for `main/resources`.
I will create the ticket to add these binaries with git annex after setup it

PR #798, opened 2020-05-19
Remove failure nodes from tests that passed on retry

adamfilipow92:
Fixes #796 

## Checklist

- [X] Documented
- [X] Clean up code before mark as ready to review
- [X] Local merge should contains all error nodes
- [X] Save CI Report to local fs
- [X] Upload CI Report to gcs
- [X] Unit tested
- [X] release_notes.md updated

PR #935, opened 2020-08-05
Process junit xml even when full junit is not enabled

### LINKED ISSUE #902: Clearly explain how to enable smart flank for Android in documentation

As an Android developer, I want the documentation to clearly explain how to enable smart flank for Android.

After exploring the source code, I found the config need to fulfill two requirements:
- Set a valid value to `smart-flank-gcs-path`
- Set either the `legacy-junit-result` or `full-junit-result` to be `true`

It would be better if the document could state these two points.😀

kozaxinan:
Fixes #902 

#798 introduced a new argument `full-junit-result`. It is false by default. First problem, it is not added to keys list so flank doctor report new argument as unknown key.

Second problem is about junit xml report for smart flank run. `full-junit-result` and `legacy-junit-result` are used to decide to send junit report. As I understand from #798, there are now there junit result, legacy and full. Consideration; when both flag are true, flank can throw exception at argument checking phase.

While using flank without legacy and full junit test report, smart test sharding and junit xml upload need to work.

## Checklist

- [x] Documented
- [x] Unit tested
- [ ] release_notes.md updated (Waiting for review before updating notes)

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -33,7 +33,7 @@ Run `./gradlew check` to fix lint issues
 
 ## Adding new gcloud property common to iOS and Android
 
-- Add property to `GcloudYml` and update `keys`
+- Add property to `GcloudYml` and update `keys` in config files
 - Update `IArgs` with new property
 - Update `AndroidArgs` to reference the propery and update `toString`
 - Update `IosArgs` to reference the propery and `toString`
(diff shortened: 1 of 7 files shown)


bootstraponline:
Thanks for the pull request! The team will review this soon.

kozaxinan:
I will update the release notes and merge if everything looks good. Next step is preparing fladle for Android uswr for full xml report.

PR #969, opened 2020-08-10
feat: add locales description command for ios and android

### LINKED ISSUE #966: Implement `locales describe` feature

As a Flank user, I want to use locales description feature so I can check what locales parameters are.

Audit all the gcloud CLI commands. Check that:

 - The command exists in Flank
 - The flags are named the same
 - The yaml (if supported) is the same and works both with gcloud and Flank
 - The output from gcloud matches the output from flank

**Additional context**

This issue is part of #959

adamfilipow92:
Fixes #966 

When run 

```

     flank firebase test android|ios locales describe LOCALE_ID

```

flank should return output like gcloud cli when execute

```

gcloud alpha firebase test android|ios locales describe LOCALE_ID


```

example for ``` flank firebase test android locales describe en ```

```

id: en
name: English
tags:
- default

```

example for ``` flank firebase test android locales describe pl ```

```

id: pl
name: Polish

```

example for the wrong LOCALE_ID ``` flank firebase test android locales describe test ```

```

ERROR: 'test' is not a valid locale

```

## Checklist

- [X] Unit tested

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -2,6 +2,7 @@
 - [#987](https://github.com/Flank/flank/pull/987) Flank Error Monitoring readme addition ([sloox](https://github.com/Sloox))
 - [#990](https://github.com/Flank/flank/pull/990) Fix: exclusion of @Suppress test. ([piotradamczyk5](https://github.com/piotradamczyk5))
 - [#988](https://github.com/Flank/flank/pull/988) Add versions description command for ios and android. ([adamfilipow92](https://github.com/adamfilipow92))
+- [#969](https://github.com/Flank/flank/pull/969) Add locales description command for ios and android. ([adamfilipow92](https://github.com/adamfilipow92))
 -
 
 ## v20.08.1
@@ -17,6 +18,7 @@
 - [#970](https://github.com/Flank/flank/pull/970) Fixing Flood of snapshot  releases. ([piotradamczyk5](https://github.com/piotradamczyk5))
 - [#972](https://github.com/Flank/flank/pull/972) Fix printing release version. ([piotradamczyk5](https://github.com/piotradamczyk5))
 - [#964](https://github.com/Flank/flank/pull/964) Implement network-profiles describe feature. ([pawelpasterz](https://github.com/pawelpasterz))
+-
 
 ## v20.08.0
 - [#890](https://github.com/Flank/flank/pull/890) Convert bitrise ubuntu workflow into GitHub actions. ([piotradamczyk5](https://github.com/piotradamczyk5))
--- a/release_notes.md
+++ b/release_notes.md
@@ -18,7 +18,6 @@
 - [#970](https://github.com/Flank/flank/pull/970) Fixing Flood of snapshot  releases. ([piotradamczyk5](https://github.com/piotradamczyk5))
 - [#972](https://github.com/Flank/flank/pull/972) Fix printing release version. ([piotradamczyk5](https://github.com/piotradamczyk5))
 - [#964](https://github.com/Flank/flank/pull/964) Implement network-profiles describe feature. ([pawelpasterz](https://github.com/pawelpasterz))
--
 
 ## v20.08.0
 - [#890](https://github.com/Flank/flank/pull/890) Convert bitrise ubuntu workflow into GitHub actions. ([piotradamczyk5](https://github.com/piotradamczyk5))
--- a/test_runner/src/main/kotlin/ftl/android/AndroidCatalog.kt
+++ b/test_runner/src/main/kotlin/ftl/android/AndroidCatalog.kt
@@ -2,10 +2,12 @@ package ftl.android
 
 import com.google.api.services.testing.model.AndroidDevice
 import com.google.api.services.testing.model.AndroidDeviceCatalog
+import com.google.api.services.testing.model.Locale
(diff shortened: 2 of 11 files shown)


release_notes.md
```diff
@@ -15,6 +15,7 @@
 - [#970](https://github.com/Flank/flank/pull/970) Fixing Flood of snapshot  releases. ([piotradamczyk5](https://github.com/piotradamczyk5))
 - [#972](https://github.com/Flank/flank/pull/972) Fix printing release version. ([piotradamczyk5](https://github.com/piotradamczyk5))
 - [#964](https://github.com/Flank/flank/pull/964) Implement network-profiles describe feature. ([pawelpasterz](https://github.com/pawelpasterz))
+-
```
piotradamczyk5, line 18: not needed :)

test_runner/src/main/kotlin/ftl/environment/LocalesDescription.kt
```diff
@@ -0,0 +1,22 @@
...
+
+private fun String.addTagsIfExists(locale: Locale) =
+    if (locale.tags.isNullOrEmpty().not()) StringBuilder(this).appendln("\ntags:").appendTagsToList(locale)
+    else this
+
+private fun StringBuilder.appendTagsToList(locale: Locale) = apply {
+    locale.tags.filterNotNull().forEach { tag -> appendln("- $tag") }
+}.toString().trim()
+
+private fun String?.orErrorMessage(locale: String) = this ?: "ERROR: '$locale' is not a valid locale".trimIndent()
```
piotradamczyk5, line 22: I think that `trimIndent()` is not needed for this simple string message

test_runner/src/main/kotlin/ftl/android/AndroidCatalog.kt
```diff
@@ -31,7 +32,11 @@ object AndroidCatalog {
 
     fun supportedOrientationsAsTable(projectId: String) = deviceCatalog(projectId).runtimeConfiguration.orientations.asPrintableTable()
 
-    fun localesAsTable(projectId: String) = deviceCatalog(projectId).runtimeConfiguration.locales.asPrintableTable()
+    fun localesAsTable(projectId: String) = getLocales(projectId).asPrintableTable()
+
+    fun getLocaleDescription(projectId: String, locale: String) = getLocales(projectId).getLocaleDescription(locale)
+
+    private fun getLocales(projectId: String) = deviceCatalog(projectId).runtimeConfiguration.locales
```
pawelpasterz, line 39: I am wondering why do we need `projectId` to make this request? I mean, I can see from the code that we use it to set a property for request under the hood, but why since it is publicly available information, maybe it's due to 'default' tag...I don't know and just thinking :)

piotradamczyk5: Probably device catalogue could be different for different project ids, so API needs this to return catalogue which is available for current `projectID`.
I played some time with documentation and API tester and I think that is the best explanation

test_runner/src/main/kotlin/ftl/environment/LocalesDescription.kt
adamfilipow92: Thanks! Changed

test_runner/src/main/kotlin/ftl/android/AndroidCatalog.kt
adamfilipow92: @Pasterzp requested change done. Thanks for the tests. I check if projectID is really required and get feedback

release_notes.md
adamfilipow92: Removed thanks :)

test_runner/src/main/kotlin/ftl/android/AndroidCatalog.kt
piotradamczyk5: As per [documenation](https://cloud.google.com/sdk/gcloud/reference#--project):

`--project` and its fallback` core/project` property play two roles in the invocation. It specifies the project of the resource to operate on. 

It has fallback option, so project is not specified in environmental variable and not provided by config for me it is ok to use empty

adamfilipow92:
> I know I am super annoying but let's also handle those ugly `%` chars at the end of line
> Screenshot 2020-08-12 at 18 06 50
> I don't know if this is only me but `gcloud` does not print them
Hmm, I don't see this when I testing. You have right it's should be fixed thanks!  I check and fix it.

test_runner/src/main/kotlin/ftl/android/AndroidCatalog.kt
bootstraponline: hey, we need projectId

the reason is select projects have access to secret device catalogs 🙂 These catalogs are not available to the public.

bootstraponline: > Probably device catalogue could be different for different project ids,

this is correct!

bootstraponline: Let's add a comment that explains why projectId is sent to the device catalog

adamfilipow92:
> For incorrect locale, gcloud ends run with exit code 1 where flank's s 0. I think we should also reflect this.

Thanks, changed 👍

PR #979, opened 2020-08-11
SmartFlank: Use average test time for new tests when possible

### LINKED ISSUE #977: SmartFlank default test time is too big

As a user, I want to reduce the default test time so I can waste less resources.

When I run flank with SmartFlank enabled I see it's using too many shards because it counts 120s for each test without previous times.
Flank suggest (and Fladle sets by default) a maximum shard time of 120s, so, each new test is using a shard for itself, and in our case it can run in 2 seconds, so, we are paying for 58 seconds that we don't use, multiplied per the number of new tests.

There are 2 scenarios:
- first build: then we need to use the default test time, so probably having a parameter for it is the right solution.
- new tests: probably here we could use the average duration from the other tests

With the average we would fix the worst issue because it happens all the time, but in any case DEFAULT_TEST_TIME_SEC == suggested SHARD_TIME doesn't make any sense.

None that fixes the issue without adding others.

**Additional context**
The docs say that the default is 10s but it was changed to 120s #635 making it a big issue that otherwise won't be too noticeable.

CristianGM:
Fixes #977 

I've modified the previous tests to show it's behavior change

## Checklist

- [ ] Documented
- [x] Unit tested

### DIFF
--- a/test_runner/src/main/kotlin/ftl/shard/TestMethodTime.kt
+++ b/test_runner/src/main/kotlin/ftl/shard/TestMethodTime.kt
@@ -11,8 +11,12 @@ const val DEFAULT_TEST_TIME_SEC = 120.0
 const val IGNORE_TEST_TIME = 0.0
 
 fun getTestMethodTime(flankTestMethod: FlankTestMethod, previousMethodDurations: Map<String, Double>): Double {
-    return if (flankTestMethod.ignored) IGNORE_TEST_TIME else previousMethodDurations.getOrDefault(
-        flankTestMethod.testName,
-        DEFAULT_TEST_TIME_SEC
-    )
+    return if (flankTestMethod.ignored) IGNORE_TEST_TIME else previousMethodDurations[flankTestMethod.testName] ?: avgTestTime(previousMethodDurations)
+}
+
+private fun avgTestTime(previousMethodDurations: Map<String, Double>): Double {
+    return previousMethodDurations.values
+        .filter { it > 0.01 }
+        .ifEmpty { listOf(DEFAULT_TEST_TIME_SEC) }
+        .average()
 }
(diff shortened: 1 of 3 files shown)


CristianGM:
Adding a flag for average would be nice, and IMHO adding a parameter for the default was a must.
I understand trying to limit the number of parameters but it seems the difference on what's the "default" for different devs is too big to just say 10s or 120s.

I'm assuming here that for a project all the tests are kind of the same (for sure there will be a variance, but I don't expect it to be minutes). Having a slightly bigger shard shouldn't be an issue, even smaller is fine if it's not under 1 minute so we have the $$$ issue (just paying for unused seconds, no drama but it's still $).

bootstraponline:
> @bootstraponline [#635 (comment)](https://github.com/Flank/flank/pull/635#issuecomment-581721754) still we don't want an option for it?

@jan-gogo Thanks for this comment Jan! I think you're correct we should have an option. I hoped that we could select one value for all the users, however an option makes the most sense.

bootstraponline:
> Adding a flag for average would be nice, and IMHO adding a parameter for the default was a must.

Agreed that we should have a parameter for the default. Thanks for sharing your use case so we can improve Flank!

CristianGM:
So...
Parameter for the default?
boolean parameter for using average "prediction"?

Do you want me to do the changes? or is it something you want to do? (first time i look at Flank code, but I guess adding the parameter won't be too complex, naming may be)

bootstraponline:
> Do you want me to do the changes? or is it something you want to do? (first time i look at Flank code, but I guess adding the parameter won't be too complex, naming may be)

I think it'll be faster to have the team do the changes. We'll discuss the best way to implement this.

bootstraponline:
I'm closing this pull request for now. We discussed as a team and prioritized the ticket in the ranked column on ZenHub. There are a few higher priority tasks that'll be worked on first (such as fixing the rate limit issue).

PR #987, opened 2020-08-12
docs: Analytics addition to readme

### LINKED ISSUE #973: Add analytics-uuid/Bugsnag to README
As an end user, I want to understand what analytics Flank sends and how to control them so I can make good decisions for my business.

The only mention of how to disable analytics (eg, BugSnag) is in code. That's findable, but a bit bad for end users. For example, this test class shows how to disable analytics. 

https://github.com/Flank/flank/blob/acd142d8062c3ab38ec5ca85c38bb21476d8587a/test_runner/src/test/kotlin/ftl/util/FlankBugsnagInitHelperTest.kt

It would be great if this was documented in the README so that it would be easier to understand what analytics Flank is sending. From there, I could make a better decision about leaving them enabled or turning them off. The README should also contain information about how to turn analytics off, for instance `echo "DISABLED" > ~/.gsutil/analytics-uuid`

Sloox:
Fixes #973 

## Checklist

- [x] Documented

### DIFF
--- a/docs/analytics.md
+++ b/docs/analytics.md
@@ -1,0 +1,10 @@
+# Flank Analytics
+
+Flank makes use of various analytics, below contains a list of the analytics and how to enable/disable them from being sent to FTL.
+
+## Flank
+
+ - BugSnag
+   - **Disable** - ADD "DISABLE" to the analytics_uuid for example: echo "DISABLED" > ~/.gsutil/analytics-uuid
+   - **Enable** - Remove "DISABLE" from the analytics_uuid file for example: echo "$(grep -v "DISABLED" ~/.gsutil/analytics-uuid)" > ~/.gsutil/analytics-uuid
+- 
--- a/docs/analytics.md
+++ b/docs/analytics.md
@@ -1,10 +1,22 @@
 # Flank Analytics
 
-Flank makes use of various analytics, below contains a list of the analytics and how to enable/disable them from being sent to FTL.
+Flank makes use of Bugsnag as its only analytics platform. It uses Bugsnag as an error monitor and project stability tool.
 
-## Flank
+For more information about Bugsnag please visit its website [here](https://www.bugsnag.com/) and documentation [here](https://docs.bugsnag.com/)
 
- - BugSnag
-   - **Disable** - ADD "DISABLE" to the analytics_uuid for example: echo "DISABLED" > ~/.gsutil/analytics-uuid
-   - **Enable** - Remove "DISABLE" from the analytics_uuid file for example: echo "$(grep -v "DISABLED" ~/.gsutil/analytics-uuid)" > ~/.gsutil/analytics-uuid
-- 
+## Control
+
+It is possible to disable the Bugsnag analytics integration used by flank or to make use of custom Bugsnag key so that any analytics can be displayed for custom use.
+
+ - To disable, place the word ```DISABLED``` in the file ````~/.gsutil/analytics-uuid```` and Flank will no longer send any analytics data
+   - Or simply running the following code (make sure the file exists!) will disable: ```echo "DISABLED" > ~/.gsutil/analytics-uuid```
+
+ - To enable or reenable bugsnag simply remove the file ```~/.gsutil/analytics-uuid``` and the project will default to make use of the flanks Bugsnag key
+
+ - To make use of custom Bugsnag integration, place your bugsnag key/UUID in ```~/.gsutil/analytics-uuid``` and the flank project will make use of it instead of the default key.
+
+ ## More information
+
+ To see how Bugsnag integration is used within the project please see the Flank Bugsnag testcase [here](../test_runner/src/test/kotlin/ftl/util/FlankBugsnagInitHelperTest.kt)
+
+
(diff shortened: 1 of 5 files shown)


docs/readme.md
```diff
@@ -75,6 +75,10 @@ Client may be generated manually using `testing_v1.json` and the master branch o
 
 - `gradle build`
 
+## Analytics
+
+Flank makes use of various analytics. For more information on what they are and how to control them please see [analyitics.md](analytics.md)
```
piotradamczyk5, line 80: I think that we should be consistent here, later on, we said that only Bugsnag is used (not various analytics)
```suggestion
Flank makes use of Bugsnag for analytics. For more information about analytics and how to control them please see [analyitics.md](analytics.md)
```

What do you think?

docs/analytics.md
```diff
@@ -0,0 +1,22 @@
...
+## Control
+
+It is possible to disable the Bugsnag analytics integration used by flank or to make use of custom Bugsnag key so that any analytics can be displayed for custom use.
+
+ - To disable, place the word ```DISABLED``` in the file ````~/.gsutil/analytics-uuid```` and Flank will no longer send any analytics data
+   - Or simply running the following code (make sure the file exists!) will disable: ```echo "DISABLED" > ~/.gsutil/analytics-uuid```
+
+ - To enable or reenable bugsnag simply remove the file ```~/.gsutil/analytics-uuid``` and the project will default to make use of the flanks Bugsnag key
+
+ - To make use of custom Bugsnag integration, place your bugsnag key/UUID in ```~/.gsutil/analytics-uuid``` and the flank project will make use of it instead of the default key.
```
bootstraponline, line 16: Is this true? I believe we only check the contents of this file for "DISABLED" A user can opt out of bugsnag, however they can't point flank to their own bugsnag instance.

docs/analytics.md
```diff
@@ -0,0 +1,22 @@
...
+
+Flank makes use of Bugsnag as its only analytics platform. It uses Bugsnag as an error monitor and project stability tool.
+
+For more information about Bugsnag please visit its website [here](https://www.bugsnag.com/) and documentation [here](https://docs.bugsnag.com/)
+
+## Control
+
+It is possible to disable the Bugsnag analytics integration used by flank or to make use of custom Bugsnag key so that any analytics can be displayed for custom use.
+
+ - To disable, place the word ```DISABLED``` in the file ````~/.gsutil/analytics-uuid```` and Flank will no longer send any analytics data
```
bootstraponline, line 11: Let's mention that Flank follows the same opt out approach as GCloud CLI. That's where this file and logic comes from.

docs/analytics.md
```diff
@@ -0,0 +1,22 @@
+# Flank Analytics
+
+Flank makes use of Bugsnag as its only analytics platform. It uses Bugsnag as an error monitor and project stability tool.
```
bootstraponline, line 3: maybe instead of analytics, it's more accurate to rebrand this as Error Monitoring

> Automated error monitoring, reporting, alerting
>
> https://www.bugsnag.com/

bugsnag tracks errors in flank, it doesn't watch what users do.

docs/readme.md
Sloox: Agreed. Thanks @piotradamczyk5. I thought it used various analytics at first beyond Bugsnag. But @bootstraponline cleared that it was only Bugsnag and i did not clean this definition up

docs/analytics.md
Sloox: It appears that I was mistaken here. There is a test class that works with a provided UUID which, after some investigation, is more important for gclouid than for flank. After another lookthrough the code if the user supplies a UUID it will simply make use of Flanks API key. Only Mock, Disabled will change the behaviour.

docs/analytics.md
Sloox: Sure done!

docs/analytics.md
Sloox: Sure reworded especially as analytics is ambiguous here.

docs/error_monitoring.md
```diff
@@ -0,0 +1,35 @@
...
+
+The goal is to use language that demonstrates value to Flank's customers.
+
+For example American express wants a stable and reliable test runner. Bugsnag is helping us deliver this too them.
+
+
+BugSnag sumarizes the following about how it makes use of the data it collects and stores:
+
+ - Bugsnag monitors web, mobile and server applications and programs under the direction of its Customers, to provide error Event Data about the Customer’s applications and programs to the   Customer. Event Data may contain Personal Information of the individuals who use the Customer’s applications. Bugsnag itself does not collect this end user Personal Information and Bugsnag has no direct relationship with these individuals. For example, a Customer may direct Bugsnag to log certain software application’s end user information associated with Event Data, such as the end user’s device ID, email and name. Bugsnag is a data processor and follows the instructions of the Customer in these cases.
```
bootstraponline, line 31: Let's specify exactly what Flank stores. For example we store the version of flank when flank crashes on bugsnag. We don't collect any personal information.

bootstraponline:
@Sloox I made a bunch of edits 😁 I think this looks good now!

PR #991, opened 2020-08-13
feat: Validate orientation and fail fast

### LINKED ISSUE #928: Fast fail when device orientation is misspelled

As a user, I want to fast fail when I misspell orientation so I can faster fix my config and avoid unnecessary upload apks to gcloud.

Fast fail with exit code: 18 when orientation is misspelled. 

**Additional context**
Verifying this parameter allows the user to keep some time. I checked gcloud cli and it verifying orientation before upload apks.

WDYT?
@bootstraponline @pawelpasterz @jan-gogo @piotradamczyk5

Sloox:
Fail fast if the Orientation is misspelt for runs
Fixes #928 

It has been unit tested.

## Checklist

- [x] Documented
- [x] Unit tested


test_runner/src/main/kotlin/ftl/args/ValidateCommonArgs.kt
```diff
@@ -8,13 +8,22 @@ fun CommonArgs.validate() {
     assertShardTime()
     assertRepeatTests()
     assertSmartFlankGcsPath()
+    assertOrientationCorrectness()
+}
+
+private fun CommonArgs.assertOrientationCorrectness() {
+    val orientationsList = listOf("portrait", "landscape", "default")
```
Sloox, line 15: This is the crux of the logic. The question here is hardcoded values ok? Or shall it actually go ahead and try to retrieve the values that are available? 
WDTTT?

test_runner/src/main/kotlin/ftl/args/ValidateCommonArgs.kt
```diff
@@ -8,13 +8,22 @@ fun CommonArgs.validate() {
...
+    devices.map { it.orientation }.forEach { ori ->
+        if (orientationsList.filter { ori.toLowerCase().contentEquals(it) }.size != 1)
```
adamfilipow92, line 17: Maybe
```
      if (ori.toLowerCase() !in orientationsList)

```
We can configure more thane one device so
what do you think about display all misspelled orientations on fail?

Sloox: well i must say that `!in `is a great keyword i am not familiar with thanks.
Totally doable :)

test_runner/src/main/kotlin/ftl/args/ValidateCommonArgs.kt
adamfilipow92: On the one hand, I think is ok because orientation values shouldn't be changed, but if you want to fetch this parameter from API look on AndroidCatalog.IosCatalog.supportedOrientationsAsTable

Sloox: @adamfilipow92 it what i was thinking too. In this case it might be fine as orientations for phones should not change.

test_runner/src/main/kotlin/ftl/args/ValidateCommonArgs.kt
```diff
@@ -8,13 +8,27 @@ fun CommonArgs.validate() {
     assertShardTime()
     assertRepeatTests()
     assertSmartFlankGcsPath()
+    assertOrientationCorrectness()
+}
+
+private fun CommonArgs.assertOrientationCorrectness() {
+    val orientationsList = listOf("portrait", "landscape", "default")
+    val mispeltOrientations = mutableListOf<String>()
+    devices.map { it.orientation }.forEach {
```
adamfilipow92, line 17: Maybe we can use filter instead of forEach and mutableList?

```kotlin

    val mispeltOrientations = devices.map { it.orientation }.filter { it !in orientationsList }

```

or more functional

```kotlin


private fun CommonArgs.assertOrientationCorrectness() = devices.deviceWithMisspelledOrientation(listOf("portrait", "landscape", "default")).throwIfAnyMisspelled()

private fun List<Device>.deviceWithMisspelledOrientation(availableOrientations: List<String>) = filter { it.orientation !in availableOrientations }.map { it.orientation }

private fun List<String>.throwIfAnyMisspelled() =
    if (isNotEmpty()) throw FlankConfigurationError("Orientation misspelled or incorrect, found ${joinToString()}. Aborting.")
    else Unit

```

But this only my idea. WDYT?

Sloox: I think for now i will rewrite it to the more functional way just because the project is focused this way and its good practise. :) else its such a small section i am not sure of its worth besides that.

pawelpasterz:
And I've just noticed that `gcloud` returns with exit code `1` (flank `2`)

Sloox:
> I like the solution, nice!
> I am wondering if we could print more detailed info about the device with misspelled orientation. In case when flank user has like 5 or more devices in the config file...just thinking

What information actually would be important, just the devices that have misspelt orientations?

Sloox:
> And I've just noticed that `gcloud` returns with exit code `1` (flank `2`) for the same misspelled orientation

Also

release_notes.md
```diff
@@ -1,4 +1,5 @@
 ## next (unreleased)
+- [#928](https://github.com/Flank/flank/pull/991) Fail fast on orientation mispell ([sloox](https://github.com/Sloox))
```
adamfilipow92, line 2: In [ ] should be the number of the pull request in this case [#991]

Sloox: Thanks 👍

PR #994, opened 2020-08-14
Fix size annotations support

pgreze:
This PR is fixing size annotation filtering.

So far annotation names for sizes were just specifying the class name without any package, but the resolution was reusing withAnnotation implementation, which should use the full path as shown by samples in gcloud documentation: https://cloud.google.com/sdk/gcloud/reference/firebase/test/android/run#--test-targets
So when requesting annotations for a class, a size annotation like "xxx.xxx.LargeTest" was not matching "LargeTest".

Another problem is that several annotations were [introduced with time](https://github.com/android/android-test/blob/461ca02299807020c0b317b1805eb709d3c25147/tools/device_broker/java/com/google/android/apps/common/testing/broker/HostTestSize.java#L28):
- historical ones (?): https://developer.android.com/reference/android/test/suitebuilder/annotation/package-summary.html
- Android Support like https://developer.android.com/reference/android/support/test/filters/LargeTest
- AndroidX like https://developer.android.com/reference/androidx/test/filters/LargeTest

At the end I adopted a strategy of just filtering based on the class name (as the original implementation was probably aiming for).
Not sure if it's completely matching androidx

## Test Plan

> How do we know the code works?

TestFiltersTest was already including tests for:
+ error if an invalid size is used
+ a test for each size
+ tests when using multiple sizes at the same time

I've slightly changed test samples in order to introduce full path of each generation of test sizes.

## Checklist

- [x] Documented
- [x] Unit tested


test_runner/src/main/kotlin/ftl/filter/TestFilters.kt
```diff
@@ -114,8 +114,18 @@ object TestFilters {
...
+    private fun withSize(args: List<String>): TestFilter = args.map { Size.from(it).annotationName }.let { annotationNames ->
+        TestFilter(
+            describe = "withSize (${annotationNames.joinToString(", ")})",
+            shouldRun = { testMethod ->
+                // Ensure that all annotation with a name matching this size are detected, like
+                // https://developer.android.com/reference/android/support/test/filters/LargeTest or
+                // https://developer.android.com/reference/androidx/test/filters/LargeTest
+                testMethod.annotationNames.any { it.split(".").last() in annotationNames }
+            },
+            isAnnotation = true
```
pgreze, line 126: The original implementation was reusing  withAnnotation so `isAnnotation` was true + size is annotation based I figure. Please let me know if this make sense or this was supposed to only match (not)withAnnotation instructions.

pawelpasterz:
Hey @pgreze thanks for your contribution, LGTM!
We need another reviewer and I think @piotradamczyk5 would be perfect.

pgreze:
@pawelpasterz Thank you for your approval!

@piotradamczyk5 please let me resolve conflicts with master after your review because the release_notes.md file is evolving quite often recently 😄

pawelpasterz:
> @pawelpasterz Thank you for your approval!
> 
> @piotradamczyk5 please let me resolve conflicts with master after your review because the release_notes.md file is evolving quite often recently 😄

Tell me about it, fortunately Piotr is working on some solution that will automate that

piotradamczyk5:
> > @pawelpasterz Thank you for your approval!
> > @piotradamczyk5 please let me resolve conflicts with master after your review because the release_notes.md file is evolving quite often recently 😄
> 
> Tell me about it 😺 , fortunately, Piotr is working on some solution that will automate appending `release_notes`

Yeah just skip this file 😄

piotradamczyk5:
`Check Pull Request / check_title (pull_request) ` fails because of Github actions does not work well with fork pull request so, do not worry about this

bootstraponline:
@pgreze can you grant edit access on the PR so we can resolve the conflict and merge?

pgreze:
@bootstraponline I guess this is involving giving you write access to my forked repository.
I granted you the permission, please let me know if it's enough.

Issue #1034, opened 2020-08-20
Firebase throws error on 250 virtual shards, says limit is now 200


RainNapper:
Version 20.08.01

```
com.google.api.client.googleapis.json.GoogleJsonResponseException: 400 Bad Request
POST https://testing.googleapis.com/v1/projects/test-lab-test-project/testMatrices
{
  "code" : 400,
  "errors" : [ {
    "domain" : "global",
    "message" : "Matrix expands to 250 supported executions which is more than the maximum allowed 200",
    "reason" : "badRequest"
  } ],
  "message" : "Matrix expands to 250 supported executions which is more than the maximum allowed 200",
  "status" : "INVALID_ARGUMENT"
}
        at com.google.api.client.googleapis.json.GoogleJsonResponseException.from(GoogleJsonResponseException.java:149)
   ```

This is the first run really looked at that was running on >200 shards, so I'm not sure if it is supposed to be 200 or 250 now.
https://github.com/Flank/flank/pull/915 changed it to 250, so I'm guessing it was changed recently on the firebase side.

May abandon this PR:
From firebase:
> Looks like one code location didn't get updated to the new 250 limit. We'll try to have a fix for that in the next release.

bootstraponline:
https://firebase-community.slack.com/archives/C1MTSQ5QT/p1597951056010300?thread_ts=1597950018.009700&cid=C1MTSQ5QT

bootstraponline:
issue should be resolved by FTL now, they're considering reworking the sharding limit.

https://firebase-community.slack.com/archives/C72V6UW8M/p1597979321010300?thread_ts=1597950801.008900&cid=C72V6UW8M

PR #1035, opened 2020-08-20
Drop virtual shard limit to 200

RainNapper:
Fixes https://github.com/Flank/flank/issues/1034

Build locally.
Verify firebase does not throw an error when we pass in 200.

This is still out of sync with https://cloud.google.com/sdk/gcloud/reference/beta/firebase/test/android/run, which still says 250.

## Checklist

- [no] Documented (Should we link https://cloud.google.com/sdk/gcloud/reference/beta/firebase/test/android/run#--num-uniform-shards for documentation?)
- [no] Unit tested


bootstraponline:
according to https://firebase-community.slack.com/archives/C1MTSQ5QT/p1597951056010300?thread_ts=1597950018.009700&cid=C1MTSQ5QT 250 is the correct limit.

Until FTL fixes the issue, probably the best work around is using 200 shards

PR #1043, opened 2020-08-22
feat: Add printing messages for MatrixCanceledError and InfrastructureError

### LINKED ISSUE #1036: feat: Add printing detailed messages for Flank's errors.
In Flank there is a function `Utlis#withGlobalExecptionHandling` which is a kind of wrapper to `main` and its duty is to handle all flank's custom errors.
`MatrixCanceledError` and `InfrastructureError` detailed messages are not printed. 

We should implement a logic to print detailed message (or default one)

pawelpasterz:
Fixes #1036 

The build is finished and all tests pass.

The actual change is in `ExceptionHandler.kt` file (which is basically extracted `withExceptionHandling` function).
You can skip the rest change, those are only changes in imports.

My bad I did not make it in separate commits


test_runner/src/main/kotlin/ftl/run/exception/ExceptionHandler.kt
```diff
@@ -0,0 +1,80 @@
...
+                exitProcess(UNEXPECTED_ERROR)
+            }
+        }
+    }
+}
+
+private val FlankException.messageOrUnavailable: String
+    get() = if (message.isNullOrBlank()) "[Unavailable]" else message as String
+
+private fun err(string: String?) = System.err.println(string)
```
piotradamczyk5, line 76: how about renaming it to `printError()`?

pawelpasterz: Sure, good point 👍

piotradamczyk5 (approved):

PR #1048, opened 2020-08-25
fix: Flank finding and running duplicate tests

MatthewTPage:
## The Problem

Flank is detecting more tests than there are in my test suit (there should be about 2,000 give or take, but it’s running 12,000). When I look at the junit test report in each matrix, I see that some tests are being ran once as expected, but some are being ran 5 times and some even 40+ times. Each test only appears once in my source code, num-test-runs is set to 1, and num-flaky-test-attempts is set to 0.

Flank version: v20.08.3
Example of android_shards.json with duplicated tests: https://gist.github.com/MatthewTPage/99dcf8a1876716f815b802c0bc2a5d41
(note that I obfuscated the class and method names, but the important information of "the tests are duplicated" is there).
Flank configuration that produced the above (with some bucket and apk names obfuscated): https://gist.github.com/MatthewTPage/ecd0c40fb4e089b6b6db028c716cb7f0

## The Solution

It seems that the distinct function call I changed in this PR wasn't working as intended - possibly it was distincting objects simply by the object reference rather than information about the actual test? So I made it distinct by the test name. Seems to work all fine and dandy.

## Verification

I ran this on my test suit and it came back with what seems like the right number of tests, and I don't see any duplicated.

## Unit Test Coverage

Seems a little excessive to figure out how to mock the dex parser, unless if someone would be strongly opinionated about me doing so and would have a suggestion.

### DIFF
--- a/test_runner/src/main/kotlin/ftl/run/platform/android/CreateAndroidTestContext.kt
+++ b/test_runner/src/main/kotlin/ftl/run/platform/android/CreateAndroidTestContext.kt
@@ -64,11 +64,11 @@ private fun InstrumentationTestContext.getFlankTestMethods(
 ): List<FlankTestMethod> =
     getParametrizedClasses().let { parameterizedClasses: List<String> ->
         DexParser.findTestMethods(test.local).asSequence()
-            .distinct()
-            .filter(testFilter.shouldRun)
-            .filterNot(parameterizedClasses::belong)
-            .map(TestMethod::toFlankTestMethod).toList()
-            .plus(parameterizedClasses.map(String::toFlankTestMethod))
+                .distinctBy { it.testName }
+                .filter(testFilter.shouldRun)
+                .filterNot(parameterizedClasses::belong)
+                .map(TestMethod::toFlankTestMethod).toList()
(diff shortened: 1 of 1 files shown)


MatthewTPage:
Hmmm it looks like the check_title check re-ran when I updated the PR's title but it's still failing. Not sure why.

pawelpasterz:
@MatthewTPage Hey! Thanks for your contribution!

> Hmmm it looks like the check_title check re-ran when I updated the PR's title but it's still failing. Not sure why.

I think GH actions does not work well with forked repos

pawelpasterz:
I like your solution 👍 
It would be great if you could add a test to our test suite. If not I think we could handle it

pawelpasterz (approved):

piotradamczyk5 (approved):

piotradamczyk5:
> @MatthewTPage Hey! Thanks for your contribution!
> 
> > Hmmm it looks like the check_title check re-ran when I updated the PR's title but it's still failing. Not sure why.
> 
> I think GH actions does not work well with forked repos

That's true however I found some workaround at #1051

bootstraponline:
unit test will be done in https://github.com/Flank/flank/issues/1049

PR #1054, opened 2020-08-26
fix: firebase refresh fails when test zip file doesn't exist (#1052)

### LINKED ISSUE #1052: firebase refresh fails when test zip file doesn't exist
**Describe the bug**
Running `flank firebase refresh` for ios project fails when the `test` zip file does not exist, while it does not use it.

We run the tests in async mode in our CI setup, and then we run the refresh later from another machine. If it is not necessary we don't want to be moving the test zip file just for the refresh to run.

**To Reproduce**

Flank version: v20.08.2

1. Build app and run `flank firebase test ios run --async`
2. Store `matrix_ids.json`
3. On clean workspace (another machine) restore the `matrix_ids.json` and run 'flank firebase refresh'
2. See error `'/home/runner/_work/ios/ios/UITests-Build.zip' not found (UITests-Build.zip)`

**Expected behavior**

We expect not to look for the zip file and run the refresh.

```
version: v20.08.2
revision: 5f9a81d148ae89a84e5c12410eef75ea9abbf698
```

**Additional info**

When we "fake" the zip file just by creating empty file, the refresh run is ok - indicating that the zip file is not even used at all during the refresh command.

axelzuziak-gogo:
Fixes #1052 

* renamed `String.proccessFilePath` method to `String.normalizeFilePath`
* catching errors from `ArgsHelper.evaluateFilePath`
* removed validation from `IosArgsCompanion.load` - only for iOS for now. We need different validations for different commands.

TODO:
~* remove validation from `AndroidArgsCompanion.load`~
~* update tests~

* The test passes
* The `firebase refresh` command doesn't fail with iOS config when `xctestrunZip` and `xctestrunFile` are missing

## Conclusions
This error occurs because `firebase refresh` command uses the same validation as `android run`.
The solution is to create dedicated `args` validators for specific commands.
Additionally all fast fails will be moved from `args` creation into `args` validation layer.
As result this fix will bring also some changes in android implementation because some fast fails currently are shared between android and ios args creation.

## Checklist

- [ ] Documented
- [ ] Unit tested


adamfilipow92 (approved):
Great work 👍

jan-gogo (approved):

Issue #936, opened 2020-08-05
Add full-junit-result to Fladle


bootstraponline:
https://github.com/runningcode/fladle/issues/157

adamfilipow92:
If I fully understand.

If we want to use smart-flank we need to have enabled one of the options
1. full-junit-result
2. legacy-junit-result

WIthout one of that smart-flank not working
My proposal is
1. Fix legacy-junit-result flag [947](https://github.com/Flank/flank/issues/947)
2. Add fast fail when smart-flank enabled but any of junit command not set with clearly explain what to do.
3. Add docs about using smart-flank mentioned here [902](https://github.com/Flank/flank/issues/902)

WDYT?
@bootstraponline @jan-gogo @Pasterzp @piotradamczyk5 @Sloox

bootstraponline:
@adamfilipow92 I think that makes sense!

Do we know why smart flank requires those junit options? I think we just need the timing info from a complete run.

bootstraponline:
https://github.com/Flank/flank/blob/master/docs/smart_flank.md is the information we have documented about smart flank.

adamfilipow92:
> @adamfilipow92 I think that makes sense!
> 
> Do we know why smart flank requires those junit options? I think we just need the timing info from a complete run.

In JUnitResult we remove fail nodes if any success [#798] https://github.com/Flank/flank/pull/798 and I think there can be a problem but I need to more investigate it to be 100% sure.

bootstraponline:
> In JUnitResult we remove fail nodes if any success [#798](https://github.com/Flank/flank/pull/798) and I think there can be a problem but I need to more investigate it to be 100% sure.

I think smart flank ignores failed nodes though (as specified in the smart flank doc).

adamfilipow92:
Hmmm, true, smart flank using only passed tests. I analyzed where is the problem. When we did not set legacy/full junit flag, flank not uploading smart-flank report. I tested https://github.com/Flank/flank/pull/935/files and this fixes the problem.

adamfilipow92:
For now, we can run smart flank without set ```--full-junit-result``` and ```--legacy-junit-result``` but if we use FullJUnitResult in gcs path and current run don't have set flag ```--full-junit-result``` flank override full junit result by junit result it's can be little confusing. I add fast fail to verify if we use the same type of JUnit result in this pr: https://github.com/Flank/flank/pull/1056

bootstraponline:
sounds good!

[...]

PR #1056, opened 2020-08-26
feat: Added validation of smart-flank-gcs-path to not override different junit results

### LINKED ISSUE #1057: Validate of smart-flank-gcs-path to not override different results type

As a flank user, I want to be informed when flank can override JUnitResult configured in ```smart-flank-gcs-path``` with a different type in the current run so I can easily avoid situations like that 

If the test is flaky, flank removing stack traces from JUnitResult, FullJunitResult keeps stack traces.

Fast fail when Flank can potentially override JUnitResult with different report type.

adamfilipow92:
After some discussion with @jan-gogo about https://github.com/Flank/flank/issues/936, this solution can be useful to avoid override results by different result types.  

Fixes #1057 

1. If we set FullJUnitResult.xml in ```smart-flank-gcs-path``` and in run we haven't set flag ```--full-junit-result``` flank should fast fail

2. If we set JUnitResult.xml in ```smart-flank-gcs-path``` and in run we have set flag ```--full-junit-result``` flank should fast fail

3. If user set different XML name in ```smart-flank-gcs-path``` flank shouldn't validate type. 

## Checklist

- [X] Unit tested

### DIFF
--- a/docs/bugs/smart_flank_upload_results_rules.md
+++ b/docs/bugs/smart_flank_upload_results_rules.md
@@ -1,0 +1,25 @@
+# Smart Flank rules of validation result types
+
+Flank trying to avoid override smart-flank-gcs-path by different JUnit report type. That's means: 
+
+1. If user select in ```smart-flank-gcs-path``` command FulJUnitResult.xml and flag ```--full-junit-result``` is not set Flank fail with the message
+
+    ```txt
+
+    smart-flank-gcs-path is set with FullJUnitReport.xml but in this run --full-junit-result is disabled, please set --full-junit-result flag
+
+
+    ```
+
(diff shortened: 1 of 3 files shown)


pawelpasterz:
I think this change is worth being documented.

pawelpasterz:
How about the case when `smartFlankDisableUpload = true`. IMO we can skip this validation

Sloox (approved):

adamfilipow92:
> How about the case when `smartFlankDisableUpload = true`. IMO we can skip validation then

Done!

> I think this change is worth being documented.

Docs added, thanks!

piotradamczyk5 (approved):

PR #1067, opened 2020-08-28
fix: Matrix path not found in json

### LINKED ISSUE #844: [iOS] matrix path not found in json
**Describe the bug**
I see the following warning for all shards in the flank ouput `matrix path not found in json ...`

**To Reproduce**
Steps to reproduce the behaviour:
1. Run ios test suite via flank
2. See warning in the output

**Expected behaviour**
none warning should be shown

 - Have you tested on the latest Flank snapshot? no
 - flank version: v20.06.0
- revision: 77f18db6109b49496fe82c99a553f16414f7366a

Screenshot 2020-06-11 at 10 40 23

piotradamczyk5:
Fixes #844

- regex to find matrix file find proper shard name, but later on created path was not properly, because of duplicated `objectName`
- when using legacy junit result for Android and for iOS warning is printed when results-dir exist on GCloud Storage 

## Checklist

- [x] Unit tested


Sloox (approved):

Sloox (approved):

adamfilipow92:
Check this issue: https://github.com/Flank/flank/issues/1068 I think we should rethink about fast-fail. Maybe we can clear the bucket before a run. WDYT?

test_runner/src/main/kotlin/ftl/reports/util/ReportManager.kt
```diff
@@ -28,21 +28,19 @@ import kotlin.math.roundToInt
...
-            if (it.name.matches(Artifacts.testResultRgx)) {
-                xmlFiles.add(it)
-            }
+    private fun findXmlFiles(
+        matrices: MatrixMap,
+        args: IArgs
+    ) = File(resolveLocalRunPath(matrices, args)).walk()
+        .filter { it.name.matches(Artifacts.testResultRgx) }
+        .fold(mutableListOf<File>()) { xmlFiles, file ->
+            xmlFiles.apply { add(file) }
```
jan-gogo, line 37:
```suggestion
        .fold(listOf<File>()) { xmlFiles, file ->
            xmlFiles + file
```

pawelpasterz:
I am not sure if this implementation covers all cases. When I launched android flank run but with legacy results I received
Screenshot 2020-09-07 at 14 01 52
I think the common denominator is parsing XML based results (for iOS and legacy android). We should double-check if web link is present (or not) in XML fetched from FTL (when the legacy mode is on)

piotradamczyk5:
> I am not sure if this implementation covers all cases. When I launched android flank run but with legacy results I received
> Screenshot 2020-09-07 at 14 01 52
> I think the common denominator is parsing XML based results (for iOS and legacy android). We should double-check if web link is present (or not) in XML fetched from FTL (when the legacy mode is on)

As we talked yesterday there is no perfect method to handle it, to avoid all of these errors we should clear result directory or fast fail if it already exists, so, for now, there is a warning displayed which clearly show information to the user why reusing results directory is a bad idea

pawelpasterz:
> > I am not sure if this implementation covers all cases. When I launched android flank run but with legacy results I received
> > Screenshot 2020-09-07 at 14 01 52
> > I think the common denominator is parsing XML based results (for iOS and legacy android). We should double-check if web link is present (or not) in XML fetched from FTL (when the legacy mode is on)
> 
> As we talked yesterday there is no perfect method to handle it, to avoid all of these errors we should clear result directory or fast fail if it already exists, so, for now, there is a warning displayed which clearly show information to the user why reusing results directory is a bad idea

I think we are talking about the different cases here. One that you mentioned -- yeah, that's correct and I totally agree with.
What I am talking about is (at least if I remember correctly) different XML structure in the past and once it was changed on FTL side, flank missed that update and `webLinks` are not reachable anymore (in old way). I think pure FTL xml based results are different for iOS and Android (but need to double-check it).

Just checked with empty local results dir and unique bucket name. Of course with legacy results

adamfilipow92:
I tested solution wth ```--legacy-junit-result```

If you look on ReportManager.findMatrixPath  we looking for a matrix with gcs path ends with  something like ```2020-09-08_10-58-22.764234_jdxf/matrix_0``` but on the end of gcs path we have an additional ```/```

```

test-lab-v9cn46bb990nx-kz69ymd4nm9aq/2020-09-08_10-58-22.764234_jdxf/matrix_0/

```
I modyfi line ReportManager:203 to ```.firstOrNull { savedMatrix -> savedMatrix.gcsPath.endsWith(this).plus("/") }```

and now it works  (```getWebLink``` returning link instead of null, flank don't print error).

piotradamczyk5:
@Mergifyio rebase

pawelpasterz:
@adamfilipow92 
> I modyfi line ReportManager:203 to .firstOrNull { savedMatrix -> savedMatrix.gcsPath.endsWith(this).plus("/") }

Did you mean `savedMatrix.gcsPath.endsWith(this.plus("/"))`

Perfect, so let's add this as well. This change doesn't break iOS run?

adamfilipow92:
> @adamfilipow92
> 
> > I modyfi line ReportManager:203 to .firstOrNull { savedMatrix -> savedMatrix.gcsPath.endsWith(this).plus("/") }
> 
> Did you mean `savedMatrix.gcsPath.endsWith(this.plus("/"))`
> 
> Perfect, so let's add this as well. This change doesn't break iOS run?

In iOS, this solution breaks so we should think about how to solve this conflict.

piotradamczyk5:
> > @adamfilipow92
> > > I modyfi line ReportManager:203 to .firstOrNull { savedMatrix -> savedMatrix.gcsPath.endsWith(this).plus("/") }
> > 
> > 
> > Did you mean `savedMatrix.gcsPath.endsWith(this.plus("/"))`
> > Perfect, so let's add this as well. This change doesn't break iOS run?
> 
> In iOS, this solution breaks so we should think about how to solve this conflict.

@pawelpasterz @adamfilipow92 do we know why `gcsPath` is different for iOS and Android?

piotradamczyk5:
> > > @adamfilipow92
> > > > I modyfi line ReportManager:203 to .firstOrNull { savedMatrix -> savedMatrix.gcsPath.endsWith(this).plus("/") }
> > > 
> > > 
> > > Did you mean `savedMatrix.gcsPath.endsWith(this.plus("/"))`
> > > Perfect, so let's add this as well. This change doesn't break iOS run?
> > 
> > 
> > In iOS, this solution breaks so we should think about how to solve this conflict.
> 
> @pawelpasterz @adamfilipow92 do we know why `gcsPath` is different for iOS and Android?
> 
> **Edit**
> I am proposing to making regex to match end with `/` or without it

Done, should works fine from now

adamfilipow92 (approved):
Tested with:
```--legacy-junit-result```

ios

without ```--legacy-junit-result```

everything working, great work 👍

piotradamczyk5:
@Mergifyio refresh

PR #1071, opened 2020-09-01
build: Added options to minimize the shadowJar process

### LINKED ISSUE #1063: ShadowJar minimization
Investigate the option of turning the minimisation of the shadowjar gradle task to ON.
Also check the consequences as removing classes form the build could lead to errors due to class not found issues.

**Additional context**
More information here:https://imperceptiblethoughts.com/shadow/configuration/minimizing/
Apart of the #774 issue chain

Sloox:
Fixes #1063 

Tests passes and tested manually with bash files ./update-flank and the executable flank.
For example: `flank firebase test android run --output-style=single -c=../src/test/kotlin/ftl/fixtures/test_app_cases/flank-multiple-flaky.yml --project=flank-open-source` 

`flank-multiple-flaky.yml` needs a minor update if run from the bash folder, the pathing is incorrect. eg:
```
 app: ../src/test/kotlin/ftl/fixtures/tmp/apk/app-debug.apk
 test: ../src/test/kotlin/ftl/fixtures/tmp/apk/app-multiple-flaky-debug-androidTest.apk
```

The difficulty associated with minimizing a build results in possible missing classes within the shadowJar which is a difficult thing to pick up as you need integration tests rather than unit tests to achieve this. The tests were run manually and have passed. I do suggest we consider having a set of manually run external tests that can be run 
WDYT?

### DIFF
--- a/test_runner/build.gradle.kts
+++ b/test_runner/build.gradle.kts
@@ -27,9 +27,23 @@ shadowJar.apply {
     archiveClassifier.set("")
     archiveBaseName.set(artifactID)
     mergeServiceFiles()
+    minimize {
+        exclude(dependency(Libs.KOTLIN_REFLECT))
+        exclude(dependency(Libs.JACKSON_XML))
+        exclude(dependency(Libs.JACKSON_DATABIND))
+        exclude(dependency(Libs.JACKSON_KOTLIN))
+        exclude(dependency(Libs.JACKSON_YAML))
+        exclude(dependency(Libs.GSON))
+    }
     @Suppress("UnstableApiUsage")
     manifest {
         attributes(mapOf("Main-Class" to "ftl.Main"))
+    }
+    dependencies {
+        exclude(dependency(Libs.TRUTH))
+        exclude(dependency(Libs.MOCKK))
+        exclude(dependency(Libs.JUNIT))
+        exclude(dependency(Libs.DETEKT_FORMATTING))
     }
 }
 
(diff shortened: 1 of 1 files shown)


jan-gogo:
> I do suggest we consider having a set of manually run external tests that can be run
WDYT?

Yup, you are absolutely right. I was thinking about automated integration tests some times ago. 👍

Sloox:
> > I do suggest we consider having a set of manually run external tests that can be run
> > WDYT?
> 
> Yup, you are absolutely right. I was thinking about automated integration tests some times ago. 👍

it would greatly help for reliability on this topic as its almost impossible to know if the minimization process has broken the jar and the time taken to actually test this is huge.

jan-gogo (commented):
The changes looks ok, time for testing 😄

jan-gogo:
> it would greatly help for reliability on this topic as its almost impossible to know if the minimization process has broken the jar and the time taken to actually test this is huge.

I am not afraid of minimization issues. We didn't use reflection in flank as I know, all type usage are explicit. The minimization shouldn't remove necessary types

Sloox:
> > it would greatly help for reliability on this topic as its almost impossible to know if the minimization process has broken the jar and the time taken to actually test this is huge.
> 
> I am not afraid of minimization issues. We didn't use reflection in flank as I know, all type usage are explicit. The minimization shouldn't remove necessary types

Interesting there was issues with Kotlin_reflect, jackson and gson, they all caused issues with minimization.

adamfilipow92 (approved):
Tested:
- [X] simple ios
- [X] android flaky
- [X] multiple
- [X] single
- [X] multi modules
- [X] with smart flank
- [X] with full junit result
- [X] with legcy junit result
All works! Great work! 👍

jan-gogo (approved):

Issue #1072, opened 2020-09-01
Create & Run Integration Tests for Flank


Sloox:
**Author the user story for this feature**

As a developer, I want to run integration test on flank so I can test the library with a set of default test cases as though i was using flank as a user.

**Is your feature request related to a problem? Please describe.**
It is linked to the need to test a full production based build of flank, and to test if the minimization of flank will not introduce errors. It can also help with bug testing issues where we can replicate the issue by running it on a production version of flank.

**Describe the solution you'd like**
Either have it built into test_runner project or a seperate project like flank_scripts, whereby we can run a set of integration tests simply and easily.

Sloox:
@bootstraponline @jan-gogo @pawelpasterz @piotradamczyk5 @adamfilipow92 
do we have any **_preferences/ideas/comments_** as to how integrations testing may look like for this task?

shall it appear as a subproject i.e. test_integration with its own gradle build etc? Will there actually be any code? or all gradle commands? Should we head in the direction of the flank_scripts and have a fully fledged sub project that we can run?

At the very least it should in my opinion start by compiling flank and either copying that lib/binary out of the test_runner folder and running a set of tests on that binary so we have a reference point.

Any ideas or suggestions would be greatly appreciated.
WDYT?

bootstraponline:
> Should we head in the direction of the flank_scripts and have a fully fledged sub project that we can run?

Yes. I think we'd create a new Kotlin project responsible for running Flank and verifying the tests execute successfully. The integration tests can be JUnit tests. Inside the Junit test, we'd use a freshly built version of Flank to run an existing apk/ipa and verify that the final result from Firebase Test Lab is what we expect.

I think these should be smoke tests.

>Smoke tests are a subset of test cases that cover the most important functionality of a component or system, used to aid assessment of whether main functions of the software appear to work correctly.[
>
> https://en.wikipedia.org/wiki/Smoke_testing_(software)

The 3 critical integration tests I see are:
- Given a test apk on Android, does the test run successfully
- Given a test bundle for iOS, does the test run successfully on macOS
- Given a test bundle for iOS, does the test run successfully on Linux

Each test should be simple, designed to execute quickly, and be stable. If we have 3 solid integration smoke tests, then we can think about adding more advanced tests in the future.

jan-gogo:
> Each test should be simple, designed to execute quickly, and be stable. If we have 3 solid integration smoke tests, then we can think about adding more advanced tests in the future.

Right, I love the idea to get something simple first.

PR #1078, opened 2020-09-02
feat: add proguard initial step into Flank configuration

### LINKED ISSUE #1064: R8 And/Or Progaurd
Investigate the use of R8 or Progaurd as a code optimisation, obsfucation and or minimisation tool.

**Additional context**
See more information here: https://jakewharton.com/shrinking-a-kotlin-binary/
Apart of the https://github.com/Flank/flank/issues/774 issue chain.

Sloox:
Fixes #1064 

Adds proguard optimization & minimization onto flank binary. Effectively reducing the binary size to 20MB at time of this pull request. 
This particular pull request does not influence the flank binary size yet due to constraint of: https://github.com/Flank/flank/issues/1072

For now a new task is made and can be run via ./gradlew applyProguard.
It will create and update the flank.jar file, placing it in bash folder under a new name: flank-proguard.jar.  This jar can then be used to 
We done we are waiting for https://github.com/Flank/flank/issues/1072
For now however the basic test cases work for the yaml files.

Again i must reiterate. It does not influence binary size as of yet. However the work was done before the creation of other tasks.


test_runner/proguard/config.pro
```diff
@@ -0,0 +1,70 @@
...
+-keep class ftl.** { *; }
+-keep interface ftl.** { *; }
+
+#google auth
+-keep class com.google.auth.** { *; }
+-keep interface com.google.auth.** { *; }
+
+#ctc wstx
+-keep class com.ctc.wstx.** { *; }
+-keep interface com.ctc.wstx.** { *; }
```
jan-gogo, line 70:
```suggestion
-keep interface com.ctc.wstx.** { *; }

```

jan-gogo: There is an option in Intellij that automatically adds an empty line on save but I don't remember the name.

Sloox: Actually have it installed. It maybe didn't apply it as the file type is not supported

test_runner/build.gradle.kts
```diff
@@ -31,6 +31,13 @@ shadowJar.apply {
     manifest {
         attributes(mapOf("Main-Class" to "ftl.Main"))
     }
+    dependencies {
```
pawelpasterz, line 34: We have a dependency block at L#186, I assume this is intentional?

Sloox: Yes, it appears that if i put it at L186 it is not working as intended. Hence why it is put here. However with the changes coming with the single gradle project it most likely wont be needed anymore.

pawelpasterz: Interesting, might be due too the fact that Gradle process script from top to bottom?

pawelpasterz (approved):

piotradamczyk5 (approved):

PR #1097, opened 2020-09-07
feat: print the matrices web link at the end of a run

### LINKED ISSUE #1094: Print out matrix weblink at the end of runs in addition to at the beginning
As a developer, I want to be able to access the matrix weblink at the end of the output, so that I can easily find it when CI might truncate console output to the final lines. This isn't a big deal, but would be a nice quality of life improvement that would save having to click to "view the full output" just to find the weblink url. 

On Circle CI, and other CI systems, console output is truncated to just the final few lines. Because Flank only prints out the weburl at the start of a run, it's not available without going through an additional step to view the job's full output. 

The matrix weburl and id should be printed both at the start and the end of a run.

adamfilipow92:
Fixes #1094

On the end of the output Flank should reprint matrices webLinks like:

```

MatrixResultsReport
  1 / 1 (100.00%)
┌─────────┬──────────────────────┬────────────────────────────┬──────────────────────┐
│ OUTCOME │      MATRIX ID       │      TEST AXIS VALUE       │     TEST DETAILS     │
├─────────┼──────────────────────┼────────────────────────────┼──────────────────────┤
│ success │ matrix-a6sbyeuihlwta │ NexusLowRes-28-en-portrait │ 20 test cases passed │
└─────────┴──────────────────────┴────────────────────────────┴──────────────────────┘
  Uploading MatrixResultsReport.txt .
  Uploading JUnitReport.xml .

Matrices webLink
  matrix-a6sbyeuihlwta https://console.firebase.google.com/project/flank-open-source/testlab/histories/bh.da0c237aaa33732/matrices/9050578927670125735


Process finished with exit code 0


```

### DIFF
--- a/test_runner/src/main/kotlin/ftl/run/NewTestRun.kt
+++ b/test_runner/src/main/kotlin/ftl/run/NewTestRun.kt
@@ -5,6 +5,7 @@ import ftl.args.IArgs
 import ftl.args.IosArgs
 import ftl.json.SavedMatrix
 import ftl.json.updateMatrixMap
+import ftl.json.validate
 import ftl.reports.util.ReportManager
 import ftl.run.common.fetchArtifacts
 import ftl.run.common.pollMatrices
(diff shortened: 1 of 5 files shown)


test_runner/src/main/kotlin/ftl/run/NewTestRun.kt
```diff
@@ -23,8 +24,14 @@ suspend fun newTestRun(args: IArgs) = withTimeoutOrNull(args.parsedTimeout) {
     if (!args.async) {
         cancelTestsOnTimeout(args.project, matrixMap.map) { pollMatrices(matrixMap.map.keys, args).updateMatrixMap(matrixMap) }
         cancelTestsOnTimeout(args.project, matrixMap.map) { fetchArtifacts(matrixMap, args) }
-
-        ReportManager.generate(matrixMap, args, testShardChunks, ignoredTests)
+        runCatching {
+            ReportManager.generate(matrixMap, args, testShardChunks, ignoredTests)
```
adamfilipow92, line 28: After report generating, flank validates matrices if any in other state than success flank throws an exception, we need to catch this error print matrices, and rethrow the error.  Maybe there is a better way to realize it or maybe we can return the validation state by an enum and throw here. WDYT?

Sloox (approved):

test_runner/src/main/kotlin/ftl/run/NewTestRun.kt
jan-gogo: Perfect question. IMO, validation is a different step than generating report, so `matrices.validate(args.ignoreFailedTests)` function shouldn't be called in `ReportManager.generate` but in the `newTestRun` and `refreshLastRun`.
After that `runCatching` and rethrow shouldn't be necessary.

adamfilipow92: Thanks for the feedback! I move validation outside of report generation. 👍

piotradamczyk5 (approved):

PR #1104, opened 2020-09-09
feat: send slack message for releases

### LINKED ISSUE #1096: Send a slack message via xoxc token & github actions

As a firebase community slack user I want to be able to know when slack is released.

Firebase community will most likely not accept the installation of a bot. Therefore to be able to notify the community it will be necessary to make use of the private api to send messages about releases

A github action that takes a valid xoxc token and sends a release message via slack to the firebase community

**Additional context**

Sloox:
Fixes #1096 

New iteration of slack message sending baked into a custom github action.

Will run tests on local slack then swap over to firebase community


action.yml
```diff
@@ -0,0 +1,23 @@
+name: slack-user-bot
```
piotradamczyk5, line 1: Could you rename this file to more meaningful name ex. `post_slack_message_action.yml`

flank-actions/slackService.kt
```diff
@@ -0,0 +1,36 @@
...
+import com.github.kittinunf.fuel.Fuel
+import com.github.kittinunf.fuel.core.Headers
+
+fun sendMessage(args: Array<String>): Int{
+    val token = args[Common.ARGS_TOKEN]
+    val channel = args[Common.ARGS_CHANNEL]
+    val message = args[Common.ARGS_MESSAGE]
+    val cookie = args.copyOfRange(Common.ARGS_COOKIE, args.lastIndex).joinToString(" ")
+
+    debug("Sending message:$message to channel:$channel with token: $token with cookie:$cookie")
```
piotradamczyk5, line 11: is it safe to print token?

flank-actions/sendMessage.kts
```diff
@@ -0,0 +1,15 @@
...
+//INCLUDE slackService.kt
+//INCLUDE common.kt
+
+val result = sendMessage(args)
+
+if (result != 0){
+    exitProcess(result)
+}
+
+println("Message has been sent with result $result")
```
piotradamczyk5, line 15: How about?

```suggestion
if (sendMessage(args) != 0){
    exitProcess(result)
}

println("Message has been sent")
```

Because there will be always `0` as result

piotradamczyk5 (changes requested):

flank-actions/slackService.kt
bootstraponline: no, good catch!

action.yml
Sloox: i cant. Its required to be this name

flank-actions/slackService.kt
Sloox: Removed!

flank-actions/sendMessage.kts
Sloox: Ive changed it as this could be modified in the future so I have moved the printLn elsewhere.

piotradamczyk5 (approved):

pawelpasterz (approved):

Issue #896, opened 2020-07-16
Move binaries from resources to external repo (or other place)


piotradamczyk5:
**Author the user story for this feature**

As a Developer, I want to move all binaries from `main/resources` to someplace to avoid using submodules (based on the research of task #849) so I can have a flank repository smaller.

**Is your feature request related to a problem? Please describe.**
External binaries should not be part of source code, to have quicker checkout and to keep repository size as small as possible.

**Describe the solution you'd like**
The solution should be chosen based on the research of task #849

Sloox:
The integration task might influence this: https://github.com/Flank/flank/issues/1072

piotradamczyk5:
Since binaries directory size is around 18,2 MB and the content inside is not changing so often, we could use GitHub lfs which has suitable storage for us.
What is more, these files are only needed for Linux and windows.
WDYT?

piotradamczyk5:
Based on slack discussion

> bandwidth for lfs on github is expensive
> outages are caused when the bandwidth is is exceeded
> there's flakiness associated with lfs which make the builds unstable
> the recommended approach is to publish these as a package (either to github packages) or a content store like GCS or github releases (which is backed by S3)

PR #1107, opened 2020-09-09
fix: Running iOS test on Linux

piotradamczyk5:
Fixes #

.

## Checklist

- [ ] Documented
- [ ] Unit tested


.github/workflows/ubuntu-workflow.yml
```diff
@@ -16,6 +16,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with: 
+          submodules: true
```
bootstraponline, line 20: I think we should move away completely from using submodules if possible.

piotradamczyk5: I totally agree, we have a [task](https://github.com/Flank/flank/issues/896) for it, but for now, this is a critical issue and should be fixed fast

Sloox (approved):

bootstraponline (approved):

PR #1109, opened 2020-09-10
feat: Integration tests for Flank

### LINKED ISSUE #1072: Create & Run Integration Tests for Flank

adamfilipow92:
Fixes #1072 

Integration tests should run successfully after build on Linux/Mac and in the future on windows.

Doc can be found [here](docs/instrumentation_tests.md)

## Checklist

- [X] Documented
- [X] Test for Android
- [X] Test for iOS
- [X] Prepared artifacts and ymls dedicated to the integration tests - I add simple ymls for tests but artifacts are used same in fixtures. Let me know its fine.
- [X] Integration with GitHub actions


Sloox:
Great. excited for this 👍

.github/workflows/release.yml
```diff
@@ -65,7 +65,19 @@ jobs:
...
+      with:
+        gradle-executable: "./integration_tests/gradlew"
+        arguments: "-p integration_tests test --tests IntegrationTests.shouldMatchAndroidSuccessExitCodeAndPattern -Dflank-path=../test_runner/build/libs/flank.jar -Dyml-path=./src/test/resources/flank_android.yml"
+    
+    - name: Gradle Integration Tests iOS
+      uses: eskatos/gradle-command-action@v1
+      with:
+        gradle-executable: "./integration_tests/gradlew"
+        arguments: "-p integration_tests test --tests IntegrationTests.shouldMatchIosSuccessExitCodeAndPattern -Dflank-path=../test_runner/build/libs/flank.jar -Dyml-path=./src/test/resources/flank_ios.yml"
+
```
piotradamczyk5, line 80: I think that they should be called on each PR (in `macOS` and `Ubuntu` workflows) to prevent merging code which brokes Flank behaviour.

WDYT?

integration_tests/buildSrc/src/main/kotlin/Deps.kt
```diff
@@ -0,0 +1,139 @@
+object Versions {
```
piotradamczyk5, line 1: do we need all these dependencies?

.github/workflows/release.yml
```diff
@@ -65,7 +65,19 @@ jobs:
...
+      uses: eskatos/gradle-command-action@v1
+      with:
+        gradle-executable: "./integration_tests/gradlew"
+        arguments: "-p integration_tests test --tests IntegrationTests.shouldMatchAndroidSuccessExitCodeAndPattern -Dflank-path=../test_runner/build/libs/flank.jar -Dyml-path=./src/test/resources/flank_android.yml"
+    
+    - name: Gradle Integration Tests iOS
+      uses: eskatos/gradle-command-action@v1
+      with:
+        gradle-executable: "./integration_tests/gradlew"
+        arguments: "-p integration_tests test --tests IntegrationTests.shouldMatchIosSuccessExitCodeAndPattern -Dflank-path=../test_runner/build/libs/flank.jar -Dyml-path=./src/test/resources/flank_ios.yml"
```
Sloox, line 79: if we want to add more integration tests. is this line going to get longer and longer?

integration_tests/buildSrc/src/main/kotlin/Deps.kt
Sloox: was going to ask the same thing

Sloox: And if we do, we should check if they will always match with what is done in flank test_runner and if it does then we should have a more global build config

Sloox: infact if i am correct @jan-gogo is working on a mono gradle repo and i am sure this would solve it :)

Sloox:
@adamfilipow92 so for this project it is only running one integration test? one for ios and one for android?

.github/workflows/release.yml
adamfilipow92: You can also define all paths in the test and run it by class or run all. But maybe in our purposes, these parameters are redundant? WDYT?

Sloox: I think what I am trying to discuss is, is our goal here to run 1big test that does all the integration test? or lots of smaller tests so we could simply add test after test after test.

adamfilipow92:
> @adamfilipow92 so for this project it is only running one integration test? one for ios and one for android?

Yes, this start point as mentioned in https://github.com/Flank/flank/issues/1072#issuecomment-687116654

I think the second step should be add run these tests on different OS (Linux/Mac/Windows). WDYT?

.github/workflows/release.yml
Sloox: Good idea!

piotradamczyk5:
> > @adamfilipow92 so for this project it is only running one integration test? one for ios and one for android?
> 
> Yes, this start point as mentioned in [#1072 (comment)](https://github.com/Flank/flank/issues/1072#issuecomment-687116654)
> 
> I think the second step should be add run these tests on different OS (Linux/Mac/Windows). WDYT?

I think that we could easily fix this by moving running integration tests to pull_request checks workflow

Sloox:
> > > @adamfilipow92 so for this project it is only running one integration test? one for ios and one for android?
> > 
> > 
> > Yes, this start point as mentioned in [#1072 (comment)](https://github.com/Flank/flank/issues/1072#issuecomment-687116654)
> > I think the second step should be add run these tests on different OS (Linux/Mac/Windows). WDYT?
> 
> I think that we could easily fix this by moving running integration tests to pull_request checks workflow(with @Sloox PR we have all platforms covered)

makes sense 👍

.github/workflows/release.yml
Sloox: This has been answered 👍

.github/workflows/release.yml
adamfilipow92: Thanks, I moved integration tests from release to ubuntu/macos workflows

.github/workflows/macos_workflow.yml
```diff
@@ -43,3 +43,21 @@ jobs:
       with:
         # Repository upload token - get it from codecov.io. Required only for private repositories
         token: ${{secrets.CODECOV_TOKEN}}
+
+    - name: Gradle clean build
+      uses: eskatos/gradle-command-action@v1
+      with:
+        gradle-executable: "./test_runner/gradlew"
+        arguments: "-p test_runner clean build"
```
piotradamczyk5, line 51: I think that there is no point to run `check` and `build` together, if you need builded flank, just get rid of `check` step

adamfilipow92: Thanks, changed

piotradamczyk5 (approved):

integration_tests/build.gradle.kts
```diff
@@ -0,0 +1,43 @@
...
+    failFast = true // fail build on any finding
+    input = files("src/main/kotlin", "src/test/kotlin")
+    config = files("../config/detekt.yml")
+    autoCorrect = true //auto format for detekt via klint
+    buildUponDefaultConfig = true // preconfigure defaults
+
+    reports {
+        html.enabled = true // observe findings in your browser with structure and code snippets
+        xml.enabled = true // checkstyle like format mainly for integrations like Jenkins
+        txt.enabled = true // similar to the console output, contains issue signature to manually edit baseline files
```
axelzuziak-gogo, line 26: Do we need this comments?

axelzuziak-gogo (approved):

PR #1111, opened 2020-09-10
feat: check if gcs path exist before run

### LINKED ISSUE #1099: Check if file exists on gcs before run

As a flank user, I want to flank fail fast if the file doesn't exist on gcs bucket
This is a follow up task for https://github.com/Flank/flank/issues/1098

https://github.com/Flank/flank/issues/1098

Flank should fail fast when file doesn't exist or in the case when file is present but the user doesn't have access to that file

piotradamczyk5:
Fixes #1099 

If the file does not exist on GCS `FlankGeneralError` is thrown

## Checklist

- [x] Unit tested

### DIFF
--- a/test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
+++ b/test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
@@ -42,20 +42,18 @@ object ArgsHelper {
     }
 
     fun assertFileExists(file: String, name: String) {
-        // flank can't detect (for now) if files exists on GCS, therefore we want to skip validation
-        // todo https://github.com/Flank/flank/issues/1099
-        if (file.startsWith(GCS_PREFIX)) return
-
-        if (!File(file).exists()) {
+        if (!file.exist())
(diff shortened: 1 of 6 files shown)


Sloox (approved):

test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
```diff
@@ -42,20 +42,18 @@ object ArgsHelper {
     }
 
     fun assertFileExists(file: String, name: String) {
-        // flank can't detect (for now) if files exists on GCS, therefore we want to skip validation
-        // todo https://github.com/Flank/flank/issues/1099
-        if (file.startsWith(GCS_PREFIX)) return
-
-        if (!File(file).exists()) {
+        if (!file.exist())
```
pawelpasterz, line 45: Again, totaly cosmetic change, IMO can improve readability a bit. How about `exists().not()` or `notExists()` ? I mean, this negation mark can be skipped by mistake. It's not a change request :)

pawelpasterz (approved):
Nice work!

PR #1126, opened 2020-09-16
feat: Allow Flank using different environment variables between test-apks

### LINKED ISSUE #1090: Support identification of which androidTest APK was run on which matrix
As an app developer, I want to identify which androidTest APK was run on which matrix so I can align `coverage.ec` files back to a specific module.

We are currently uploading multiple module's androidTest APKs to FTL via Flank.   When we collect the artifacts, we can't identify which `coverage.ec` file belongs to which module.  This is impacting us as we'd like to measure coverage on a per module basis and we need to know which coverage file belongs to which module. 

We can specify the specific file name for `coverage.ec` , but since we're using `additional-app-test-apks`, we cannot pair the two pieces of information together. 

We'd like to be able to set `environment-variables` per `additional-app-test-apk` so that we can append some unique information about which module our `coverageFile` is related to. 

We think this is the relevant file with both pieces of information - https://github.com/Flank/flank/blob/fff49e3f176c953b3998c1226ede6c6e2de82a6f/cloud_testing/kotlin_poc/src/main/kotlin/ftl/gc/GcAndroidTestMatrix.kt

We'd be open to alternatives so that we can connect the dots between an androidTest APK and a matrix run. 

**Additional context**
Some additional discussion is at https://firebase-community.slack.com/archives/C72V6UW8M/p1599155970017700

adamfilipow92:
Fixes #1090 

Doc can be found [here](docs/using_different_environment_variables_in_different_matrices.md)

1. Configure test apks with enabled coverage
2. Prepare yaml config for flank like

```
gcloud:
  test: ./src/test/kotlin/ftl/fixtures/tmp/apk/app-multiple-success-debug-androidTest.apk
  environment-variables:
    coverage: true
    coverageFile: /sdcard/main.ec
    clearPackageData: true
  directories-to-pull:
    - /sdcard/
  use-orchestrator: false

flank:
  disable-sharding: false
  max-test-shards: 2
  files-to-download:
    - .*/sdcard/[^/]+\.ec$
  additional-app-test-apks:
    - test: ./src/test/kotlin/ftl/fixtures/tmp/apk/app-multiple-success-debug-androidTest.apk
      environmentVariables:
        coverageFile: /sdcard/module_1.ec
    - test: ./src/test/kotlin/ftl/fixtures/tmp/apk/app-multiple-success-debug-androidTest.apk
      environmentVariables:
        coverageFile: /sdcard/module_2.ec
      
```

3. Run flank
4. When flank finished, you should see on the bucket coverages files with different names.

## Checklist

- [X] Documented
- [X] Unit tested

### DIFF
--- a/test_runner/src/main/kotlin/ftl/run/platform/android/AndroidTestConfig.kt
+++ b/test_runner/src/main/kotlin/ftl/run/platform/android/AndroidTestConfig.kt
@@ -4,7 +4,6 @@ import ftl.args.FlankRoboDirective
 import ftl.args.ShardChunks
 
 sealed class AndroidTestConfig {
-
     data class Instrumentation(
         val appApkGcsPath: String,
         val testApkGcsPath: String,
@@ -14,7 +13,8 @@ sealed class AndroidTestConfig {
         val disableSharding: Boolean,
         val testShards: ShardChunks,
         val numUniformShards: Int?,
-        val keepTestTargetsEmpty: Boolean
+        val keepTestTargetsEmpty: Boolean,
+        val env: Map<String, String> = emptyMap()
     ) : AndroidTestConfig()
 
     data class Robo(
@@ -22,4 +22,5 @@ sealed class AndroidTestConfig {
         val flankRoboDirectives: List<FlankRoboDirective>?,
         val roboScriptGcsPath: String?
     ) : AndroidTestConfig()
+
 }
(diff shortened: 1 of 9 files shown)


Sloox (approved):

test_runner/src/main/kotlin/ftl/run/platform/android/AndroidTestConfig.kt
```diff
@@ -4,17 +4,17 @@ import ftl.args.FlankRoboDirective
...
 
 sealed class AndroidTestConfig {
-
     data class Instrumentation(
         val appApkGcsPath: String,
         val testApkGcsPath: String,
         val testRunnerClass: String?,
         val orchestratorOption: String?,
-        // sharding
+            // sharding
```
piotradamczyk5, line 12:
```suggestion
        // sharding
```

piotradamczyk5 (approved):

PR #1129, opened 2020-09-18
build: Disable IT for iOS

pawelpasterz:
Currently, iOS has only physical devices. Integration tests are launched at every commit for PR and both combined causes the` flank-open-source` project hits the quota limit extremely fast.
With quota limit exceeded PR automerge is blocked (IT tests are failing - quota limit)
This PR disables IT for iOS.


adamfilipow92 (approved):
We need to rethink when run integration tests. I think another good place to run it, can be merge to master. WDYT? This should reduce quota usage a lot.

piotradamczyk5 (approved):

PR #1133, opened 2020-09-18
feat: Auto update dependencies

### LINKED ISSUE #848: Auto update gradle dependencies

As a Flank developer, I want to automatically update dependencies so I can develop features using the latest libraries.

---

There are probably gradle plugins or other ideas that will auto update deps. Maybe we should use dependabot from GitHub?
https://github.com/dependabot/dependabot-core/issues/2238

or use a gradle plugin?

piotradamczyk5:
Fixes #848

1. Generate report using command ` ./gradlew dependencyUpdates -DoutputFormatter=json -DoutputDir=..`
2. Build flank scripts using script ` ./flank-scripts/bash/buildFlankScripts.sh`
3. Run `./flank-scripts/bash/flankScripts dependencies update`

`Versions.kt` and `gradle-wrapper.properties` files should be updated

Github action will run each Monday on 5 AM and will create new PR with a dependency update
image

[Working example](https://github.com/piotradamczyk5/flank/pull/3)
## Checklist

- [x] Update dependencies
- [x] Update gradle
- [x] Update plugins
- [x] Create scheduled Github Action workflow
- [x] Documented
- [x] Unit tested


Sloox:
What's the difference with just using the + operator on the version number for gradle dependencies?
For Example:
```
object Versions {
    // https://github.com/bugsnag/bugsnag-java/releases
    const val BUGSNAG = "3.6.1
...
```
Could be:

```
object Versions {
    // https://github.com/bugsnag/bugsnag-java/releases
    const val BUGSNAG = "+"
```

More information here: https://docs.gradle.org/current/userguide/dynamic_versions.html

Sloox (changes requested):
What's the difference with just using the + operator on the version number for gradle dependencies?
For Example:
```
object Versions {
    // https://github.com/bugsnag/bugsnag-java/releases
    const val BUGSNAG = "3.6.1
...
```
Could be:

```
object Versions {
    // https://github.com/bugsnag/bugsnag-java/releases
    const val BUGSNAG = "+"
```

More information here: https://docs.gradle.org/current/userguide/dynamic_versions.html

bootstraponline:
> What's the difference with just using the + operator on the version number for gradle dependencies?

The reason is we want reproducible builds. Explicitly specifying our dependencies allows us to ensure the same build is produced for a given commit. This has security and debugging benefits.

> Dynamic versions add nondeterminism to your build. In other words, the same source code may not yield the same output. 
>
> https://blog.danlew.net/2015/09/09/dont-use-dynamic-versions-for-your-dependencies/

Sloox (approved):

piotradamczyk5:
@Mergifyio rebase

pawelpasterz:
I like the solution 👍 
The only thing I am wondering if we should exclude or  resolve a somehow different case with `GOOGLE_API_TOOLRESULTS`
```
-    const val GOOGLE_API_TOOLRESULTS = "v1beta3-rev20200803-1.30.10"
+    const val GOOGLE_API_TOOLRESULTS = "v1-rev20200224-1.30.8"
```
The proposed version is older than currently used and once merged will break flank build.
Being up to date with the latest beta version is rather the strength of flank and allows us to use some crucial features.

bootstraponline:
> Being up to date with the latest beta version is rather the strength of flank and allows us to use some crucial features.

I agree we should use the latest version of the API libraries. That's why for the test results, we generate the library ourself to have the most up to date version. 🙂

piotradamczyk5:
> > Being up to date with the latest beta version is rather the strength of flank and allows us to use some crucial features.
> 
> I agree we should use the latest version of the API libraries. That's why for the test results, we generate the library ourself to have the most up to date version. 🙂

It seems that some configuration of `Gradle Versions Plugin` should be done to avoid this or we should resolve it before merging

pawelpasterz:
Yeah, that was actually what I was thinking of. Is it possible to have `GOOGLE_API_TOOLRESULTS` excluded and has different rule from the rest? Sth like: "take the latest version but it needs to start with `v1beta3` "

piotradamczyk5:
> Yeah, that was actually what I was thinking of. Is it possible to have `GOOGLE_API_TOOLRESULTS` excluded and has different rule from the rest? Sth like: "take the latest version but it needs to start with `v1beta3` "

According to [documentation](https://github.com/ben-manes/gradle-versions-plugin), it is possible, I will check it

piotradamczyk5:
> > Yeah, that was actually what I was thinking of. Is it possible to have `GOOGLE_API_TOOLRESULTS` excluded and has different rule from the rest? Sth like: "take the latest version but it needs to start with `v1beta3` "
> 
> According to [documentation](https://github.com/ben-manes/gradle-versions-plugin), it is possible, I will check it

Done 🔥

pawelpasterz (approved):

PR #1138, opened 2020-09-21
fix: Fix cost report

### LINKED ISSUE #1135: CostReport.txt calculates with physical device for NexusLowRes
**Describe the bug**
I'm using device model NexusLowRes API 28, yet the CostReport.txt is:
```
CostReport
  Physical devices
    $9.25 for 1h 51m
```

Also the FetchArtifacts seem to be wrong. Here you can clearly see that the test device is "NexusLowRes" yet the cost report shows physical device.
```
FetchArtifacts
  .................................................
  Updating matrix file

CostReport
  Physical devices
    $10.00 for 2h 0m

  Uploading CostReport.txt .
MatrixResultsReport
  0 / 1 (0.00%)
  1 matrices failed

┌─────────┬──────────────────────┬────────────────────────────┬─────────────────────────────────┐
│ OUTCOME │      MATRIX ID       │      TEST AXIS VALUE       │          TEST DETAILS           │
├─────────┼──────────────────────┼────────────────────────────┼─────────────────────────────────┤
│ failure │ matrix-1x51e23wd43q3 │ NexusLowRes-28-en-portrait │ 17 test cases failed, 32 passed │
└─────────┴──────────────────────┴────────────────────────────┴─────────────────────────────────┘
More details are available at:
https://console.firebase.google.com/project/xyz-apps/testlab/histories/YY/matrices/XX

  Uploading MatrixResultsReport.txt .
  Uploading HtmlErrorReport.html .
```

**To Reproduce**
My flank.yml

```yaml
gcloud:
  test-targets:
    - class com.xyz.LoadTest
  record-video: true
  timeout: 30m
  async: false
  project: xyz-apps
  results-history-name: xyz-name
  app: ../app/build/outputs/apk/integration/debug/app-integration-debug.apk
  test: ../app/build/outputs/apk/androidTest/integration/debug/app-integration-debug-androidTest.apk
  auto-google-login: true
  use-orchestrator: true
  environment-variables:
    clearPackageData: true
  directories-to-pull:
    - /sdcard/screenshots
  performance-metrics: true
  device:
    - model: NexusLowRes
      version: 28
flank:
  # set to -1 to use one shard per test.
  testShards: -1
  repeatTests: 1
```

**Expected behavior**

According to [AndroidCatalogTest.java#L41 ](https://github.com/Flank/flank/blob/master/test_runner/src/test/kotlin/ftl/android/AndroidCatalogTest.kt#L41)this "NexusLowRes" is a virtual device.

Just like in the Android Catalog described:
image

Furthermore, in the Firebase Console on the cost break-down the costs don't match what is calculated by flank.

**Details:**

I'm also using the latest flank version
```sh
echo "Building INT app APK and test APK"
./gradlew assembleIntegrationDebug assembleIntegrationDebugAndroidTest

pushd cicd || exit

if [ ! -f flank.jar ]; then
  echo "Downloading the latest flank jar..."
  wget --quiet https://github.com/TestArmada/flank/releases/download/flank_snapshot/flank.jar -O ./flank.jar
fi

echo "Running flank"
java -jar flank.jar firebase test android run
popd || exit

exit 0
```

```sh
java -version
openjdk version "1.8.0_265"
```

pawelpasterz:
Fixes #1135 
Due to an incorrect device model check, all android models were considered as physical ones.

1) Start test run for any virtual device, observe cost report calculated for virtual device
2) Start test run for both, virtual and physical device, observe cost report consists of a section for physical and virtual devices

## Checklist

- [x] Unit tested

### DIFF
--- a/test_runner/src/main/kotlin/ftl/android/AndroidCatalog.kt
+++ b/test_runner/src/main/kotlin/ftl/android/AndroidCatalog.kt
@@ -73,14 +73,16 @@ object AndroidCatalog {
     fun isVirtualDevice(modelId: String, projectId: String): Boolean {
         val form = deviceCatalog(projectId).models
             .find { it.id.equals(modelId, ignoreCase = true) }?.form
-            ?: "PHYSICAL".also { println("Unable to find device type for $modelId. PHYSICAL used as fallback in cost calculations") }
+            ?: DeviceType.PHYSICAL.name.also {
+                println("Unable to find device type for $modelId. PHYSICAL used as fallback in cost calculations")
+            }
 
-        return form.equals("VIRTUAL", ignoreCase = true)
+        return form.equals(DeviceType.VIRTUAL.name, ignoreCase = true)
     }
+}
 
-    fun deviceType(modelId: String, projectId: String) =
-        if (isVirtualDevice(modelId, projectId)) "VIRTUAL"
-        else "PHYSICAL"
+enum class DeviceType {
+    VIRTUAL, PHYSICAL
 }
 
 sealed class DeviceConfigCheck
--- a/test_runner/src/main/kotlin/ftl/android/AndroidCatalog.kt
+++ b/test_runner/src/main/kotlin/ftl/android/AndroidCatalog.kt
@@ -71,7 +71,10 @@ object AndroidCatalog {
         ?: false
 
     fun isVirtualDevice(modelId: String, projectId: String): Boolean {
-        val form = deviceCatalog(projectId).models.find { it.id.equals(modelId, ignoreCase = true) }?.form ?: "PHYSICAL"
+        val form = deviceCatalog(projectId).models
+            .find { it.id.equals(modelId, ignoreCase = true) }?.form
+            ?: "PHYSICAL".also { println("Unable to find device type for $modelId. PHYSICAL used as fallback in cost calculations") }
+
         return form.equals("VIRTUAL", ignoreCase = true)
     }
 
--- a/test_runner/src/main/kotlin/ftl/android/AndroidCatalog.kt
+++ b/test_runner/src/main/kotlin/ftl/android/AndroidCatalog.kt
@@ -65,12 +65,19 @@ object AndroidCatalog {
         return SupportedDeviceConfig
     }
 
-    fun isVirtualDevice(device: AndroidDevice?, projectId: String): Boolean = device?.androidModelId?.let { isVirtualDevice(it, projectId) } ?: false
+    fun isVirtualDevice(device: AndroidDevice?, projectId: String): Boolean = device
+        ?.androidModelId
+        ?.let { isVirtualDevice(it, projectId) }
+        ?: false
 
     fun isVirtualDevice(modelId: String, projectId: String): Boolean {
-        val form = deviceCatalog(projectId).models.find { it.id == modelId }?.form ?: "PHYSICAL"
-        return form == "VIRTUAL"
+        val form = deviceCatalog(projectId).models.find { it.id.equals(modelId, ignoreCase = true) }?.form ?: "PHYSICAL"
+        return form.equals("VIRTUAL", ignoreCase = true)
     }
+
+    fun deviceType(modelId: String, projectId: String) =
+        if (isVirtualDevice(modelId, projectId)) "VIRTUAL"
+        else "PHYSICAL"
 }
 
 sealed class DeviceConfigCheck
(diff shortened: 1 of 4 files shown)


test_runner/src/main/kotlin/ftl/android/AndroidCatalog.kt
```diff
@@ -65,12 +65,22 @@ object AndroidCatalog {
...
+        ?.androidModelId
+        ?.let { isVirtualDevice(it, projectId) }
+        ?: false
 
     fun isVirtualDevice(modelId: String, projectId: String): Boolean {
-        val form = deviceCatalog(projectId).models.find { it.id == modelId }?.form ?: "PHYSICAL"
-        return form == "VIRTUAL"
+        val form = deviceCatalog(projectId).models
+            .find { it.id.equals(modelId, ignoreCase = true) }?.form
+            ?: "PHYSICAL".also { println("Unable to find device type for $modelId. PHYSICAL used as fallback in cost calculations") }
```
Sloox, line 76: Use the DeviceType enum that was created below. `DeviceType.Physical.GetName` should do.

Sloox: There are a few places around this that the enum can be used. I wont single out everyone as it is spam. But i think you get the idea 👍

test_runner/src/main/kotlin/ftl/reports/outcome/BillableMinutes.kt
```diff
@@ -10,33 +11,31 @@ data class BillableMinutes(
...
-    timeoutValue: Long
-): BillableMinutes =
-    groupByDeviceType(projectId).run {
+fun List<Step>.calculateAndroidBillableMinutes(projectId: String, timeoutValue: Long): BillableMinutes = this
+    .groupBy { deviceType(it.deviceModel(), projectId) }
+    .run {
         BillableMinutes(
-            virtual = get(true)?.sumBillableMinutes(timeoutValue) ?: 0,
-            physical = get(false)?.sumBillableMinutes(timeoutValue) ?: 0
+            virtual = get(DeviceType.VIRTUAL)?.sumBillableMinutes(timeoutValue) ?: 0,
```
Sloox, line 18: What do you think of introducing an extension function for cases like this eg 
virtual = get(DeviceType.VIRTUAL)?.sumBillableMinutes(timeoutValue) ?: 0,
```suggestion
            virtual = get(DeviceType.VIRTUAL)?.sumBillableMinutes(timeoutValue).orZero(),
```
We already make use of it fairly extensively in other ways such as `orUnknown` `orNull` etc

test_runner/src/main/kotlin/ftl/android/AndroidCatalog.kt
pawelpasterz: 👍  Good idea, I don't like depending on strings

test_runner/src/main/kotlin/ftl/reports/outcome/BillableMinutes.kt
pawelpasterz: I think it won't work with this particular case. `sumBillableMinutes` never returns `null` so we would need to move `null` through rewritten `sumBillableMinutes` sth like
```kotlin
private fun List<Step>?.sumBillableMinutes(timeout: Long) = this?.run {
    mapNotNull { it.getBillableSeconds(default = timeout) }
    .map { billableMinutes(it) }
    .sum()
}
```
which is now having additional logic (null check)
Then it should (IMO) be renamed to `sumBillableMinutesOrNull`

I like the idea but I am not convinced if this is good place to implement it.
On the other hand this would follow the convention that we have (as you mentioned) in the project.
I don't have strong opinion about it, let me know what do you think.

test_runner/src/main/kotlin/ftl/reports/outcome/BillableMinutes.kt
```diff
@@ -10,33 +11,31 @@ data class BillableMinutes(
...
+    .sum()
 
 private fun Step.getBillableSeconds(default: Long) =
     testExecutionStep?.testTiming?.testProcessDuration?.seconds?.let {
         min(it, default)
     }
+
+private enum class DeviceType {
+    VIRTUAL, PHYSICAL
+}
```
piotradamczyk5, line 37: duplicate enum with `AndroidCatalog` ?

pawelpasterz: Thanks!

piotradamczyk5 (approved):

jan-gogo (approved):

PR #1165, opened 2020-09-25
style: commit files corrected by Detekt

### LINKED ISSUE #1146: Review and unify detekt settings
Detekt seems to be missing some settings. Some problems were spotted like unused imports, exceeded line lenght.
Configuration should be review.

As an addition to this issue we can unify detekt configuration for all submodules

piotradamczyk5:
Fixes #1146 

This PR fixes files reported by Detekt, update its configuration, and make GitHub actions push files changed by Detekt during auto-correct

Detekt does not report any issues.
Files auto fixed by Detekt are committed to the source branch

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -21,6 +21,12 @@ repositories {
     mavenCentral()
 }
 
+tasks.named("detekt") {
+    dependsOn(":test_runner:detekt")
+    dependsOn(":integration_tests:detekt")
+    dependsOn(":flank-scripts:detekt")
+}
+
 subprojects {
     afterEvaluate {
         if (tasks.findByName("detekt") != null) {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -21,12 +21,6 @@ repositories {
     mavenCentral()
 }
 
-tasks.named("detekt") {
-    dependsOn(":test_runner:detekt")
-    dependsOn(":integration_tests:detekt")
-    dependsOn(":flank-scripts:detekt")
-}
-
 subprojects {
     afterEvaluate {
         if (tasks.findByName("detekt") != null) {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -12,12 +12,19 @@ configurations.all {
 
 plugins {
     kotlin(Plugins.Kotlin.PLUGIN_JVM) version Versions.KOTLIN
+    id(Plugins.DETEKT_PLUGIN) version Versions.DETEKT
     id(Plugins.BEN_MANES_PLUGIN) version Versions.BEN_MANES
 }
 
 repositories {
     jcenter()
     mavenCentral()
+}
+
+subprojects {
+    tasks.withType<Test> {
+        maxParallelForks = Runtime.getRuntime().availableProcessors() / 2
+    }
 }
 
 tasks.named("dependencyUpdates", DependencyUpdatesTask::class.java).configure {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -16,20 +16,19 @@ plugins {
     id(Plugins.BEN_MANES_PLUGIN) version Versions.BEN_MANES
 }
 
-tasks.named("detekt") {
-    dependsOn(":test_runner:detekt")
-    dependsOn(":integration_tests:detekt")
-    dependsOn(":flank-scripts:detekt")
-}
-
-tasks.named("clean"){
-    dependsOn(":test_runner:clean")
-    dependsOn(":integration_tests:clean")
-}
-
 repositories {
     jcenter()
     mavenCentral()
+}
+
+subprojects {
+    afterEvaluate {
+        if (tasks.findByName("detekt") != null) {
+            tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile> {
+                dependsOn(tasks["detekt"])
+            }
+        }
+    }
 }
 
 tasks.named("dependencyUpdates", DependencyUpdatesTask::class.java).configure {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -32,12 +32,6 @@ repositories {
     mavenCentral()
 }
 
-subprojects {
-    tasks.withType<Test> {
-        maxParallelForks = Runtime.getRuntime().availableProcessors()
-    }
-}
-
 tasks.named("dependencyUpdates", DependencyUpdatesTask::class.java).configure {
 
     gradleReleaseChannel = "release-candidate"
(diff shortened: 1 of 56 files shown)


adamfilipow92 (approved):
👍

build.gradle.kts
```diff
@@ -12,9 +12,21 @@ configurations.all {
...
     id(Plugins.BEN_MANES_PLUGIN) version Versions.BEN_MANES
 }
 
+tasks.named("detekt") {
+    dependsOn(":test_runner:detekt")
+    dependsOn(":integration_tests:detekt")
+    dependsOn(":flank-scripts:detekt")
+}
+
+tasks.named("clean"){
```
pawelpasterz, line 25: Is it necessary? This task already invokes those:

Screenshot 2020-09-30 at 09 09 18

Or is it the order that matters?

build.gradle.kts
```diff
@@ -12,9 +12,21 @@ configurations.all {
 
 plugins {
     kotlin(Plugins.Kotlin.PLUGIN_JVM) version Versions.KOTLIN
+    id(Plugins.DETEKT_PLUGIN) version Versions.DETEKT
     id(Plugins.BEN_MANES_PLUGIN) version Versions.BEN_MANES
 }
 
+tasks.named("detekt") {
```
pawelpasterz, line 19: Same here

config/detekt.yml
```diff
@@ -106,17 +106,19 @@ formatting:
...
   android: false
   autoCorrect: true
+  ArgumentListWrapping:
+    active: false
   AnnotationOnSeparateLine:
     active: false
     autoCorrect: true
   ChainWrapping:
-    active: true
+    active: false
```
pawelpasterz, line 115: Why disabled?

pawelpasterz: And I think this is not consistent with the official style guide
https://kotlinlang.org/docs/reference/coding-conventions.html#chained-call-wrapping

config/detekt.yml
```diff
@@ -106,17 +106,19 @@ formatting:
...
     autoCorrect: true
   CommentSpacing:
     active: true
     autoCorrect: true
-  EnumEntryNameCase: 
-    active: true
+  EnumEntryNameCase:
+    active: false
```
pawelpasterz, line 121: Isn't the general convention?
https://kotlinlang.org/docs/reference/coding-conventions.html#enumerations

pawelpasterz (commented):
Just a couple of questions

integration_tests/build.gradle.kts
```diff
@@ -1,7 +1,7 @@
 plugins {
     java
     kotlin(Plugins.Kotlin.PLUGIN_JVM)
-    id(Plugins.DETEKT_PLUGIN) version Versions.DETEKT
+    id(Plugins.DETEKT_PLUGIN)
 }
```
pawelpasterz, line 6:
```suggestion
tasks.withType<KotlinCompile> {
    dependsOn(tasks["detekt"])
}
```

That would run detekt task before compilation which could be more efficient when problems were found

test_runner/build.gradle.kts
```diff
@@ -132,12 +132,18 @@ publishing {
     }
 }
 
+subprojects {
```
pawelpasterz, line 135: `test_runner` doesn't have any subprojects (or am I wrong?) so `subprojects` is redundant here.
But I think we could move it to the root build gradle file

pawelpasterz: Ah... I think `test_runner:test` task can't use parallelism. We have failed tests then.
Anyway, IMHO it would be worth to have global config with parallel forks and explicit single fork configuration for `test_runner` project (with comment why). We can also create an issue to make it possible to test in parallel

build.gradle.kts
```diff
@@ -12,9 +12,21 @@ configurations.all {
...
+    dependsOn(":test_runner:detekt")
+    dependsOn(":integration_tests:detekt")
+    dependsOn(":flank-scripts:detekt")
+}
+
+tasks.named("clean"){
+    dependsOn(":test_runner:clean")
+    dependsOn(":integration_tests:clean")
+}
+
```
pawelpasterz, line 29:
```suggestion
subprojects {
    afterEvaluate {
        if (tasks.findByName("detekt") != null) {
            tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile> {
                dependsOn(tasks["detekt"])
            }
        }
    }
}
```

That would configure detekt task to run before every kotlin compilation (for all subprojects)

build.gradle.kts
Sloox: the clean was for testing will remove

build.gradle.kts
Sloox: this one is used.

pawelpasterz: Understand, yet I think it's redundant
Screenshot 2020-09-30 at 11 28 34

test_runner/build.gradle.kts
piotradamczyk5: Hey Pawel it was here before, it is just moved upper.
If we move it to root `build.gradle.kts` then it makes that tests fail as you mention, so we could remove it from here

integration_tests/build.gradle.kts
Sloox: skipping

test_runner/build.gradle.kts
pawelpasterz: Sounds good to me 👍

build.gradle.kts
Sloox: hmmm will test to see if it is. We did make some changes and maybe it is now redundant. Without it the tasks didnt run at all

Sloox: awesome, after testing they now all fire correctly its been removed.

build.gradle.kts
Sloox: I think i will apply it here as you mentioned. It will force us to work with the code immediately if its picked up with detekt.

config/detekt.yml
Sloox: actually will re-enable

config/detekt.yml
Sloox: It was conflicting with IDE settings. but will re-enable

pawelpasterz: Oh, that was not my intention to revert your change, rather discuss why did you do so :)

test_runner/build.gradle.kts
Sloox: lets remove it for now and think on it later. It can greatly speedup the build but it has some issues like the failed tests

pawelpasterz: Are you sure? I think it has no effect if placed here

Sloox: I edited my comment ;)

pawelpasterz (approved):

PR #1186, opened 2020-10-02
feat: Grant permissions addition

### LINKED ISSUE #1182: Add --grant-permissions support

As an Android developer, I want to control if the app is granted permission so I can test my app completely.

Add support for the new grant permissions flag
https://cloud.google.com/sdk/gcloud/reference/alpha/firebase/test/android/run#--grant-permissions

Review the list of flags on gcloud alpha to see if there's any other new updates that we're missing in Flank.

Sloox:
Added grant permissions to android args

Fixes #1182 
.
## Checklist

- [ ] Documented
- [ ] Unit tested


piotradamczyk5:
Please add documentation about `grant-permissions` to example `.yaml` (`test_runner/flank.yaml`) and for root `README.md`

Sloox:
> Please add documentation about `grant-permissions` to example `.yaml` (`test_runner/flank.yaml`) and for root `README.md`

done

test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
```diff
@@ -45,8 +45,7 @@ object ArgsHelper {
     }
 
     fun assertFileExists(file: String, name: String) {
-        if (!file.exist())
-            throw FlankGeneralError("'$file' $name doesn't exist")
+        if (!file.exist()) throw FlankGeneralError("'$file' $name doesn't exist")
```
axelzuziak-gogo, line 48: 👍

axelzuziak-gogo (approved):

README.md
```diff
@@ -324,6 +324,11 @@ gcloud:
   # directories-to-pull:
   #   - /sdcard/
 
+  ## Whether to grant runtime permissions on the device before the test begins.
+  ## By default, all permissions are granted. PERMISSIONS must be one of: all, none
+  # grant-permissions:
+  # all or none
```
piotradamczyk5, line 330: single line

piotradamczyk5:
```suggestion
  # grant-permissions: all or none
```

test_runner/flank.yml
```diff
@@ -82,6 +82,11 @@ gcloud:
   # directories-to-pull:
   #   - /sdcard/
 
+  ## Whether to grant runtime permissions on the device before the test begins.
+  ## By default, all permissions are granted. PERMISSIONS must be one of: all, none
+  # grant-permissions:
+  # all or none
```
piotradamczyk5, line 88: same here should be 
```yaml
  # grant-permissions: all or none
```

piotradamczyk5 (approved):

test_runner/src/main/kotlin/ftl/config/android/AndroidGcloudConfig.kt
```diff
@@ -210,6 +218,7 @@ data class AndroidGcloudConfig @JsonIgnore constructor(
             autoGoogleLogin = FlankDefaults.DISABLE_AUTO_LOGIN
             useOrchestrator = true
             environmentVariables = emptyMap()
+            grantPermissions = null
```
pawelpasterz, line 221: I think we should not set `null` here. It will result in printed output `grant-permissions: null` which might be confusing for users. Option description has info about `all` default value.

pawelpasterz (changes requested):
I think your implementation is missing matrix setup with `setDontAutograntPermissions()` flag
`GcAndroidTestMatrix`
```kotlin
val testSetup = TestSetup()
            .setAccount(account)
            .setNetworkProfile(args.networkProfile)
            .setDirectoriesToPull(args.directoriesToPull)
            .setAdditionalApks(additionalApkGcsPaths.mapGcsPathsToApks())
            .setFilesToPush(otherFiles.mapToDeviceFiles())
            .setEnvironmentVariables(args, androidTestConfig)
            .setDontAutograntPermissions(true/false)            <=======
```
And tests for this as well :)

test_runner/src/main/kotlin/ftl/config/android/AndroidGcloudConfig.kt
Sloox: so you assume it should be set to ALL at this point?

Sloox: defaulted

Sloox:
> I think your implementation is missing matrix setup with `setDontAutograntPermissions()` flag
> `GcAndroidTestMatrix`
> 
> ```kotlin
> val testSetup = TestSetup()
>             .setAccount(account)
>             .setNetworkProfile(args.networkProfile)
>             .setDirectoriesToPull(args.directoriesToPull)
>             .setAdditionalApks(additionalApkGcsPaths.mapGcsPathsToApks())
>             .setFilesToPush(otherFiles.mapToDeviceFiles())
>             .setEnvironmentVariables(args, androidTestConfig)
>             .setDontAutograntPermissions(true/false)            <=======
> ```
> 
> * tests

added.

test_runner/src/main/kotlin/ftl/config/Load.kt
```diff
@@ -15,10 +15,8 @@ import ftl.util.loadFile
 import java.io.Reader
 import java.nio.file.Path
 
-fun loadAndroidConfig(
-    path: Path? = null,
-    reader: Reader = loadFile(path!!)
-): AndroidConfig = modifyAndThrow(reader, android = true).run {
+fun loadAndroidConfig(path: Path? = null, reader: Reader = loadFile(path!!)): AndroidConfig =
```
jan-gogo, line 18: Keeping arguments in lines is more readable & scalable IMO

test_runner/src/main/kotlin/ftl/config/android/AndroidGcloudConfig.kt
pawelpasterz: I think so, I would be surprised seeing `null` (but it's only my subjective opinion)

test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
```diff
@@ -13,6 +13,7 @@ data class AndroidArgs(
     val roboScript: String?,
     val environmentVariables: Map<String, String>, // should not be printed, becuase could contains sensitive informations
     val directoriesToPull: List<String>,
+    val grantPermissions: String?,
```
jan-gogo, line 16: I guess that this property should be used somewhere, but it is not or I cannot find.

test_runner/src/main/kotlin/ftl/gc/GcAndroidTestMatrix.kt
```diff
@@ -57,6 +57,7 @@ object GcAndroidTestMatrix {
             .setAdditionalApks(additionalApkGcsPaths.mapGcsPathsToApks())
             .setFilesToPush(otherFiles.mapToDeviceFiles())
             .setEnvironmentVariables(args, androidTestConfig)
+            .setDontAutograntPermissions(true)
```
pawelpasterz, line 60: It should be depending on `grantPermissions` option: `all => false` and `none => true`

https://github.com/Flank/gcloud_cli/blob/master/google-cloud-sdk/lib/googlecloudsdk/api_lib/firebase/test/android/matrix_creator.py#L211

pawelpasterz (changes requested):

test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
Sloox: it will be used below. with @pawelpasterz comment

test_runner/src/main/kotlin/ftl/config/Load.kt
Sloox: reverted 👍

test_runner/src/main/kotlin/ftl/gc/GcAndroidTestMatrix.kt
Sloox: ahh ok ok I understand. and done

test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
```diff
@@ -79,6 +81,8 @@ AndroidArgs
     }
 }
 
+val AndroidArgs.isShouldAutoGrantPermissions
+    get() = grantPermissions.isNotNull() && (grantPermissions.equals("all"))
```
pawelpasterz, line 85: How about renaming it to match gcloud's `dontAutograntPermissions` (don't need to be exactly the same name though). Also, I think we could use an enum instead of string values.

```suggestion
    get() = grantPermissions.isNotNull() && grantPermissions.equals("none")
```

test_runner/src/main/kotlin/ftl/gc/GcAndroidTestMatrix.kt
```diff
@@ -57,6 +58,7 @@ object GcAndroidTestMatrix {
             .setAdditionalApks(additionalApkGcsPaths.mapGcsPathsToApks())
             .setFilesToPush(otherFiles.mapToDeviceFiles())
             .setEnvironmentVariables(args, androidTestConfig)
+            .setDontAutograntPermissions(!args.isShouldAutoGrantPermissions)
```
pawelpasterz, line 61: Changing the name to something similar to gcloud's would allow us to get rid of `!`

test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
Sloox: for two enums I am not convinced? why do you think so?

jan-gogo (approved):

PR #1187, opened 2020-10-02
test: Remove verification test for edge cases of incorrect outcome in flaky tests

### LINKED ISSUE #918: Incorrect outcome for flaky tests [Edge Case]
**Describe the bug**

Sometimes flank's outcome is `success` when there are `flaky` tests

**To Reproduce**

So far no reliable way to reproduce it (with 10/10 rate) found. One that gave the same result is to run a test annotated with `@Ignore` in `SavedMatrixTest`

**Expected behavior**

If there is at least one `flaky` test outcome should be `flaky` as well. The ignored test mentioned earlier must pass.

>

Flank v20.07.0

**Additional context**

It is suspected that in the case of `summary=null` for both, step and execution, flank is unable to deduce outcome correctly.

This description should be updated in case of any new information.

adamfilipow92:
Fixes #918 

Ignored test should be removed from SavedMatrixTest

### DIFF
--- a/test_runner/src/test/kotlin/ftl/json/SavedMatrixTest.kt
+++ b/test_runner/src/test/kotlin/ftl/json/SavedMatrixTest.kt
@@ -236,33 +236,4 @@ class SavedMatrixTest {
             savedMatrix.outcome
         )
     }
-
-    @Ignore("Should be used to verify https://github.com/Flank/flank/issues/918 fix")
-    @Test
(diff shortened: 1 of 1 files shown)


pawelpasterz (approved):

piotradamczyk5 (approved):

piotradamczyk5:
I changed a title, cause this is a change for tests only

adamfilipow92:
> I changed a title, cause this is a change for tests only

thanks!

PR #1214, opened 2020-10-05
feat: Added printing missing options

### LINKED ISSUE #1185: Some config options are not printed during test run
There are some configuration options that are not printed with flank run:
1. `disable-results-upload`
2. ~`default-test-time`~
3. `default-class-test-time`
4. (can be more, needs to be verified)

Are those not printed intentionally? 
  - **No**  we should verify the remaining options and add to the printing logic
  - **Yes** then we should decide how to hide them? Even though not printed it is still possible to set them via yml file

piotradamczyk5:
Fixes #1185

All options are printed during Flank run

## Checklist

- [x] Documented
- [x] Unit tested
- [x] Ensure that all options are printed

### DIFF
--- a/test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
+++ b/test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
@@ -73,6 +73,8 @@ AndroidArgs
       legacy-junit-result: $useLegacyJUnitResult
       ignore-failed-tests: $ignoreFailedTests
       output-style: ${outputStyle.name.toLowerCase()}
+      disable-results-upload: $disableResultsUpload
+      default-class-test-time: $defaultClassTestTime
    """.trimIndent()
     }
 }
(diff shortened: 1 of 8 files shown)


pawelpasterz (approved):

Sloox (approved):

test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
```diff
@@ -11,7 +11,7 @@ data class AndroidArgs(
     val useOrchestrator: Boolean,
     val roboDirectives: List<FlankRoboDirective>,
     val roboScript: String?,
-    val environmentVariables: Map<String, String>,
+    val environmentVariables: Map<String, String>, // should not be printed, becuase could contains sensitive informations
```
adamfilipow92, line 14: 👍

adamfilipow92 (approved):

Issue #1205, opened 2020-10-05
Android | Add --scenario-numbers


bootstraponline:
**Author the user story for this feature**

As an Android developer, I want to use --scenario-numbers so I can run tests like gcloud.

**Describe the solution**

> --scenario-numbers=int,[int,…]
>
> A list of game-loop scenario numbers which will be run as part of the test (default: all scenarios). A maximum of 1024 scenarios may be specified in one test matrix, but the maximum number may also be limited by the overall test --timeout setting.
>
> https://cloud.google.com/sdk/gcloud/reference/alpha/firebase/test/android/run#--scenario-numbers

Issue #1204, opened 2020-10-05
Android | Add --scenario-labels


bootstraponline:
**Author the user story for this feature**

As an Android developer, I want to use --scenario-labels so I can run tests like gcloud.

**Describe the solution**

> --scenario-labels=LABEL,[LABEL,…]
>
> A list of game-loop scenario labels (default: None). Each game-loop scenario may be labeled in the APK manifest file with one or more arbitrary strings, creating logical groupings (e.g. GPU_COMPATIBILITY_TESTS). If --scenario-numbers and --scenario-labels are specified together, Firebase Test Lab will first execute each scenario from --scenario-numbers. It will then expand each given scenario label into a list of scenario numbers marked with that label, and execute those scenarios.
>
> https://cloud.google.com/sdk/gcloud/reference/alpha/firebase/test/android/run#--scenario-labels

Issue #1203, opened 2020-10-05
Android | Add --test-targets-for-shard


bootstraponline:
**Author the user story for this feature**

As an Android developer, I want to use --test-targets-for-shard so I can run tests like gcloud.

**Describe the solution**

> --test-targets-for-shard=TEST_TARGETS_FOR_SHARD
>
> Specifies a group of packages, classes, and/or test cases to run in each shard (a group of test cases). The shards are run in parallel on separate devices. You can repeat this flag up to 50 times to specify multiple shards when one or more physical devices are selected, or up to 250 times when no physical devices are selected.
>
> https://cloud.google.com/sdk/gcloud/reference/alpha/firebase/test/android/run#--test-targets-for-shard

pawelpasterz:
Flank has `test-targets-always-run`, as I see it its behavior is the same as `--test-targets-for-shard `, or isn't it?

Issue #1202, opened 2020-10-05
Android | Add --obb-files


bootstraponline:
**Author the user story for this feature**

As an Android developer, I want to upload OBB files so I can run tests successfully.

**Describe the solution**

> --obb-files=OBB_FILE,[OBB_FILE]
>
> A list of one or two Android OBB file names which will be copied to each test device before the tests will run (default: None). Each OBB file name must conform to the format as specified by Android (e.g. [main|patch].0300110.com.example.android.obb) and will be installed into <shared-storage>/Android/obb/<package-name>/ on the test device.
>
> https://cloud.google.com/sdk/gcloud/reference/alpha/firebase/test/android/run#--obb-files

PR #1230, opened 2020-10-12
feat: Add type

### LINKED ISSUE #1201: Android | Add --type

As an Android developer, I want to specify the type of test to run so I can run tests in the same way as gcloud.

**Describe the solution**

> --type=TYPE
> The type of test to run. TYPE must be one of: instrumentation, robo, game-loop.
> https://cloud.google.com/sdk/gcloud/reference/alpha/firebase/test/android/run#--type

Sloox:
Fixes #1201 

Adds type
## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/test_runner/src/main/kotlin/ftl/args/yml/Type.kt
+++ b/test_runner/src/main/kotlin/ftl/args/yml/Type.kt
@@ -9,7 +9,7 @@ enum class Type(val ymlName: String) {
         fun fromString(stringVal: String): Type {
             val filtered = values().filter { it.ymlName == stringVal }
             if (filtered.isEmpty()) {
-                throw FlankGeneralError("Unsupported Type given `$stringVal` only [${values().joinToString(",")}] supported.")
+                throw FlankGeneralError("Unsupported Type given `$stringVal` only [${values().joinToString(","){it.ymlName}}] supported.")
             }
             return filtered.first()
         }
--- a/test_runner/src/main/kotlin/ftl/args/yml/Type.kt
+++ b/test_runner/src/main/kotlin/ftl/args/yml/Type.kt
@@ -1,5 +1,17 @@
 package ftl.args.yml
 
+import ftl.run.exception.FlankGeneralError
+
 enum class Type(val ymlName: String) {
-    INSTURMENTATION("instrumentation"), ROBO("robo"), GAMELOOP("game-loop")
+    INSTRUMENTATION("instrumentation"), ROBO("robo"), GAMELOOP("game-loop");
+
+    companion object {
+        fun fromString(stringVal: String): Type {
+            val filtered = values().filter { it.ymlName == stringVal }
+            if (filtered.isEmpty()) {
+                throw FlankGeneralError("Unsupported Type given `$stringVal` only [${values().joinToString(",")}] supported.")
+            }
+            return filtered.first()
+        }
+    }
 }
--- a/test_runner/src/main/kotlin/ftl/args/yml/Type.kt
+++ b/test_runner/src/main/kotlin/ftl/args/yml/Type.kt
@@ -1,0 +1,5 @@
+package ftl.args.yml
+
+enum class Type(val ymlName: String) {
+    INSTURMENTATION("instrumentation"), ROBO("robo"), GAMELOOP("game-loop")
+}
--- a/test_runner/src/test/kotlin/ftl/args/AndroidArgsTest.kt
+++ b/test_runner/src/test/kotlin/ftl/args/AndroidArgsTest.kt
@@ -5,6 +5,7 @@ import com.google.common.truth.Truth.assertThat
 import ftl.args.IArgs.Companion.AVAILABLE_PHYSICAL_SHARD_COUNT_RANGE
 import ftl.args.IArgs.Companion.AVAILABLE_VIRTUAL_SHARD_COUNT_RANGE
 import ftl.args.yml.AppTestPair
+import ftl.args.yml.Type
 import ftl.cli.firebase.test.android.AndroidRunCommand
 import ftl.config.Device
 import ftl.config.FtlConstants.defaultAndroidModel
@@ -98,7 +99,7 @@ class AndroidArgsTest {
           - /sdcard/screenshots
           - /sdcard/screenshots2
           grant-permissions: all
-          type: null
+          type: instrumentation
           other-files:
             /sdcard/dir1/file1.txt: $appApk
             /sdcard/dir2/file2.jpg: $testApk
@@ -239,6 +240,7 @@ class AndroidArgsTest {
             assert(useOrchestrator, false)
             assert(environmentVariables, linkedMapOf("clearPackageData" to "true", "randomEnvVar" to "false"))
             assert(directoriesToPull, listOf("/sdcard/screenshots", "/sdcard/screenshots2"))
+            assert(grantPermissions, "all")
             assert(
                 otherFiles,
                 mapOf(
@@ -246,6 +248,7 @@ class AndroidArgsTest {
                     "/sdcard/dir2/file2.jpg" to testApkAbsolutePath
                 )
             )
+            assert(type, Type.INSTRUMENTATION)
             assert(performanceMetrics, false)
             assert(testRunnerClass, "com.foo.TestRunner")
             assert(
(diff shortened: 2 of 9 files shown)


axelzuziak-gogo (approved):
Looks good

test_runner/src/test/kotlin/ftl/args/AndroidArgsTest.kt
```diff
@@ -98,6 +98,7 @@ class AndroidArgsTest {
           - /sdcard/screenshots
           - /sdcard/screenshots2
           grant-permissions: all
+          type: null
```
piotradamczyk5, line 101: I think that `androidNonDefault` should contain non-default value of property

piotradamczyk5 (approved):

pawelpasterz (changes requested):
I think the same option is for iOS.
https://cloud.google.com/sdk/gcloud/reference/alpha/firebase/test/ios/run#--type

I think implementing this feature requires more work than it seems at first glance.
Please, take a look at gcloud's implementation
https://github.com/Flank/gcloud_cli/blob/master/google-cloud-sdk/lib/googlecloudsdk/api_lib/firebase/test/android/matrix_creator.py#L244-L255
https://github.com/Flank/gcloud_cli/blob/e58409d11d3079338a2f01ef70d8e1fb283b3977/google-cloud-sdk/lib/googlecloudsdk/api_lib/firebase/test/android/matrix_creator.py#L311

I think some updates are required https://github.com/Flank/flank/blob/master/test_runner/src/main/kotlin/ftl/gc/android/SetupAndroidTest.kt (implement case for `androidLoopTest`
Analogous changes for iOS.

We should also consider some validation/fail-fast strategy for missing properties
```python
def _BuildAndroidGameLoopTestSpec(self):
    """Build a TestSpecification for an AndroidTestLoop."""
    spec = self._BuildGenericTestSpec()
    app_apk, app_bundle = self._BuildAppReference(self._args.app)
    spec.androidTestLoop = self._messages.AndroidTestLoop(
        appApk=app_apk,
        appBundle=app_bundle,
        appPackageId=self._args.app_package)
    if self._args.scenario_numbers:
      spec.androidTestLoop.scenarios = self._args.scenario_numbers
    if self._args.scenario_labels:
      spec.androidTestLoop.scenarioLabels = self._args.scenario_labels
    return spec
```

Sloox:
@pawelpasterz 
Looking into what you linked and how the test is run, I have added validation and fail fast so that you cannot run the without missing parameters.
However a close look into the python code we see that our validation already handles alot of the cases.
We therefore can simplify it down to ensuring the appApk is not null and taking into account the testApk and testRunnerClass.
The rest of the arguments are actually optional as can be seen with cases for each test, robo, game-loop and instrumentation.


```
def _BuildAndroidInstrumentationTestSpec(self):
    """Build a TestSpecification for an AndroidInstrumentationTest."""
    spec = self._BuildGenericTestSpec()
    app_apk, app_bundle = self._BuildAppReference(self._args.app)
    spec.androidInstrumentationTest = self._messages.AndroidInstrumentationTest(
        appApk=app_apk,
        appBundle=app_bundle,
        testApk=self._BuildFileReference(os.path.basename(self._args.test)),
        appPackageId=self._args.app_package,
        testPackageId=self._args.test_package,
        testRunnerClass=self._args.test_runner_class,
        testTargets=(self._args.test_targets or []),
        orchestratorOption=self._GetOrchestratorOption(),
        shardingOption=self._BuildShardingOption())
    return spec

  def _BuildAndroidRoboTestSpec(self):
    """Build a TestSpecification for an AndroidRoboTest."""
    spec = self._BuildGenericTestSpec()
    app_apk, app_bundle = self._BuildAppReference(self._args.app)
    spec.androidRoboTest = self._messages.AndroidRoboTest(
        appApk=app_apk,
        appBundle=app_bundle,
        appPackageId=self._args.app_package,
        roboDirectives=self._BuildRoboDirectives(self._args.robo_directives))
    if getattr(self._args, 'robo_script', None):
      spec.androidRoboTest.roboScript = self._BuildFileReference(
          os.path.basename(self._args.robo_script))
    return spec

  def _BuildAndroidGameLoopTestSpec(self):
    """Build a TestSpecification for an AndroidTestLoop."""
    spec = self._BuildGenericTestSpec()
    app_apk, app_bundle = self._BuildAppReference(self._args.app)
    spec.androidTestLoop = self._messages.AndroidTestLoop(
        appApk=app_apk,
        appBundle=app_bundle,
        appPackageId=self._args.app_package)
    if self._args.scenario_numbers:
      spec.androidTestLoop.scenarios = self._args.scenario_numbers
    if self._args.scenario_labels:
      spec.androidTestLoop.scenarioLabels = self._args.scenario_labels
    return spec

```

What we can see above that is common is:
appApk --> added as a fail fast
appBundle --> not configurable by us but is auto created as far as i am aware (if it is not we will need to add)
appPackageID --> same as above
testPackageID --> same as above

The rest are have been added as fail fast.
WDYT?

Sloox:
@pawelpasterz also take into account there are tasks that will be adding scenario labels and scenario numbers in the future

pawelpasterz:
@Sloox Agreed, let me take a quick, second look

test_runner/src/main/kotlin/ftl/args/yml/Type.kt
```diff
@@ -0,0 +1,17 @@
...
+import ftl.run.exception.FlankGeneralError
+
+enum class Type(val ymlName: String) {
+    INSTRUMENTATION("instrumentation"), ROBO("robo"), GAMELOOP("game-loop");
+
+    companion object {
+        fun fromString(stringVal: String): Type {
+            val filtered = values().filter { it.ymlName == stringVal }
+            if (filtered.isEmpty()) {
+                throw FlankGeneralError("Unsupported Type given `$stringVal` only [${values().joinToString(",")}] supported.")
```
pawelpasterz, line 12:
```suggestion
                throw FlankGeneralError("Unsupported Type given `$stringVal` only [${values().joinToString(", ") { it.ymlName }}] supported.")
```

I think it would be better to display yaml names
```
Unsupported Type given `instrumentationasda` only [instrumentation, robo, game-loop] supported.
```
instead of
```
Unsupported Type given `instrumentationasda` only [INSTRUMENTATION,ROBO,GAMELOOP] supported.
```

pawelpasterz (changes requested):

pawelpasterz:
Also with config:
```yml
gcloud:
  app: [path]/app-debug.apk
  use-orchestrator: false
  timeout: 15m
  type: game-loop
  grant-permissions: none
  device:
   - model: NexusLowRes
     version: 28
```
flank starts sanity robo run (simple robo test without script/directives) where gcloud throws error
```
ERROR: (gcloud.alpha.firebase.test.android.run) Invalid value for [use-orchestrator]: may not be used with test type [game-loop].
```
After I removed `use-orchestrator`:
```
ERROR: (gcloud.alpha.firebase.test.android.run)
Matrix [matrix-2tzdzhvxpdam7] failed during validation: The app does not have a correctly formatted game-loop intent filter.
```

so there is some logic that is currently missing in flank. As you mentioned, there are tasks to implement missing configs for game-loop. As I see it, this PR is only part of the 'enable game-loop tests in flank' feature. Therefore I was thinking if flank should print some info that this feature is not ready (when `type: game-loop`). But no strong opinion about it since we are working on a snapshot version anyway.
I just wanted to ensure we are on the same page (or I am missing something?)

Sloox:
@pawelpasterz I understand the concern. I think that i will proceed and complete all the tasks that influence the game-loop option. This includes: #1205 , #1202 , #1204 , #1203  These all i think are required to actually get it fully functional. 
I also feel there will be a final task making sure it all is actually testable and works as intended as there are bound to be nuances that might have been missed that we can fix (such as failing fast on use-orchestrator & a properly formatted game-loop intent filter)
But at the exact moment i do not think the testing you applied above will work until the rest of these tasks are completed.

This tasks is required actually to complete the next task i am working on add labels.

test_runner/src/main/kotlin/ftl/args/yml/Type.kt
Sloox: nice spot, fixed 👍

pawelpasterz:
> @pawelpasterz I understand the concern. I think that i will proceed and complete all the tasks that influence the game-loop option. This includes: #1205 , #1202 , #1204 , #1203 These all i think are required to actually get it fully functional.
> I also feel there will be a final task making sure it all is actually testable and works as intended as there are bound to be nuances that might have been missed that we can fix (such as failing fast on use-orchestrator & a properly formatted game-loop intent filter)
> But at the exact moment i do not think the testing you applied above will work until the rest of these tasks are completed.
> 
> This tasks is required actually to complete the next task i am working on add labels.

Perfect, looking forward to having this feature fully implemented :)

pawelpasterz (approved):

PR #1233, opened 2020-10-12
fix: Fix withClassName filter

### LINKED ISSUE #1227: gcloud test-targets flag can match multiple classes even if only one is configured
**Describe the bug**

If you have a package with the following files:
- com.xyz.MyTest1.kt
- com.xyz.MyTest2.kt
- com.xyz.MyTest3.kt
- com.xyz.MyTest10.kt
- com.xyz.MyTest11.kt
- com.xyz.MyTest12.kt
- com.xyz.MyTest13.kt

And configure the flank yaml as follows:
```yml
gcloud:
  test-targets:
    - class com.xyz.MyTest1
```

Flank will match the following test classes: MyTest1; MyTest10; MyTest11; MyTest12; MyTest13

And — let's say each test class has one test method — flank will print the following output

```
Uploading app.apk ....
Uploading app-androidTest.apk .
5 tests / 1 shard
```

**To Reproduce**
1. Create the test files as described above, each with one test method

My flank.yml

```yml
gcloud:
  test-targets:
    - class com.xyz.MyTest1
  record-video: true
  timeout: 60m
  async: false
  app: ./app.apk
  test: ./app-androidTest.apk
  auto-google-login: true
  use-orchestrator: true
  environment-variables:
    clearPackageData: true
  directories-to-pull:
    - /sdcard/screenshots
  performance-metrics: true
  device:
    - model: NexusLowRes
      version: 28
flank:
  testShards: 1
  ignore-failed-tests: true
  repeatTests: 1
```

**Expected behavior**
It should only run the tests of the specified class.

Flank version 20.09.3

pawelpasterz:
Fixes #1227, #1234 

Similar class/method names were considered as one due to `testMethod.testName.startsWith(className)` check, with:
```
test-targets:
  - class foo.bar.TestClass1
```
all `foo.bar.TestClass1`, `foo.bar.TestClass121`, `foo.bar.TestClass1aaaa` return `true` for `shouldRun` check. The same is applicable for test methods

1. build finish without any errors
1. generate apks with `[flank root] ./test_projects/android/bash/generate_apks.sh ~`
1. start flank run with config: 
    ```yaml
    gcloud:
      app: ~/app-debug.apk
      test: ~/app-multiple-success-debug-androidTest.apk
      test-targets:
        - class com.example.test_app.similar.SimilarNameTest1
    ```
    and observe only 3 tests are launched
1. start flank run with config: 
    ```yaml
    gcloud:
      app: ~/app-debug.apk
      test: ~/app-multiple-success-debug-androidTest.apk
      test-targets:
        - class com.example.test_app.similar.SimilarNameTest1#test1
    ```
    and observe only 1 test is launched

## Checklist

- [x] Unit tested

### DIFF
--- a/test_runner/src/main/kotlin/ftl/filter/TestFilters.kt
+++ b/test_runner/src/main/kotlin/ftl/filter/TestFilters.kt
@@ -153,19 +153,21 @@ object TestFilters {
     private fun withClassName(classNames: List<String>): TestFilter {
         // splits foo.bar.TestClass1#testMethod1 into [foo.bar.TestClass1, testMethod1]
         fun String.extractClassAndTestNames() = split("#")
-
         val classFilters = classNames.map { it.extractClassAndTestNames() }
         return TestFilter(
             describe = "withClassName (${classNames.joinToString(", ")})",
-            shouldRun = { testMethod ->
-                val testMethodName = testMethod.testName.extractClassAndTestNames()
-                classFilters.any { filter ->
-                    // When filter.size == 1 all test methods from the class should run therefore we do not compare method names
-                    // When filter.size != 1 only particular test from the class should be launched and we need to compare method names as well
-                    testMethodName[0] == filter[0] && (filter.size == 1 || testMethodName[1] == filter[1])
-                }
-            }
+            shouldRun = { testMethod -> testMethod.testName.extractClassAndTestNames().matchFilters(classFilters) }
         )
+    }
+
+    private fun List<String>.matchFilters(classFilters: List<List<String>>): Boolean {
+        fun List<String>.className() = first()
+        fun List<String>.methodName() = last()
+        return classFilters.any { filter ->
+            // When filter.size == 1 all test methods from the class should run therefore we do not compare method names
+            // When filter.size != 1 only particular test from the class should be launched and we need to compare method names as well
+            className() == filter.className() && (filter.size == 1 || methodName() == filter.methodName())
+        }
     }
 
     private fun withAnnotation(annotations: List<String>): TestFilter = TestFilter(
--- a/test_runner/src/main/kotlin/ftl/filter/TestFilters.kt
+++ b/test_runner/src/main/kotlin/ftl/filter/TestFilters.kt
@@ -151,15 +151,18 @@ object TestFilters {
     )
 
     private fun withClassName(classNames: List<String>): TestFilter {
-        val splittedClassNames = classNames.map { it.split("#") }
+        // splits foo.bar.TestClass1#testMethod1 into [foo.bar.TestClass1, testMethod1]
+        fun String.extractClassAndTestNames() = split("#")
+
+        val classFilters = classNames.map { it.extractClassAndTestNames() }
         return TestFilter(
             describe = "withClassName (${classNames.joinToString(", ")})",
             shouldRun = { testMethod ->
-                val splittedName = testMethod.testName.split("#")
-                splittedClassNames.any { className ->
-                    // className.size == 1 => foo.bar.TestClass1
-                    // className.size != 1 => foo.bar.TestClass1#testMethod1
-                    splittedName[0] == className[0] && (className.size == 1 || splittedName[1] == className[1])
+                val testMethodName = testMethod.testName.extractClassAndTestNames()
+                classFilters.any { filter ->
+                    // When filter.size == 1 all test methods from the class should run therefore we do not compare method names
+                    // When filter.size != 1 only particular test from the class should be launched and we need to compare method names as well
+                    testMethodName[0] == filter[0] && (filter.size == 1 || testMethodName[1] == filter[1])
                 }
             }
         )
--- a/test_runner/src/main/kotlin/ftl/filter/TestFilters.kt
+++ b/test_runner/src/main/kotlin/ftl/filter/TestFilters.kt
@@ -150,12 +150,20 @@ object TestFilters {
         }
     )
 
-    private fun withClassName(classNames: List<String>): TestFilter = TestFilter(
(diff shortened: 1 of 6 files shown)


test_runner/src/main/kotlin/ftl/filter/TestFilters.kt
```diff
@@ -150,12 +150,20 @@ object TestFilters {
...
+    private fun withClassName(classNames: List<String>): TestFilter {
+        val splittedClassNames = classNames.map { it.split("#") }
+        return TestFilter(
+            describe = "withClassName (${classNames.joinToString(", ")})",
+            shouldRun = { testMethod ->
+                val splittedName = testMethod.testName.split("#")
+                splittedClassNames.any { className ->
+                    // className.size == 1 => foo.bar.TestClass1
+                    // className.size != 1 => foo.bar.TestClass1#testMethod1
+                    splittedName[0] == className[0] && (className.size == 1 || splittedName[1] == className[1])
```
piotradamczyk5, line 162: I know that there is a comment, but could you wrap this code in some function(s) with good naming, because next month I think that it will be hard to understand

test_runner/src/main/kotlin/ftl/filter/TestFilters.kt
```diff
@@ -150,12 +150,20 @@ object TestFilters {
...
     )
 
-    private fun withClassName(classNames: List<String>): TestFilter = TestFilter(
-        describe = "withClassName (${classNames.joinToString(", ")})",
-        shouldRun = { testMethod ->
-            withPackageName(classNames).shouldRun(testMethod)
-        }
-    )
+    private fun withClassName(classNames: List<String>): TestFilter {
+        val splittedClassNames = classNames.map { it.split("#") }
```
piotradamczyk5, line 154: it.split make and class + method name, do you mean 
`val splittedClassNames = classNames.map { it.split("#").first() }` ?

test_runner/src/main/kotlin/ftl/filter/TestFilters.kt
```diff
@@ -150,12 +150,20 @@ object TestFilters {
...
+        return TestFilter(
+            describe = "withClassName (${classNames.joinToString(", ")})",
+            shouldRun = { testMethod ->
+                val splittedName = testMethod.testName.split("#")
```
piotradamczyk5, line 158: same here should not it be:
`testMethod.testName.split("#").last()` ?

test_runner/src/main/kotlin/ftl/run/platform/android/CreateAndroidTestContext.kt
```diff
@@ -70,7 +70,9 @@ internal fun InstrumentationTestContext.getFlankTestMethods(
             .filter(testFilter.shouldRun)
             .filterNot(parameterizedClasses::belong)
             .map(TestMethod::toFlankTestMethod).toList()
-            .plus(parameterizedClasses.map(String::toFlankTestMethod))
+            .plus(parameterizedClasses
+                .filter { testFilter.shouldRun(TestMethod(it, emptyList())) }
+                .map(String::toFlankTestMethod))
```
piotradamczyk5, line 75: could we wrap
```kotlin
parameterizedClasses
                .filter { testFilter.shouldRun(TestMethod(it, emptyList())) }
                .map(String::toFlankTestMethod)
```
to some method ex:`parametrizedClassesTestMethodWhichShouldRun` 
to avoid nested chains, WDYT?

test_runner/src/main/kotlin/ftl/filter/TestFilters.kt
pawelpasterz: Good point, I'll rewrite it to be more self-explanatory

test_runner/src/main/kotlin/ftl/filter/TestFilters.kt
pawelpasterz: I can't implement it that way since I need both: test class name and test method name. But I agree that part needs small refactor

test_runner/src/main/kotlin/ftl/filter/TestFilters.kt
pawelpasterz: I need to preserve information about the array size of the splitted string.

test_runner/src/main/kotlin/ftl/filter/TestFilters.kt
piotradamczyk5: @pawelpasterz  how about making it:
```kotlin

    private fun withClassName(classNames: List<String>): TestFilter {
        // splits foo.bar.TestClass1#testMethod1 into [foo.bar.TestClass1, testMethod1]
        fun String.extractClassAndTestNames() = split("#")
        val classFilters = classNames.map { it.extractClassAndTestNames() }
        return TestFilter(
            describe = "withClassName (${classNames.joinToString(", ")})",
            shouldRun = { testMethod -> testMethod.testName.extractClassAndTestNames().matchFilters(classFilters) }
        )
    }
    
    private fun List<String>.matchFilters(classFilters: List<List<String>>): Boolean {
        fun List<String>.className() = first()
        fun List<String>.methodName() = last()
        return classFilters.any { filter ->
            // When filter.size == 1 all test methods from the class should run therefore we do not compare method names
            // When filter.size != 1 only particular test from the class should be launched and we need to compare method names as well
            className() == filter.className() && (filter.size == 1 || methodName() == filter.methodName())
        }
    }
```

piotradamczyk5 (approved):

Sloox (approved):

adamfilipow92 (approved):

PR #1287, opened 2020-10-29
feat: ios scenario numbers

### LINKED ISSUE #1196: iOS | Add --scenario-numbers

As an iOS developer, I want to control scenario numbers so I can control which scenarios run.

> --scenario-numbers=int,[int,…]
> A list of game-loop scenario numbers which will be run as part of the test (default: scenario 1). A maximum of 1024 scenarios may be specified in one test matrix, but the maximum number may also be limited by the overall test --timeout setting. This flag is only valid when --type=game-loop is also set.
> 
> https://cloud.google.com/sdk/gcloud/reference/alpha/firebase/test/ios/run#--scenario-numbers

Sloox:
Fixes #1196 

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/test_runner/src/main/kotlin/ftl/args/ValidateIosArgs.kt
+++ b/test_runner/src/main/kotlin/ftl/args/ValidateIosArgs.kt
@@ -13,6 +13,13 @@ fun IosArgs.validate() = apply {
     assertTestFiles()
     checkResultsDirUnique()
     assertAdditionalIpas()
+    assertGameloop()
+}
+
+fun IosArgs.assertGameloop() {
+    /*if (scenarioNumbers.isNotEmpty() && (type == null || type != Type.GAMELOOP))
+        throw FlankConfigurationError("Scenario numbers defined but Type is not Game-loop.")*/
+    scenarioNumbers.forEach { it.toIntOrNull() ?: throw FlankConfigurationError("Invalid scenario number provided - $it") }
 }
 
 fun IosArgs.validateRefresh() = apply {
--- a/test_runner/src/main/kotlin/ftl/args/ValidateIosArgs.kt
+++ b/test_runner/src/main/kotlin/ftl/args/ValidateIosArgs.kt
@@ -19,9 +19,10 @@ fun IosArgs.validate() = apply {
 }
 
 fun IosArgs.assertGameloop() {
-    if (scenarioNumbers.isNotEmpty() && (type == null || type != Type.GAMELOOP))
+    if (scenarioNumbers.isNotEmpty() && (type != Type.GAMELOOP))
         throw FlankConfigurationError("Scenario numbers defined but Type is not Game-loop.")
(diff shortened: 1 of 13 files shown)


test_runner/src/main/kotlin/ftl/args/ValidateIosArgs.kt
```diff
@@ -13,6 +13,13 @@ fun IosArgs.validate() = apply {
     assertTestFiles()
     checkResultsDirUnique()
     assertAdditionalIpas()
+    assertGameloop()
+}
+
+fun IosArgs.assertGameloop() {
+    /*if (scenarioNumbers.isNotEmpty() && (type == null || type != Type.GAMELOOP))
+        throw FlankConfigurationError("Scenario numbers defined but Type is not Game-loop.")*/
```
piotradamczyk5, line 21: if this commented code needed?

Sloox: this is a draft pull request. changes will come.

test_runner/src/main/kotlin/ftl/args/ValidateIosArgs.kt
```diff
@@ -15,6 +15,13 @@ fun IosArgs.validate() = apply {
     checkResultsDirUnique()
     assertAdditionalIpas()
     validType()
+    assertGameloop()
+}
+
+fun IosArgs.assertGameloop() {
+    if (scenarioNumbers.isNotEmpty() && (type == null || type != Type.GAMELOOP))
```
piotradamczyk5, line 22:
```suggestion
    if (scenarioNumbers.isNotEmpty() && (type != Type.GAMELOOP))
```

test_runner/src/main/kotlin/ftl/args/ValidateIosArgs.kt
```diff
@@ -15,6 +15,13 @@ fun IosArgs.validate() = apply {
     checkResultsDirUnique()
     assertAdditionalIpas()
     validType()
+    assertGameloop()
+}
+
+fun IosArgs.assertGameloop() {
```
piotradamczyk5, line 21: you could check also if `scenariosNumber.size` is below `1024`

test_runner/src/main/kotlin/ftl/args/ValidateIosArgs.kt
Sloox: 👍

test_runner/src/main/kotlin/ftl/args/ValidateIosArgs.kt
Sloox: 👍

piotradamczyk5 (approved):

pawelpasterz (approved):

Issue #1276, opened 2020-10-27
iOS | Add --app


piotradamczyk5:
**Author the user story for this feature**

As an iOS developer, I want to set app so I can test my app using GameLoop type

**Describe the solution**
https://cloud.google.com/sdk/gcloud/reference/alpha/firebase/test/ios/run#--app

> --app=APP
> The path to the application archive (.ipa file) for game-loop testing. The path may be in the local filesystem or in Google Cloud Storage using gs:// notation. This flag is only valid when --type=game-loop is also set.

PR #1289, opened 2020-11-02
feat: add ios app for gameloop

Sloox:
Adds #1276 
IOS add app parameter

## Checklist

- [x] Documented
- [x] Unit tested


test_runner/src/test/kotlin/ftl/args/IosArgsTest.kt
```diff
@@ -245,6 +245,7 @@ IosArgs
         - $testIpa2
       scenario-numbers: 
       type: xctest
+      app:
```
piotradamczyk5, line 248: `nonDefault` should contain some value

Sloox: @piotradamczyk5 problem here is it will cause a conflict if i provide something here. Will get a validate exception.
`"Cant have xctest and app defined, must be game-loop"` 
WDYT?

piotradamczyk5: ok sure, so leave it 👍

piotradamczyk5 (approved):

pawelpasterz (approved):

PR #1299, opened 2020-11-04
feat: add test targets for shard

### LINKED ISSUE #1203: Android | Add --test-targets-for-shard

Sloox:
Fixes #1203 
add test targets for shard
https://cloud.google.com/sdk/gcloud/reference/alpha/firebase/test/android/run#--test-targets-for-shard

1. Run config [flank-single-shards-simple.yml](https://github.com/Flank/flank/blob/40adadab36a14a6707c2b53a0f4d1fc31e025ae4/test_runner/src/test/kotlin/ftl/fixtures/test_app_cases/flank-single-shards-simple.yml)
    ```
    gcloud:
      app: app-debug.apk
      test: app-multiple-success-debug-androidTest.apk
    test-targets-for-shard:
      - class com.example.test_app.bar.BarInstrumentedTest
      - class com.example.test_app.foo.FooInstrumentedTest
    ```
1. Observe a matrix with 2 shards is running
1. Each shard contains only 2 tests, and one of them is ignored
1. Number of dumped shards is equal to expected (2)

Above is just a simple test plan, please, feel free to play around with this feature.

## Checklist

- [x] Documented
- [x] Unit tested


test_runner/src/main/kotlin/ftl/args/ValidateAndroidArgs.kt
```diff
@@ -32,6 +33,18 @@ fun AndroidArgs.validate() = apply {
...
+private fun AndroidArgs.assertTestTargetForShards() {
+    if (testTargetsForShard.isNotEmpty()) {
+        if (numUniformShards != null) throw FlankConfigurationError("Number Of Uniform Shards and test Targets For Shards cannot be defined at the same time.")
+        if (type != Type.INSTRUMENTATION) throw FlankConfigurationError("Test target for shards can only be specified when test type is INSTRUMENTATION.")
+        if (devices.isEmpty() && testTargetsForShard.size > 500) {
+            throw FlankConfigurationError("Cannot have more than 500 Test Targets for Shards with no device provided.")
+        } else if (devices.isNotEmpty() && testTargetsForShard.size > 50) {
+            throw FlankConfigurationError("Cannot have more than 50 Test Targets for Shards if one or more devices are specified.")
+        }
+    }
```
piotradamczyk5, line 45: it will be better with `when`

test_runner/src/main/kotlin/ftl/args/ValidateAndroidArgs.kt
```diff
@@ -73,7 +88,11 @@ private fun AndroidArgs.assertType() = type?.let {
 }
 
 private fun AndroidArgs.assertGrantPermissions() = grantPermissions?.let {
-    if (it !in listOf("all", "none")) throw FlankGeneralError("Unsupported permission '$grantPermissions'\nOnly 'all' or 'none' supported.")
+    if (it !in listOf(
+            "all",
+            "none"
+        )
+    ) throw FlankGeneralError("Unsupported permission '$grantPermissions'\nOnly 'all' or 'none' supported.")
```
piotradamczyk5, line 95: could you make this `if` single line
```kotlin
if (it !in listOf( "all", "none"))
```

test_runner/src/main/kotlin/ftl/args/ValidateAndroidArgs.kt
```diff
@@ -83,8 +102,20 @@ private fun AndroidArgs.assertDevicesSupported() = devices
     .forEach { (device, check) ->
         when (check) {
             SupportedDeviceConfig -> Unit
-            UnsupportedModelId -> throw IncompatibleTestDimensionError("Unsupported model id, '${device.model}'\nSupported model ids: ${AndroidCatalog.androidModelIds(project)}")
-            UnsupportedVersionId -> throw IncompatibleTestDimensionError("Unsupported version id, '${device.version}'\nSupported Version ids: ${AndroidCatalog.androidVersionIds(project)}")
+            UnsupportedModelId -> throw IncompatibleTestDimensionError(
+                "Unsupported model id, '${device.model}'\nSupported model ids: ${
+                    AndroidCatalog.androidModelIds(
+                        project
+                    )
```
piotradamczyk5, line 109:
```kotlin
AndroidCatalog.androidModelIds(project)         
```
?

test_runner/src/main/kotlin/ftl/args/ValidateAndroidArgs.kt
```diff
@@ -83,8 +102,20 @@ private fun AndroidArgs.assertDevicesSupported() = devices
...
+                }"
+            )
+            UnsupportedVersionId -> throw IncompatibleTestDimensionError(
+                "Unsupported version id, '${device.version}'\nSupported Version ids: ${
+                    AndroidCatalog.androidVersionIds(
+                        project
+                    )
+                }"
```
piotradamczyk5, line 117: same as above single line 
```kotlin
AndroidCatalog.androidModelIds(project)         
```
?

test_runner/src/main/kotlin/ftl/args/ValidateAndroidArgs.kt
```diff
@@ -108,8 +139,8 @@ private fun AndroidArgs.assertDirectoriesToPull() {
         ?.also {
             throw FlankConfigurationError(
                 "Invalid value for [directories-to-pull]: Invalid path $it.\n" +
-                    "Path must be absolute paths under /sdcard or /data/local/tmp (for example, --directories-to-pull /sdcard/tempDir1,/data/local/tmp/tempDir2).\n" +
-                    "Path names are restricted to the characters [a-zA-Z0-9_-./+]. "
+                        "Path must be absolute paths under /sdcard or /data/local/tmp (for example, --directories-to-pull /sdcard/tempDir1,/data/local/tmp/tempDir2).\n" +
+                        "Path names are restricted to the characters [a-zA-Z0-9_-./+]. "
```
piotradamczyk5, line 143: bad formatting

test_runner/src/main/kotlin/ftl/args/ValidateIosArgs.kt
```diff
@@ -42,18 +44,21 @@ fun IosArgs.validateRefresh() = apply {
     assertMaxTestShards()
 }
 
-private fun IosArgs.assertMaxTestShards() { this.maxTestShards
+private fun IosArgs.assertMaxTestShards() {
+    this.maxTestShards
```
piotradamczyk5, line 48: I think that this line is not used

test_runner/src/main/kotlin/ftl/gc/android/CreateAndroidInstrumentationTest.kt
```diff
@@ -46,11 +48,19 @@ internal fun AndroidInstrumentationTest.setupTestTargets(
...
+                                TestTargetsForShard().setTestTargets(it.split(',', ';'))
+                            }
+                        )
+                    } else {
+                        ManualSharding().setTestTargetsForShard(
+                            testShards.map {
+                                TestTargetsForShard().setTestTargets(it)
+                            }
+                        )
+                    }
```
piotradamczyk5, line 63: How about ?

```kotlin
manualSharding = ManualSharding().setTestTargetsForShard(getTestShards())
...
fun getTestShards() = = if (testFlagForShard.isNotEmpty()) {
    testFlagForShard.map { TestTargetsForShard().setTestTargets(it.split(',', ';')) }
} else {
    testShards.map {TestTargetsForShard().setTestTargets(it) }
}
```

piotradamczyk5, line 63: this `else (in when)` definitely deserves a separate function

test_runner/src/main/kotlin/ftl/args/ValidateAndroidArgs.kt
```diff
@@ -58,7 +71,9 @@ private fun AndroidArgs.assertLabelContent() {
 
     if (scenarioNumbers.isNotEmpty() && (type != Type.GAMELOOP))
         throw FlankConfigurationError("Scenario numbers defined but Type is not Game-loop.")
-    scenarioNumbers.forEach { it.toIntOrNull() ?: throw FlankConfigurationError("Invalid scenario number provided - $it") }
+    scenarioNumbers.forEach {
+        it.toIntOrNull() ?: throw FlankConfigurationError("Invalid scenario number provided - $it")
+    }
```
piotradamczyk5, line 76:
```
    scenarioNumbers.validateScenarionsNumbers()
...
fun List<String>.validateScenarionsNumbers() =  forEach {
        it.toIntOrNull() ?: throw FlankConfigurationError("Invalid scenario number provided - $it")
    }
```

pawelpasterz (changes requested):
I've also noticed some inconsistency between shards printed to stdout and those that were actually run. With simple config
```
gcloud:
  app: (...)
  test: (...)
  test-targets-for-shard:
    - com.example.test_app.bar.BarInstrumentedTest
```
I get
Screenshot 2020-11-07 at 21 52 45
and `android_shards.json`
```
{
  "matrix-0": {
    "app": "[...]/app-debug.apk",
    "test": "[...]/app-multiple-success-debug-androidTest.apk",
    "shards": {
      "shard-0": [
        "class com.example.test_app.ParameterizedTest",
        "class com.example.test_app.parametrized.EspressoParametrizedClassParameterizedNamed",
        "class com.example.test_app.parametrized.EspressoParametrizedClassTestParameterized",
        "class com.example.test_app.parametrized.EspressoParametrizedMethodTestJUnitParamsRunner",
        "class com.example.test_app.InstrumentedTest#test0",
        "class com.example.test_app.InstrumentedTest#test1",
        "class com.example.test_app.InstrumentedTest#test2",
        "class com.example.test_app.bar.BarInstrumentedTest#testBar",
        "class com.example.test_app.foo.FooInstrumentedTest#testFoo",
        "class com.example.test_app.similar.SimilarNameTest1#test1",
        "class com.example.test_app.similar.SimilarNameTest1#test19",
        "class com.example.test_app.similar.SimilarNameTest1#test2",
        "class com.example.test_app.similar.SimilarNameTest10#test1",
        "class com.example.test_app.similar.SimilarNameTest10#test19",
        "class com.example.test_app.similar.SimilarNameTest10#test2"
      ]
    },
    "junit-ignored": [
      "class com.example.test_app.InstrumentedTest#ignoredTestWitSuppress",
      "class com.example.test_app.InstrumentedTest#ignoredTestWithIgnore",
      "class com.example.test_app.bar.BarInstrumentedTest#ignoredTestBar",
      "class com.example.test_app.foo.FooInstrumentedTest#ignoredTestFoo"
    ]
  }
}
```
To clarify, flank creates a matrix correctly.

Secondly, flank does not attach (to the matrix) tests annotated with `@Ignore` therefore they are not present in firebase console results. With `test-targets-for-shards` this is not true.

test_runner/src/main/kotlin/ftl/args/ValidateAndroidArgs.kt
Sloox: detekt autofix, but i will revert

test_runner/src/main/kotlin/ftl/args/ValidateAndroidArgs.kt
Sloox: same as above

test_runner/src/main/kotlin/ftl/args/ValidateIosArgs.kt
Sloox: agreed. What a weird addition.

test_runner/src/main/kotlin/ftl/gc/android/CreateAndroidInstrumentationTest.kt
Sloox: nice cleanup. 👍

Sloox: unfortunately after several attempts it actually is very messy implementation due to all the requirements of the class. I will skip as it makes it more unreadable imho.

test_runner/src/main/kotlin/ftl/args/ValidateAndroidArgs.kt
Sloox: I dont think this actually achieves anything beyond giving a name to a very simple to read function. WDYT?

test_runner/src/main/kotlin/ftl/args/CreateAndroidArgs.kt
```diff
@@ -41,5 +41,6 @@ fun createAndroidArgs(
     obfuscateDumpShards = obfuscate,
     obbFiles = gcloud::obbfiles.require(),
     obbNames = gcloud::obbnames.require(),
-    grantPermissions = gcloud.grantPermissions
+    grantPermissions = gcloud.grantPermissions,
+    testTargetsForShard = gcloud.testTargetsForShard.orEmpty()
```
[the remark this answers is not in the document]
Sloox: hmmm alright will nullify

test_runner/src/main/kotlin/ftl/args/ValidateAndroidArgs.kt
```diff
@@ -32,6 +33,18 @@ fun AndroidArgs.validate() = apply {
     checkNumUniformShards()
 }
 
+private fun AndroidArgs.assertTestTargetForShards() {
+    if (testTargetsForShard.isNotEmpty()) {
+        if (numUniformShards != null) throw FlankConfigurationError("Number Of Uniform Shards and test Targets For Shards cannot be defined at the same time.")
+        if (type != Type.INSTRUMENTATION) throw FlankConfigurationError("Test target for shards can only be specified when test type is INSTRUMENTATION.")
```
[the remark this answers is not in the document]
Sloox: its changed to when. @pawelpasterz it requires instrumentation to be defined. I know in the past we did not have this type class and it was calculated automatically. But i think its good to use this as a strict enforcement.
wdyt?
Aslo can you explain, "gcloud proceeds normally?"

test_runner/src/test/kotlin/ftl/args/AndroidArgsTest.kt
```diff
@@ -1978,13 +1980,12 @@ AndroidArgs
         AndroidArgs.load(yaml).validate()
     }
 
-    @Test(expected = FlankConfigurationError::class)
+    @Test(expected = FlankGeneralError::class)
```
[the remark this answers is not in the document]
Sloox: @pawelpasterz where can i confirm this?

Sloox:
> app-multiple-success-debug-androidTest.apk



> I've also noticed some inconsistency between shards printed to stdout and those that were actually run. With simple config
> 
> ```
> gcloud:
>   app: (...)
>   test: (...)
>   test-targets-for-shard:
>     - com.example.test_app.bar.BarInstrumentedTest
> ```
> 
> I get
> Screenshot 2020-11-07 at 21 52 45
> and `android_shards.json`
> 
> ```
> {
>   "matrix-0": {
>     "app": "[...]/app-debug.apk",
>     "test": "[...]/app-multiple-success-debug-androidTest.apk",
>     "shards": {
>       "shard-0": [
>         "class com.example.test_app.ParameterizedTest",
>         "class com.example.test_app.parametrized.EspressoParametrizedClassParameterizedNamed",
>         "class com.example.test_app.parametrized.EspressoParametrizedClassTestParameterized",
>         "class com.example.test_app.parametrized.EspressoParametrizedMethodTestJUnitParamsRunner",
>         "class com.example.test_app.InstrumentedTest#test0",
>         "class com.example.test_app.InstrumentedTest#test1",
>         "class com.example.test_app.InstrumentedTest#test2",
>         "class com.example.test_app.bar.BarInstrumentedTest#testBar",
>         "class com.example.test_app.foo.FooInstrumentedTest#testFoo",
>         "class com.example.test_app.similar.SimilarNameTest1#test1",
>         "class com.example.test_app.similar.SimilarNameTest1#test19",
>         "class com.example.test_app.similar.SimilarNameTest1#test2",
>         "class com.example.test_app.similar.SimilarNameTest10#test1",
>         "class com.example.test_app.similar.SimilarNameTest10#test19",
>         "class com.example.test_app.similar.SimilarNameTest10#test2"
>       ]
>     },
>     "junit-ignored": [
>       "class com.example.test_app.InstrumentedTest#ignoredTestWitSuppress",
>       "class com.example.test_app.InstrumentedTest#ignoredTestWithIgnore",
>       "class com.example.test_app.bar.BarInstrumentedTest#ignoredTestBar",
>       "class com.example.test_app.foo.FooInstrumentedTest#ignoredTestFoo"
>     ]
>   }
> }
> ```
> 
> To clarify, flank creates a matrix correctly.
> 
> Secondly, flank does not attach (to the matrix) tests annotated with `@Ignore` therefore they are not present in firebase console results. With `test-targets-for-shards` this is not true.

Interestingly here can you paste your config out here? These tests require a testrunner. I tested this exact config and it should fail at the test-runner flag not found step.

test_runner/src/main/kotlin/ftl/args/ValidateAndroidArgs.kt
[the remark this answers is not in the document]
pawelpasterz: > Aslo can you explain, "gcloud proceeds normally?"

Sure, what I meant is, with the config I posted, I can launch a test run with `gcloud` which ends with success. With `flank` I get the error I mentioned previously.

> But i think its good to use this as a strict enforcement.
wdyt?

I agree and would like the flank to be more robust yet the rule of thumb is to behave the same as `gcloud`.
I think that the main gain is companies can easily switch from `gcloud` to `flank` just by replacing cli tool.

Asking @bootstraponline for opinion

test_runner/src/test/kotlin/ftl/args/AndroidArgsTest.kt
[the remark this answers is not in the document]
pawelpasterz: I don't know if there is another way, I just fail gcloud and observe exit code in terminal

test_runner/src/main/kotlin/ftl/args/ValidateAndroidArgs.kt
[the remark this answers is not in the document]
Sloox: Lets go with your change. If we are attempting to keep as close to the gcloud then its a good enough reason. Will update to your suggestion as it also takes into account the type variable.

test_runner/src/main/kotlin/ftl/run/platform/android/CreateInstrumentationConfig.kt
```diff
@@ -15,5 +15,6 @@ internal fun AndroidArgs.createInstrumentationConfig(
     numUniformShards = numUniformShards,
     testShards = testApk.shards.testCases,
     keepTestTargetsEmpty = disableSharding && testTargets.isEmpty(),
-    environmentVariables = testApk.environmentVariables
+    environmentVariables = testApk.environmentVariables,
+    testFlagForShard = testTargetsForShard
```
jan-gogo, line 19: Changing the name from testTargetsForShard into testFlagForShard can be confusing.

test_runner/src/main/kotlin/ftl/gc/android/CreateAndroidInstrumentationTest.kt
```diff
@@ -46,13 +48,24 @@ internal fun AndroidInstrumentationTest.setupTestTargets(
...
             }
         }
     }
 }
+
+private fun createManualSharding(testFlagForShard: List<String>?, testShards: ShardChunks) =
+    if (testFlagForShard != null && testFlagForShard.isNotEmpty()) {
+        ManualSharding().setTestTargetsForShard(
+            testFlagForShard.map {
+                TestTargetsForShard().setTestTargets(it.split(',', ';'))
```
jan-gogo, line 62: IMO this split should be done earlier, probably inside `createAndroidArgs` function. Here we want to operate on well prepared and understandable domain structures, not encoded strings.

test_runner/src/test/kotlin/ftl/args/IosArgsTest.kt
```diff
@@ -79,6 +79,7 @@ class IosArgsTest {
           num-flaky-test-attempts: 4
           type: xctest
           test-special-entitlements: true
+
```
jan-gogo, line 82: Why this extra line is needed?

test_runner/src/main/kotlin/ftl/run/platform/android/CreateInstrumentationConfig.kt
Sloox: agreed. It appears i made a mistake here with the name.

test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
```diff
@@ -23,7 +23,8 @@ data class AndroidArgs(
     val testTargets: List<String>,
     val additionalAppTestApks: List<AppTestPair>,
     override val useLegacyJUnitResult: Boolean,
-    val obfuscateDumpShards: Boolean
+    val obfuscateDumpShards: Boolean,
+    val testTargetsForShard: List<String>?
```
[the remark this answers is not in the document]
Sloox: Actually there was a bug where by making it not nullable it caused issues with an empty list and was populated earlier. Having it nullable stopped that.

test_runner/src/main/kotlin/ftl/gc/android/CreateAndroidInstrumentationTest.kt
Sloox: Please have a look at the most recent changes about this 👍

test_runner/src/test/kotlin/ftl/args/IosArgsTest.kt
Sloox: absolutely no idea why this is here. Removing

test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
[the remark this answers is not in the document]
Sloox: Made significant changes. Have another look 👍

test_runner/src/main/kotlin/ftl/args/ValidateAndroidArgs.kt
```diff
@@ -32,6 +33,17 @@ fun AndroidArgs.validate() = apply {
     checkNumUniformShards()
 }
 
+private fun AndroidArgs.assertTestTargetForShards() {
+    if (testTargetsForShard != null && testTargetsForShard.isNotEmpty()) {
```
jan-gogo, line 37: This should fix the compilation error

```suggestion
    if (testTargetsForShard.isNotEmpty()) {
```

test_runner/src/main/kotlin/ftl/args/ValidateAndroidArgs.kt
```diff
@@ -34,7 +34,7 @@ fun AndroidArgs.validate() = apply {
 }
 
 private fun AndroidArgs.assertTestTargetForShards() {
-    if (testTargetsForShard != null && testTargetsForShard.isNotEmpty()) {
+    if (testTargetsForShard.isNotEmpty()) {
```
Sloox, line 37: nice i must have missed it :

test_runner/src/test/kotlin/ftl/fixtures/test_app_cases/flank-single-shards-test.yml
```diff
@@ -0,0 +1,5 @@
+gcloud:
+  app: test_runner/src/test/kotlin/ftl/fixtures/tmp/apk/app-debug.apk
+  test: test_runner/src/test/kotlin/ftl/fixtures/tmp/apk/app-multiple-success-debug-androidTest.apk
+  test-targets-for-shard:
+    - com.example.test_app.bar.BarInstrumentedTest
```
pawelpasterz, line 5: I think there is `class` missing. 
We could also add another shard. IMO it will be more realistic
```
- class com.example.test_app.foo.FooInstrumentedTest
```

pawelpasterz (dismissed):
I've spotted some problems with printing config
Screenshot 2020-11-12 at 09 19 07

------

For config
```
gcloud:
  app: test_runner/src/test/kotlin/ftl/fixtures/tmp/apk/app-debug.apk
  test: test_runner/src/test/kotlin/ftl/fixtures/tmp/apk/app-multiple-success-debug-androidTest.apk
  test-targets-for-shard:
    - class com.example.test_app.bar.BarInstrumentedTest
flank:
  max-test-shards: 50
```
I can still see some problems with tests/classes/shards number printed
Screenshot 2020-11-12 at 09 32 11

----

For config
```
gcloud:
  app: test_runner/src/test/kotlin/ftl/fixtures/tmp/apk/app-debug.apk
  test: test_runner/src/test/kotlin/ftl/fixtures/tmp/apk/app-multiple-success-debug-androidTest.apk
  test-targets:
    - class com.example.test_app.InstrumentedTest#test2
  test-targets-for-shard:
    - class com.example.test_app.bar.BarInstrumentedTest
    - class com.example.test_app.foo.FooInstrumentedTest
```
`gcloud` return error message
Screenshot 2020-11-12 at 09 46 36
`Http error while creating test matrix: ResponseError 400: Test targets specified at the matrix level are not allowed if you are using Test Lab's manual sharding option.`
`flank` proceeds without an error although tests are shared correctly :)

test_runner/src/test/kotlin/ftl/fixtures/test_app_cases/flank-single-shards-test.yml
Sloox: hmm will look into it.

pawelpasterz:
I can also see, ignore tests are included in the test matrix that is sent to FTL (flank filter them to prevent case when all tests in a shard are ignored). Although, when someone uses manual sharding it's his responsibility to make sure all shards are correct (at least `gcloud` does not have any special feature for noticing about shard(s) with all test ignored...). On the other hand, is a nice feature from the user perspective...
🤷‍♂️

test_runner/src/test/kotlin/ftl/fixtures/test_app_cases/flank-single-shards-test.yml
Sloox: Yes class is missing, will add another class to as a reference test.

bootstraponline:
> I can also see, ignored tests are included in the test matrix that is sent to FTL (flank filters them to prevent case when all tests in a shard are ignored). Although, when someone uses manual sharding it's his responsibility to make sure all shards are correct (at least `gcloud` does not have any special feature for noticing about shard(s) with all test ignored...). On the other hand, it's a nice feature from the user perspective...

with the gcloud flags, we should try to match gcloud behavior (even if it's not great). The flank flags are our opportunity to deliver a better experience.

test_runner/src/main/kotlin/ftl/filter/TestFilters.kt
```diff
@@ -73,7 +74,11 @@ object TestFilters {
         }
     }
 
-    fun fromTestTargets(targets: List<String>): TestFilter {
+    fun fromTestTargets(testTargets: List<String>) = fromTestTargets(testTargets, emptyList())
```
jan-gogo, line 77: Is this method necessary? Why not add a default parameter for the method below?

test_runner/src/main/kotlin/ftl/run/platform/android/ResolveApks.kt
```diff
@@ -21,7 +21,7 @@ internal fun AndroidArgs.resolveApks(): List<AndroidTestContext> = listOfNotNull
 
 private fun AndroidArgs.mainApkContext() = appApk?.let { appApk ->
     when {
-        testApk != null -> InstrumentationTestContext(app = appApk.asFileReference(), test = testApk.asFileReference(), environmentVariables = emptyMap())
+        testApk != null -> InstrumentationTestContext(app = appApk.asFileReference(), test = testApk.asFileReference(), environmentVariables = emptyMap(), testTargetsForShard = testTargetsForShard)
```
jan-gogo, line 24: This line is way too long. Please wrap lines longer than 80 columns, cause are unreadable on split screen.

test_runner/src/main/kotlin/ftl/shard/Shard.kt
```diff
@@ -35,6 +36,20 @@ class com.foo.ClassName#testMethodToSkip
...
+    val testShards = mutableListOf<TestShard>()
+    args.testTargetsForShard.forEach { list ->
+        val testMethods = mutableListOf<TestMethod>()
+        list.forEach { name ->
+            testMethods.add(TestMethod(name, 0.0, false))
+        }
+        testShards.add(TestShard(testMethods[testMethods.lastIndex].time, testMethods))
+    }
+    return testShards
+}
```
jan-gogo, line 51:
```suggestion
fun createShardsByTestForShards(
    args: AndroidArgs
): List<TestShard> =
    args.testTargetsForShard.map { list ->

        val testMethods = list
            .map { name -> TestMethod(name, 0.0, false) }
            .toMutableList()

        TestShard(
            time = testMethods[testMethods.lastIndex].time,
            testMethods = testMethods
        )
    }
```

Pls use expressions and avoid operations on mutable collections where possible.

test_runner/src/main/kotlin/ftl/filter/TestFilters.kt
Sloox: old habits from java really do die hard.

test_runner/src/main/kotlin/ftl/run/platform/android/ResolveApks.kt
Sloox: i wonder why detekt didnt pick it up

test_runner/src/main/kotlin/ftl/shard/Shard.kt
```diff
@@ -35,6 +36,20 @@ class com.foo.ClassName#testMethodToSkip
 
 */
 
+fun createShardsByTestForShards(
+    args: AndroidArgs
+): List<TestShard> {
+    return args.testTargetsForShard.map { list ->
```
jan-gogo, line 42: Again, pls use expressions instead of code block ;P, the return is redundant.

test_runner/src/test/kotlin/ftl/fixtures/test_app_cases/flank-single-shards-simple.yml
```diff
@@ -0,0 +1,6 @@
+gcloud:
+  app: test_runner/src/test/kotlin/ftl/fixtures/tmp/apk/app-debug.apk
+  test: test_runner/src/test/kotlin/ftl/fixtures/tmp/apk/app-multiple-success-debug-androidTest.apk
+  test-targets-for-shard:
+    - class com.example.test_app.bar.BarInstrumentedTest;class com.example.test_app.bar.BarInstrumentedTest
+    - class com.example.test_app.foo.FooInstrumentedTest
```
adamfilipow92, line 6: Empty line on end? 😄

flank-scripts/src/test/kotlin/flank/scripts/utils/FileKtTest.kt
```diff
@@ -14,7 +14,7 @@ class FileKtTest {
 
     @Test
     fun `should create symbolic link`() {
-
+        if (isWindows) return
```
adamfilipow92, line 17: We could use ```assumeFase(isWindows)``` wdyt?

docs/index.md
```diff
@@ -526,6 +526,15 @@ gcloud:
...
   #   version: 27
 
+  ### test-targets-for-shard
+  ## Specifies a group of packages, classes, and/or test cases to run in each shard (a group of test cases). 
+  ## The shards are run in parallel on separate devices. You can repeat this flag up to 50 times to specify multiple shards when one or more physical devices are selected, 
+  ## or up to 500 times when no physical devices are selected.
+  ## Note: If you include the flags environment-variable or test-targets when running test-targets-for-shard, the flags are applied to all the shards you create.
+  # test-target-for-shard:
+  # - package com.package1.for.shard1
+  # - class com.package2.for.shard2.class
```
pawelpasterz, line 536:
```suggestion
  # - class com.package2.for.shard2.Class
```

test_runner/flank.yml
```diff
@@ -315,3 +324,12 @@ flank:
...
   # disable-results-upload: false
+
+  ### test-targets-for-shard
+  ## Specifies a group of packages, classes, and/or test cases to run in each shard (a group of test cases).
+  ## The shards are run in parallel on separate devices. You can repeat this flag up to 50 times to specify multiple shards when one or more physical devices are selected,
+  ## or up to 500 times when no physical devices are selected.
+  ## Note: If you include the flags environment-variable or test-targets when running test-targets-for-shard, the flags are applied to all the shards you create.
+  # test-target-for-shard:
+  # - package com.package1.for.shard1
+  # - class com.package2.for.shard2.class
```
pawelpasterz, line 335:
```suggestion
  # - class com.package2.for.shard2.Class
```

pawelpasterz:
Unfortunately still a printing issue
Screenshot 2020-11-17 at 06 24 17

test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
```diff
@@ -265,6 +265,23 @@ object ArgsHelper {
...
     }
+
+    fun calculateDummyShards(
+        filteredTests: List<FlankTestMethod>,
+        args: AndroidArgs,
+    ): CalculateShardsResult {
+        if (filteredTests.isEmpty()) {
+            return CalculateShardsResult(emptyList(), emptyList()) // Avoid unnecessary computing if we already know there aren't tests to run.
+        }
+        val (ignoredTests, _) = filteredTests.partition { it.ignored }
```
pawelpasterz, line 276: I think we could skip this part. As result, we will have duplicated 'ignored' tests in XML results

test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
```diff
@@ -265,6 +265,23 @@ object ArgsHelper {
...
+
+        val shards = createShardsByTestForShards(args).map { Chunk(it.testMethods) }
+
+        return CalculateShardsResult(
+            testMethodsAlwaysRun(shards, args),
+            ignoredTestCases = ignoredTests.map { it.testName }
```
pawelpasterz, line 282: https://github.com/Flank/flank/pull/1299/files#r524894160
And empty list here.

pawelpasterz:
I can see the problem `3)` from https://github.com/Flank/flank/pull/1299#pullrequestreview-528863490 (the same config used)

test_runner/src/main/kotlin/ftl/filter/TestFilters.kt
```diff
@@ -73,7 +74,9 @@ object TestFilters {
         }
     }
 
-    fun fromTestTargets(targets: List<String>): TestFilter {
+    fun fromTestTargets(testTargets: List<String>, testTargetsForShard: ShardChunks = emptyList()): TestFilter {
+        val targets = testTargets + testTargetsForShard.flatten()
```
pawelpasterz, line 78: I think this is not entirely correct. `testTargets` and `testTargetsForShard` should never be defined at the same time therefore, this line may be misleading for the reader. A new function, for `testTargetsForShard` only would be more self-explanatory. (we can reuse logic common logic ofc)
Let me know your opinion

Sloox:
> Unfortunately still a printing issue
> Screenshot 2020-11-17 at 06 24 17

I am unsure how to approach this one? Any ideas?

test_runner/src/main/kotlin/ftl/filter/TestFilters.kt
Sloox: Ironically they can be defined at the same time. 
The documentation 
`## Note: If you include the flags environment-variable or test-targets when running test-targets-for-shard, the flags are applied to all the shards you create.`

It implies if i understand it correctly that the flags can be defined together and they effect each other so our logic is to filter based on the `testTargets` i am merely flattening the testTargetsForShard and further filtering. 

Makes sense?

pawelpasterz: Ah.. you are right, I was confused with the message I got from `gcloud`. My comment is not valid anymore. Thanks for explanation 👍

adamfilipow92:
> > Unfortunately still a printing issue
> > Screenshot 2020-11-17 at 06 24 17
> 
> I am unsure how to approach this one? Any ideas?

```ShardChunks``` is a List of List of Strings this is the reason why the single string is wrapped to ```[]```

so you could do something like this

```

fun shardsChunksListToString(objects: ShardChunks): String {
        if (objects.isNullOrEmpty()) return ""
        return NEW_LINE + objects.flatten()
            .joinToString("\n") { "         $it" }
    }

```
This should produce a result like:

```
      test-targets-for-shard: 
         class com.example.test_app.bar.BarInstrumentedTest
         class com.example.test_app.foo.FooInstrumentedTest
```

**Review by adamfilipow92 — APPROVED** [review]
- [X] Tested with filters
- [X] Tested without filters
- [X] Now output looks correct
```
    num-flaky-test-attempts: 0
      test-targets-for-shard: 
        - class com.example.test_app.bar.BarInstrumentedTest
        - class com.example.test_app.foo.FooInstrumentedTest
```

Good work 👍

File: test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt [hunk 30]
```diff
@@ -265,6 +265,20 @@ object ArgsHelper {
 
         return shards.map { Chunk(find + it.testMethods.filterNot { method -> find.contains(method) }) }
     }
+
+    fun calculateDummyShards(
```
jan-gogo, line 269: Do we really need 2 methods named `calculateDummyShards` which both have only one call? IMO It's unnecessary redundancy. Besides that `ArgHelper` will be refactored soon.

test_runner/src/main/kotlin/ftl/shard/Shard.kt
```diff
@@ -35,6 +36,16 @@ class com.foo.ClassName#testMethodToSkip
 
 */
 
+fun createShardsByTestForShards(args: AndroidArgs) = args.testTargetsForShard.map { list ->
```
jan-gogo, line 39: IMO the extension function will fit better.

jan-gogo (changes requested):
~~Would be great to have a dedicated `yml` config for each testable flank's feature. So, the config from the description could be placed in the repository.~~

[EDIT] My bad, it already is

test_runner/src/main/kotlin/ftl/shard/Shard.kt
Sloox: 👍

test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
pawelpasterz: I believe the intention was to be in line with the current implementation https://github.com/Flank/flank/blob/733e444bc4db1b80c9117c409c7ecec898f2c344/test_runner/src/main/kotlin/ftl/run/platform/android/CreateAndroidTestContext.kt#L49-L52
As you mentioned, there will be a huge refactor soon so within this PR we either refactor all similar occurrences or leave it as it is

test_runner/src/main/kotlin/ftl/shard/Shard.kt
pawelpasterz: I think my comment https://github.com/Flank/flank/pull/1299#discussion_r526025879 is also applicable in this case

test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
Sloox: removed the reference from ArgsHelper 👍

Sloox: @pawelpasterz I removed the reference from ArgsHelper. and merely moved the logic into the CreateAndroidTestContext

test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
```diff
@@ -274,3 +273,9 @@ fun String.normalizeFilePath(): String =
     } catch (e: Throwable) {
         this
     }
+
+fun List<String>.normalizeToTestTargets(): ShardChunks {
+    return if (this.isEmpty().not()) {
+        this.map { it.split(',', ';') }
+    } else emptyList()
+}
```
jan-gogo, line 281: Less code to read == less code to understand
```suggestion
fun List<String>.normalizeToTestTargets(): ShardChunks =
    if (isEmpty()) emptyList()
    else map { it.split(',', ';') }
```

test_runner/src/main/kotlin/ftl/shard/Shard.kt
```diff
@@ -35,6 +36,16 @@ class com.foo.ClassName#testMethodToSkip
 
 */
 
+fun AndroidArgs.createShardsByTestForShards() = this.testTargetsForShard.map { list ->
```
jan-gogo, line 39: It's much better to expose return type instead of `this` especially when function is public
```suggestion
fun AndroidArgs.createShardsByTestForShards(): List<Chunk> = testTargetsForShard.map { list ->

```

jan-gogo (approved):

test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
Sloox: Already resolved.

test_runner/src/main/kotlin/ftl/shard/Shard.kt
Sloox: Already resolved.

piotradamczyk5 (approved):

piotradamczyk5:
@Mergifyio refresh

PR #1323, opened 2020-11-16
fix: Nested subdirectories in results-dir

### LINKED ISSUE #1309: Support for subdirectories in results-dir

As an engineer, I want to organize results from different flank runs so I can organize and simplify navigation in my Google Cloud Storage bucket.

On our pipelines we run multiple flank setups for different apps and use Fladle for the integration with Gradle.
I have configured each of the fladle blocks with the lines similar to the following:
```
fladle {
  ...
  resultsDir = "${readFlankResultDirectoryName()}/my-app-test-run-results"
}
```
In order to have multiple separate directories under the GCS bucket so that the test result artifacts (`JUnitReport`, `MatrixResults`, `CostReport`) don’t overwrite each other.

**Describe the solution**

Assuming I have 2 flank runs in a single pipeline, what I want to achieve is
```
gs://<bucket-name>/<pipeline-run>/<flank-run-1>/<artifact.txt>
gs://<bucket-name>/<pipeline-run>/<flank-run-2>/<artifact.txt>
```

What I could do as a workaround is to define a namespace to add as part of my bucket directory name.

Right now I still have to do something like

```
gs://<bucket-name>/<pipeline-run-flank-run-1>/<artifact.txt>
gs://<bucket-name>/<pipeline-run-flank-run-2>/<artifact.txt>
```

This is not optimal for dev. experience within my company because engineers have to search for all these directories on the root of my bucket that _happen to be related_. That is why I would rather organize each pipeline run into a single directory with nested directories for each flank run.

piotradamczyk5:
Fixes #1309 

Nested directories are supported out of the box in GCloud results upload. 
However Flank has an issue that it throws an exception because it did not properly store results in the local directory.
In order to support nested directories, local issue must be fixed

1. Set configuration to contains nested result directories usimg `results-dir`
ex: 
```yaml
gcloud:
...
  results-dir: nested/results
....
```
2. Run tests and see that results are properly updated to  GCloud storage and stored locally

## Checklist

- [x] Unit tested

### DIFF
--- a/test_runner/src/main/kotlin/ftl/util/MatrixUtil.kt
+++ b/test_runner/src/main/kotlin/ftl/util/MatrixUtil.kt
@@ -5,14 +5,10 @@ import ftl.json.MatrixMap
 import java.io.File
 import java.util.concurrent.TimeUnit
 
-fun resolveLocalRunPath(matrices: MatrixMap, args: IArgs): String {
-    if (args.useLocalResultDir()) return args.localResultDir
-
-    val runPath = File(matrices.runPath)
-    if (!runPath.exists()) return join(args.localResultDir, runPath.name)
-
-    // avoid File().toString() as that will use a '\' path separator.
-    return matrices.runPath
+fun resolveLocalRunPath(matrices: MatrixMap, args: IArgs) = when {
+    args.useLocalResultDir() -> args.localResultDir
+    File(matrices.runPath).exists() -> matrices.runPath
+    else -> join(args.localResultDir, matrices.runPath).also { localPath -> File(localPath).mkdirs() }
 }
 
 fun timeoutToSeconds(timeout: String): Long {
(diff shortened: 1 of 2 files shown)


pawelpasterz:
Please, update the description with more explanatory Test Plan

pawelpasterz:
I think it would be worth verifying if the subdirectories feature requires additional permissions. If it does, we could display a 
suitable message

piotradamczyk5:
> Please, update the description with more explanatory Test Plan

done 👍

piotradamczyk5:
> I think it would be worth verifying if the subdirectories feature requires additional permissions. If it does, we could display a
> suitable message

With current permissions, it works fine so I guess it does not need additional permission

Sloox (approved):

adamfilipow92 (approved):
- [X] Tested, flank finish without exception
- [X] Results uploaded properly 
👍

PR #1326, opened 2020-11-17
fix: Downloading performance Metrics

### LINKED ISSUE #1324: Cannot download performanceMetrics.json using "files-to-download" option
**Describe the bug**
Trying to use the files-to-download option to download the "performanceMetrics.json" file without a success.

**To Reproduce**
1. set the following values for files-to-download in flank yml file:
```
files-to-download:
   - .*\d{4}_logcat$
   - .*\.json$
```
2. couldn't see "performanceMetrics.json" being downloaded into the flank result folder

**Expected behavior**
"performanceMetrics.json" can be downloaded

**Details**
Tested using the locally generated flank build using the revision: 9ac7371037e (Nov 12, 2020)

piotradamczyk5:
Fixes #1324 

**Scenario A:**

1. Run flank on Android physical device. For example with configuration entry:
```yaml
gcloud:
...
  device:
    - model: OnePlus6T
      version: 28
...
```
2. Check that `performanceMetrics.json` is stored on the local results device

**Scenario B:**

1. Run flank on Android physical device. For example with configuration entry:
```yaml
gcloud:
...
  device:
    - model: OnePlus6T
      version: 28
...
```
2. Specify also file to download from GCloud bucket. For example with configuration entry:
```yaml
flank:
...
  files-to-download:
    - .*\.json$
...
```
3. Check that `performanceMetrics.json` is downloaded on the local results device

## Checklist

- [x] Save `performanceMetrics.json` by default to local results directory
- [x] Fix downloading after tests run
- [x] Fix local storage directory structure
- [x] Unit tested

### DIFF
--- a/test_runner/src/main/kotlin/ftl/reports/api/PerformanceMetrics.kt
+++ b/test_runner/src/main/kotlin/ftl/reports/api/PerformanceMetrics.kt
@@ -3,23 +3,37 @@ package ftl.reports.api
 import com.google.api.services.testing.model.TestExecution
 import com.google.api.services.toolresults.model.PerfMetricsSummary
 import ftl.android.AndroidCatalog
+import ftl.args.IArgs
 import ftl.gc.GcStorage
 import ftl.gc.GcToolResults
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.runBlocking
+import java.nio.file.Files
+import java.nio.file.Paths
 
 internal fun List<Pair<TestExecution, String>>.getAndUploadPerformanceMetrics(
-    resultBucket: String
+    args: IArgs
 ) = runBlocking {
     filterAndroidPhysicalDevicesRuns()
         .map { (testExecution, gcsStoragePath) ->
             async(Dispatchers.IO) {
-                testExecution.getPerformanceMetric().upload(resultBucket = resultBucket, resultDir = gcsStoragePath)
+                val performanceMetrics = testExecution.getPerformanceMetric()
+                performanceMetrics.save(gcsStoragePath, args)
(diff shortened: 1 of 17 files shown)


piotradamczyk5:
@Mergifyio rebase

test_runner/src/main/kotlin/ftl/reports/api/PerformanceMetrics.kt
```diff
@@ -3,25 +3,39 @@ package ftl.reports.api
...
+                val performanceMetrics = testExecution.getPerformanceMetric()
+                performanceMetrics.save(gcsStoragePath, args)
+                performanceMetrics.upload(resultBucket = args.resultsBucket, resultDir = gcsStoragePath)
             }
         }
         .awaitAll()
 }
 
+private fun PerfMetricsSummary.save(resultsDir: String, args: IArgs) {
+    val configFilePath = if (args.useLocalResultDir())
```
adamfilipow92, line 31: This is only cosmetic but i think smth like 
```
    val configFilePath =
        if (args.useLocalResultDir()) Paths.get(args.localResultDir, "performanceMetrics.json")
        else Paths.get(args.localResultDir, resultsDir, "performanceMetrics.json")
```
is little more readable. WDYT?

adamfilipow92:
So in both scenarios ```performanceMetrics.json``` should be downloaded to local storage, right?

test_runner/src/main/kotlin/ftl/reports/api/PerformanceMetrics.kt
piotradamczyk5: changed

piotradamczyk5:
> So in both scenarios `performanceMetrics.json` should be downloaded to local storage, right?

yes

adamfilipow92 (approved):
- [X] Both scenarios tested
Great work! 👍

axelzuziak-gogo (approved):

Sloox (approved):

PR #1316, opened 2020-11-12
test: [IT] Add test cases

pawelpasterz:
Fixes #1313 

This PR introduces a couple of new features
1. New `integrationTests` task in `integration_test` module
1. Few test cases for IT (sanity robo, multiple apks, multiple devices, filtering tests, ignoring failed, cancelling)
1. IT test code -- it basically compares output printed to stdout with expected (`resources/compare` files)

1. Run `./gradlew build` -- to ensure you have flank.jar
1. Run `./gradlew integrationTests`
1. All tests should finish without error (note: it may take up to 15 min to finish)

PR #1342, opened 2020-11-23
ci: Run integration tests on comment or manual trigger

### LINKED ISSUE #1314: [IT] Manual triggers for IT
Integration should have (at least) two triggers:
* `workflow_dispatch` -- clickable, from Actions tab
* comment trigger

piotradamczyk5:
Fixes #1314 

Integration tests could be run manually or by typing `@flank-it` comment

[Test Helper](https://github.com/piotradamczyk5/flank/pull/6)

## Checklist

- [x] Integration tests run on manual trigger
- [x] Integration tests run on comment trigger (`@flank-it`)
- [x] Disable 2+ approve trigger


.github/workflows/integration_tests.yml
```diff
@@ -1,34 +1,47 @@
...
-          echo "Current Approve count ${{ steps.reviews.outputs.approved }}"
+      - uses: khan/pull-request-comment-trigger@master
+        if: ${{ github.event_name == 'issue_comment'}}
+        id: check_issue_comment
+        with:
+          trigger: '@flank-it'
+          reaction: rocket
+        env:
+          GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
+# TEMPORARY DISABLED UNTIL WE FIGURE OUT THE BEST METHOD TO RUN INTEGRATION TESTS
```
Sloox, line 22: Will this comment go into master?  I can see the comment. I am just wondering if this test is for master or only for this branch?

Sloox (changes requested):

.github/workflows/integration_tests.yml
piotradamczyk5: I think that for now, it should be on master until we finish `Improve integration tests` epic

jan-gogo, line 22:
```suggestion
# TEMPORARY DISABLED UNTIL WE FIGURE OUT THE BEST METHOD TO RUN INTEGRATION TESTS
# FIXME BEFORE FINISH EPIC  #1212
```

.github/workflows/integration_tests.yml
```diff
@@ -77,8 +91,8 @@ jobs:
           arguments: "--info :integration_tests:test --tests IntegrationTests.shouldMatchIosSuccessExitCodeAndPattern -Dflank-path=../test_runner/build/libs/flank.jar -Dyml-path=./src/test/resources/flank_ios.yml"
```
pawelpasterz, line 91: This will launch the old integration test only. Once #1316 is done there will be a new Gradle task `integrationTests` that will launch all newly added cases

pawelpasterz (commented):
Small question, I don't know if this is intended or not

pawelpasterz:
Also, when IT triggered manually I think PR checks list won't be updated. Therefore some summary/outcome would be useful for the user. I checked the market and there are some actions for printing test results but I think the easiest would be to add a comment with a build scan URL. 
Current IT (on master) are quite useful (I think). They act like sanity check: - 'oh ok, so flank starts and looks fine'. On the other hand, having all IT in one place and launched together also makes sense.

.github/workflows/integration_tests.yml
piotradamczyk5: So it should be updated in task #1316 
because this task is just about trigger

piotradamczyk5:
> Also, when IT triggered manually I think PR checks list won't be updated. Therefore some summary/outcome would be useful for the user. I checked the market and there are some actions for printing test results but I think the easiest would be to add a comment with a build scan URL.
> Current IT (on master) are quite useful (I think). They act like sanity check: - 'oh ok, so flank starts and looks fine'. On the other hand, having all IT in one place and launched together also makes sense.

Thanks. I will add it

pawelpasterz:
Hm..those changes are not quite what I wanted to achieve BUT it's because of my poor description. I can see some value added within this PR though.
I'll create another task(s) but with a much more detailed description

.github/workflows/integration_tests.yml
pawelpasterz: Please check my comment https://github.com/Flank/flank/pull/1342#issuecomment-732885063
I think we could uncomment this action

piotradamczyk5:
> Hm..those changes are not quite what I wanted to achieve BUT it's because of my poor description. I can see some value added within this PR though.
> I'll create another task(s) but with much more detailed info and definition of done

You could update the description to make this PR fully meet all you criteria

.github/workflows/integration_tests.yml
piotradamczyk5: Could you give more info about this?
From my point of view, we should not require integration tests to pass until we fully finished it

piotradamczyk5:
@flank-it

Sloox (approved):
@pawelpasterz I assume another ticket will address the issue you mentioned before but for now I think the goal was reached.

pawelpasterz:
> @pawelpasterz I assume another ticket will address the issue you mentioned before but for now I think the goal was reached.

That is correct

piotradamczyk5:
@pawelpasterz based on the description of task #1349 this PR should not be merge.

pawelpasterz:
That's why I've created a separate issue. As mentioned earlier, this PR brings value and we can merge it anyway

piotradamczyk5:
> That's why I've created a separate issue. As mentioned earlier, this PR brings value and we can merge it anyway.
> It's up to you

Yeah, but it changed the behaviour of what's you described any way

piotradamczyk5:
As we discussed this PR will be closed and based on it #1349 will be solved

PR #1352, opened 2020-11-25
ci: Make flank release more automatically

### LINKED ISSUE #1346: Make Flank releases more automatically
As a Develop, I want to have more automatic releases so I can make Flank releases faster.

**Describe the solution**
1. Make Flank release at the start of the calendar month automatically(use schedule), but allow to manually run the job to update release notes
2. Push tag automatically after merging release notes PR (use some auto tagger with filters [example](https://github.com/marketplace/actions/github-pr-auto-tagger))

piotradamczyk5:
Fixes #1346 

1. Create release notes job (this which generates documentation and changelog) will run each 1st day of the month
2. After merge Pr with changelog
3. Flank-release bot will push the tag with next version to repository
4. Then release job will run as usual

## Checklist

- [x] Documented
- [x] At schedule to run release job at 1st day of the month
- [x] Create job which will push release tag after merge PR


Sloox (commented):
Can we test this somehow before we go ahead with it? Also https://github.com/Flank/flank/issues/1354 should probably be a blocking issue for this.

piotradamczyk5:
> Can we test this somehow before we go ahead with it? Also #1354 should probably be a blocking issue for this.

Agreed #1354 block release job. However this task is for simplify release process.

I have tested it on my [fork](https://github.com/piotradamczyk5/flank)

.github/workflows/create_new_release.yml
```diff
@@ -0,0 +1,34 @@
+name: Create new release
+
+on:
+  pull_request:
+    types: [ closed ]
+
+jobs:
+  create_release:
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.merged == true && contains(github.event.pull_request.labels.*.name, 'release')
```
pawelpasterz, line 10: This job can be triggered by any merged PR with a specific label (which can be added manually). (I can create PR compromised PR, add release label and merge it. If I understand the job description correctly -- it will start release)

If I am not wrong, release notes PR is created by a bot, I think we should consider adding an additional condition which verifies who is the creator or merged PR.

It's the first that came to my mind, if there is a better way to make it more secure I think it would be worth to implement it.

Of course, let me know, I might be wrong and this implementation is already secure 💪

pawelpasterz (changes requested):

.github/workflows/create_new_release.yml
piotradamczyk5: Yeah, you are 100% right that if someone uses `release` label it will generate new release,
I think that we could make additional check if issue is created by bot and/or check if title starts with `chore: release notes for`

pawelpasterz (approved):

Sloox (approved):

adamfilipow92 (approved):
Yaml looks good, I also check cron expression, and it's valid.

Action should run at
```

07:00 AM, on day 1 of the month

```

Issue #1134, opened 2020-09-21
Integration tests on Windows


Sloox:
As a developer, I want to have integration tests running on Windows workflow so I can guarantee Flank runs across all common operating systems and works as intended.

**Is your feature request related to a problem? Please describe.**
Currently Integration task workflow encounters errors and needs to be fixed.

**Describe the solution you'd like**
Integration tests running on the windows workflow

Sloox:
@bootstraponline @jan-gogo can we confirm that this might not be necessary?

bootstraponline:
@Sloox I think we want to ensure all the integration tests pass on native Windows. This matches the platform support provided by gcloud CLI.

Sloox:
The ticket which has successfully been merged https://github.com/flank/flank/issues/1140 has shown that at a minimum Flank will work with the WSL. The ticket is now blocked by the investigation done into : https://github.com/flank/flank/issues/1172

Sloox:
@piotradamczyk5 or @adamfilipowicz92  is this a duplicate ticket now?

adamfilipow92:
@Sloox I think yes, maybe we should close this ticket.

PR #1385, opened 2020-12-08
ci: Get rid of binaries submodule

### LINKED ISSUE #896: Move binaries from resources to external repo (or other place)

piotradamczyk5:
Fixes #896

This PR delete submodule for [binaries ](https://github.com/Flank/binaries) and download them at runtime from [release page ](https://github.com/Flank/binaries/releases) based on os system

`macOs` -> all should work by default because previously binaries were not copied on macOS
`linux` -> it download binaries from [release page](https://github.com/Flank/binaries/releases) at runtime to `<user_home>/.flank` directory and unpack them 
`windows` -> it works similar to `linux` version, however, full support will be available after #1134 

All tests should work properly

## Checklist

- [x] Documented
- [x] Unit tested
- [x] Downloading binaries for Linux and Windows at runtime
- [x] Create binaries release for Linux and Windows
- [x] Update binaries assets after each change on binaries repo [release page](https://github.com/Flank/binaries/releases) >> [PR](https://github.com/Flank/binaries/pull/4)
- [x] Delete submodule
- [x] Update Github Actions to do not checkout submodules


buildSrc/src/main/kotlin/DownloadBinariesTask.kt
```diff
@@ -0,0 +1,58 @@
...
+
+    @TaskAction
+    fun downloadBinaries() {
+        osName
+            .takeIf { forceUpdate || shouldDownloadBinaries() }
+            ?.let(::downloadAndUnzip)
+    }
+
+    private val osName: String
+        get() = if(isWindows) "windows" else "linux"
```
adamfilipow92, line 29: Maybe add fail if mac?

piotradamczyk5: Basically, it should contain these binaries for running flank.jar on Linux while building on ubuntu

piotradamczyk5: Actually, I figure out that we should also have binaries for windows included. I will make logic to contains download both Windows and Linux when running the build on CI

jan-gogo (commented):
The code looks well, but I would love to have this implementation in the java-library module linked to the test_runner as Gradle dependency, maybe not now, it could be in another PR. 
Some of the code looks a little bit redundant according to flank-scripts, for example, utilities for unzipping & symlinks, probably also could be moved to another library module shared between download-binaries and flank-scripts.

jan-gogo (approved):

piotradamczyk5:
> The code looks well, but I would love to have this implementation in the java-library module linked to the `test_runner` as Gradle dependency, maybe not now, it could be in another PR.
> Some of the code looks a little bit redundant according to flank-scripts, for example, utilities for unzipping & symlinks, probably also could be moved to another library module shared between `download-binaries`(?) and `flank-scripts`.

Yeah I think that we could do some shared module which has common logic for other modules, I will create a ticket for it

docs/binaries_used_in_flank_ios_testing.md
```diff
@@ -0,0 +1,21 @@
...
+They are unpacked to `<user directory>/.flank`.
+If they already exist on this path, they are not downloaded again.
+
+## Updating
+
+In order to update binaries just follow below steps:
+1. checkout binaries [repository](https://github.com/Flank/binaries)
+1. update them using:
+   - `updateBinariesWithFlankBash` will update binaries for Linux and Windows using `flank-scripts`
+   - `update.sh` (old method). It will update binaries for Linux OS
```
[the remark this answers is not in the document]
piotradamczyk5: this is standalone without flank-scripts needed

adamfilipow92 (approved):
- [X] Tested on windows, works!

Issue #1389, opened 2020-12-09
Different exit codes iOS vs android


pawelpasterz:
**Describe the bug**
> A clear and concise description of what the bug is.

When all tests are filtered android run ends with exit code 1 when ios with 2.

**To Reproduce**
> Steps to reproduce the behavior:

1. run ios test with config 
    ```
    gcloud:
      test: [path to zip]
      xctestrun-file: [path to xctestrun]
    flank:
      test-targets:
        - nonExisting/Class
    ```
1. run android test with config 
    ```
    gcloud:
      app: [path to app apk]
      test: [path to test apk]
      test-targets:
        - class non.existing.Class
    flank:
      disable-sharding: true
    ```

**Expected behavior**
> A clear and concise description of what you expected to happen.

Exit code should be the same for both, ios and android. Need to decide which one to use.

**Details (please complete the following information):**
> Have you tested on the latest Flank snapshot?
>
> Post the output of ` flank --version`.

```
version: local_snapshot
revision: 2d2982474e9d38ee39639ed7282b84fb5e33ce01
```

**Additional context**
> Add any other context about the problem here.

https://firebase.google.com/docs/test-lab/android/command-line#script_exit_codes

adamfilipow92:
I think my pr solve this. https://github.com/Flank/flank/pull/1395

Sloox:
@adamfilipow92 will confirm and close this :)

adamfilipow92:
Confirmed, exit codes are the same now

PR #1395, opened 2020-12-11
fix: Print args when iOS tests are not found

### LINKED ISSUE #1388: Output inconsistency (ios vs android)
**Describe the bug**

There are differences in flank output for iOS and Android when all test are filtered
Below is the output for https://github.com/Flank/flank/blob/master/integration_tests/src/test/kotlin/integration/AllTestFilteredIT.kt
* iOS https://gist.github.com/pawelpasterz/e517272563a0bcbf83fb4fdbcf94d549
* android https://gist.github.com/pawelpasterz/5a4d8498629b716abd86cd0b70ee4b2e

**To Reproduce**

1. run ios test with config 
    ```
    gcloud:
      test: [path to zip]
      xctestrun-file: [path to xctestrun]
    flank:
      test-targets:
        - nonExisting/Class
    ```
1. run android test with config 
    ```
    gcloud:
      app: [path to app apk]
      test: [path to test apk]
      test-targets:
        - class non.existing.Class
    flank:
      disable-sharding: true
    ```

**Expected behavior**

There is a decision to be made if flank should (or not) print args.
Based on the answer:
* fix should be implemented
* IT must be updated

>

```
version: local_snapshot
revision: 2d2982474e9d38ee39639ed7282b84fb5e33ce01
```

**Additional context**

Introduced in https://github.com/Flank/flank/pull/1219

adamfilipow92:
Fixes #1388 

1. run ios test with config

```
gcloud:
  test: [path to zip]
  xctestrun-file: [path to xctestrun]
flank:
  test-targets:
    - nonExisting/Class
```

1. Flank should print args before any validation (ios and android)
1. Exit code should be 1
1. Integration tests should works

## Checklist

- [X] Integration tests works

### DIFF
--- a/test_runner/src/main/kotlin/ftl/cli/firebase/test/android/AndroidRunCommand.kt
+++ b/test_runner/src/main/kotlin/ftl/cli/firebase/test/android/AndroidRunCommand.kt
@@ -1,12 +1,12 @@
 package ftl.cli.firebase.test.android
 
 import ftl.args.AndroidArgs
+import ftl.args.logArgs
 import ftl.args.setupLogLevel
 import ftl.args.validate
 import ftl.cli.firebase.test.CommonRunCommand
 import ftl.config.FtlConstants
 import ftl.config.emptyAndroidConfig
-import ftl.log.logLn
 import ftl.mock.MockServer
 import ftl.run.ANDROID_SHARD_FILE
 import ftl.run.dumpShards
@@ -48,7 +48,7 @@ class AndroidRunCommand : CommonRunCommand(), Runnable {
             MockServer.start()
         }
 
-        AndroidArgs.load(Paths.get(configPath), cli = this).logArgs().validate().run {
+        AndroidArgs.load(Paths.get(configPath), cli = this).setupLogLevel().logArgs().validate().run {
             runBlocking {
                 if (dumpShards) dumpShards()
                 else newTestRun()
@@ -63,7 +63,3 @@ class AndroidRunCommand : CommonRunCommand(), Runnable {
     var dumpShards: Boolean = false
 }
 
-private fun AndroidArgs.logArgs() = also {
-    setupLogLevel()
-    logLn(this)
-}
--- a/test_runner/src/main/kotlin/ftl/cli/firebase/test/android/AndroidRunCommand.kt
+++ b/test_runner/src/main/kotlin/ftl/cli/firebase/test/android/AndroidRunCommand.kt
@@ -1,12 +1,12 @@
 package ftl.cli.firebase.test.android
 
 import ftl.args.AndroidArgs
-import ftl.args.logArgs
 import ftl.args.setupLogLevel
 import ftl.args.validate
 import ftl.cli.firebase.test.CommonRunCommand
 import ftl.config.FtlConstants
 import ftl.config.emptyAndroidConfig
+import ftl.log.logLn
 import ftl.mock.MockServer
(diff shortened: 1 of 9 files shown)


jan-gogo (changes requested):
I don't like the validation is split between execution steps. Wouldn't be better to print args before any validation instead of breaking validation consistency?

adamfilipow92:
> I don't like the validation is split between execution steps. Wouldn't be better to print args before any validation instead of breaking validation consistency?
> 
> WDYT? @pawelpasterz @piotradamczyk5 @adamfilipow92

Agree

pawelpasterz:
> I don't like the validation is split between execution steps. Wouldn't be better to print args before any validation instead of breaking validation consistency?
> 
> WDYT? @pawelpasterz @piotradamczyk5 @adamfilipow92

+1

I think printing args (always) is a good idea.

test_runner/src/main/kotlin/ftl/cli/firebase/test/android/AndroidRunCommand.kt
```diff
@@ -47,8 +48,7 @@ class AndroidRunCommand : CommonRunCommand(), Runnable {
             MockServer.start()
         }
 
-        AndroidArgs.load(Paths.get(configPath), cli = this).validate().run {
-            setupLogLevel()
+        AndroidArgs.load(Paths.get(configPath), cli = this).logArgs().validate().run {
```
jan-gogo, line 51: 👍

test_runner/src/main/kotlin/ftl/cli/firebase/test/ios/IosRunCommand.kt
```diff
@@ -60,3 +60,8 @@ class IosRunCommand : CommonRunCommand(), Runnable {
     )
     var dumpShards: Boolean = false
 }
+
+private fun IosArgs.logArgs() = also {
+    setupLogLevel()
```
jan-gogo: `setupLogLevel` looks like an independent operation not only related to args printing. IMO it should be called before `IosArgs.load`.

test_runner/src/main/kotlin/ftl/cli/firebase/test/android/AndroidRunCommand.kt
```diff
@@ -62,3 +62,8 @@ class AndroidRunCommand : CommonRunCommand(), Runnable {
     )
     var dumpShards: Boolean = false
 }
+
+private fun AndroidArgs.logArgs() = also {
+    setupLogLevel()
```
jan-gogo: Same here

test_runner/src/main/kotlin/ftl/cli/firebase/test/ios/IosRunCommand.kt
adamfilipow92: Changed! Thanks!

test_runner/src/main/kotlin/ftl/cli/firebase/test/android/AndroidRunCommand.kt
adamfilipow92: Changed! Thanks!

pawelpasterz (approved):

Sloox (approved):
It appears it will fix https://github.com/Flank/flank/issues/1389

cryptoyang (approved):

jan-gogo (approved):

piotradamczyk5 (approved):

PR #1409, opened 2020-12-17
ci: trigger integration tests from pr comment

### LINKED ISSUE #1349: [IT] Comment trigger for full suite integration tests
PR #1316 introduced a new Gradle task launching full suite integration tests `./gradlew integrationTests`. Those are in addition to already implemented IT, not a replacement. Current IT (let's call them sanity IT) do their job as quick check if flank's jar works correctly and nothing unplanned causes flank to crash with a simple run. Sanity ITs are also cheap. 
Given mentioned I think it's fine to run them automatically.

Full suite IT requires several times more devices (both virtual and physical). `flank-open-source` project has a very high quota limit, yet it's not unlimited. With full suite IT running automatically (similar to sanity IT) we could burn daily quota very quickly, as result PR's would be blocked and general work on repo would be difficult, at least until the end of the day.

#### Requirements for trigger (my proposals in parenthesis):
1. GH action should run ONLY full suite ITs (separated flow/gh action?)
1. we can use a similar trigger that was implemented in #1342 (@flank-it-fs)
1. IT results should be presented to the user, there are some possibilities:
    - PR's checklist (might be tough due to missing PR context for issue comment event)
    - update/change PR description
    - add a comment (separate for each trigger invocation)
1. consider the form of presented results: text, a summary table, etc (I think the easiest would be to print gradle build scan URL, similar to ubuntu-workflow)
1. consider a limitation of who can trigger full suite IT
1. instant feedback, we should inform a user about the current status of IT/action. (This can be achieved via reactions), for instance:
    - :eyes: `noticed`- gh action was triggered 
    - :rocket: `IT started` - full suite ITs were started and are in progress
    -  🎉   - all tests finished successfully
    - :-1: - there are some failed tests
1. sanity IT (currently running) should remain
1. Add/update PR's check (need investigations if it is possible)

#### Additional info:
* as mentioned earlier issue comment is missing PR context (it's due to the fact GH considers PRs as normal issues)
* we don't need to implement super fancy flow/gh action from the beginning, Let's start with the easiest implementation that fulfills requirements

piotradamczyk5:
Fixes #1349

Current flows:
1. Schedule and manual trigger
- should cancel other builds in progress
- should run full integration tests on master branch
- should process results using flank scripts
2. Comment trigger
- should get pr number
- should post an issue comment
- should run full integration tests on PR branch
- should update comment based on test results

Integration tests run:
1. Integration tests build flank when the environmental variable `CI` is set to true
2. Integration tests do not build flank locally( when variable `CI` is set to false)

Comment trigger
1. Integration tests full suite is triggered by posting comment `@flank-it`
2. When Github action started 👀  are displayed under user comment
image
3. When Integration tests started comment is posted with a link to the integration test and timestamp (and a 🚀  )
image
4. After tests finished, comment from the previous step is replaced to result description and reaction(👍 :success, 👎 :failure)
image

[Fork playground](https://github.com/piotradamczyk5/flank/pull/2)
- integration tests failed, because of missing service account

## Checklist

- [x] Add comment trigger
- [x] Show status on PR
- [x] Integration tests should depend on build for CI builds

### DIFF
--- a/integration_tests/build.gradle.kts
+++ b/integration_tests/build.gradle.kts
@@ -72,3 +72,8 @@ val compileTestKotlin: KotlinCompile by tasks
 compileTestKotlin.kotlinOptions {
     freeCompilerArgs = listOf("-Xinline-classes")
 }
+// Locally you should have flank builded on CI we need to build it
+if(System.getenv("CI") != null) {
+    tasks["integrationTests"].dependsOn(":test_runner:build")
+}
+
--- a/integration_tests/build.gradle.kts
+++ b/integration_tests/build.gradle.kts
@@ -72,8 +72,7 @@ val compileTestKotlin: KotlinCompile by tasks
 compileTestKotlin.kotlinOptions {
     freeCompilerArgs = listOf("-Xinline-classes")
 }
-// Locally you should have flank builded on CI we need to build it
+// Locally you should have flank built already, on CI we need to build it
 if(System.getenv("CI") != null) {
     tasks["integrationTests"].dependsOn(":test_runner:build")
 }
-
--- a/.github/workflows/full_suite_integration_tests.yml
+++ b/.github/workflows/full_suite_integration_tests.yml
@@ -4,10 +4,61 @@ on:
   schedule:
     - cron: '0 0 * * 1-5' # At 00:00 on every day-of-week from Monday through Friday
   workflow_dispatch:      # or manually
+  issue_comment:
+    types: [created]      # or by comment
 
 jobs:
+  should_run_it:
+    runs-on: ubuntu-latest
+    outputs:
+      run_integration_tests: ${{ steps.should-run-step.outputs.should_run }}
+      pr_number: ${{ steps.pr_number.outputs.number }}
+    steps:
+      - name: Get PR number
+        id: pr_number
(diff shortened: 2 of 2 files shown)


integration_tests/build.gradle.kts
```diff
@@ -72,3 +72,8 @@ val compileTestKotlin: KotlinCompile by tasks
 compileTestKotlin.kotlinOptions {
     freeCompilerArgs = listOf("-Xinline-classes")
 }
+// Locally you should have flank builded on CI we need to build it
```
bootstraponline, line 75:
```suggestion
// Locally you should have flank built already, on CI we need to build it
```

Sloox (approved):
I think its fine.

Sloox:
@flank-it

piotradamczyk5:
> @flank-it

Not this time @Sloox :D must be merged first 
it works on  [Fork playground](https://github.com/piotradamczyk5/flank/pull/2)

adamfilipow92 (approved):
- [X] Build run befeore integration tests
- [X] Integration tests work as expected (some test fails but it fixed here: https://github.com/Flank/flank/pull/1410)
https://github.com/piotradamczyk5/flank/runs/1571299555?check_suite_focus=true

PR #1414, opened 2020-12-18
fix: Add attempts to performance test

jan-gogo:
This is another fix for the test issue https://github.com/Flank/flank/runs/1576759651?check_suite_focus=true, |
because the previous one (https://github.com/Flank/flank/pull/1413) was not enough.

The UT `performance calculateShardsByTime` seems to be flaky because the results depend on CPU performance and usage. 

My proposition is to rerun shards calculation if run time in ms grater than expected or attempt number not reached max value.

### DIFF
--- a/test_runner/src/test/kotlin/ftl/shard/ShardTest.kt
+++ b/test_runner/src/test/kotlin/ftl/shard/ShardTest.kt
@@ -19,8 +19,6 @@ import org.junit.Assert.assertEquals
 import org.junit.Assert.assertTrue
 import org.junit.Test
 import org.junit.runner.RunWith
-import java.util.concurrent.TimeUnit
-import kotlin.system.measureNanoTime
 import kotlin.system.measureTimeMillis
 
 @RunWith(FlankTestRunner::class)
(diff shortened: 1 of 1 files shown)


test_runner/src/test/kotlin/ftl/shard/ShardTest.kt
```diff
@@ -102,13 +103,20 @@ class ShardTest {
...
+        val maxAttempts = 5
+        val maxMs = 5000
+
+        var attempt = 0
+        var ms = Long.MAX_VALUE
 
-        val ms = TimeUnit.NANOSECONDS.toMillis(nano)
-        println("Shards calculated in $ms ms")
-        assertThat(ms).isLessThan(5000)
+        while (attempt < maxAttempts && ms >= maxMs) {
```
pawelpasterz, line 112:
```suggestion
        while (attempt++ < maxAttempts && ms >= maxMs) {
```

pawelpasterz: Then we can remove line below

test_runner/src/test/kotlin/ftl/shard/ShardTest.kt
```diff
@@ -21,6 +21,7 @@ import org.junit.Test
 import org.junit.runner.RunWith
 import java.util.concurrent.TimeUnit
 import kotlin.system.measureNanoTime
```
pawelpasterz, line 23:
```suggestion

```

test_runner/src/test/kotlin/ftl/shard/ShardTest.kt
```diff
@@ -102,13 +103,20 @@ class ShardTest {
             copy(commonArgs = commonArgs.copy(maxTestShards = 4))
         }
```
pawelpasterz, line 105: Maybe not exactly the same but something similar
```suggestion
// Although the test passes locally, it occasionally fails on VM's used in GH actions
// Therefore we need a retry approach to verify flakiness
```

pawelpasterz (approved):
Just small nits.

PR #1417, opened 2020-12-21
refactor: Update firebase_apis module

pawelpasterz:
Changes: 
* update google api client
* move client dependency to the common source of truth, which will allow to update client automatically

No new logic introduced. Normal build + integration tests should be enough (will be handled by GH actions)


pawelpasterz:
@flank-it

jan-gogo (approved):

piotradamczyk5 (approved):

PR #1421, opened 2020-12-23
feat: Default project Id - use GOOGLE_APPLICATION_CREDENTIALS first, then GOOGLE_CLOUD_PROJECT

### LINKED ISSUE #1407: Flank doesn't autodetect projectId from JSON file on CI (Buildkite if it matters)
**Describe the bug**
This is a weird one...was using fladle to set this up in our Android gradle project and we use a JSON file for the service worker.  Running "./gradlew runFlank" locally worked fine.  But on the CI system, it had permissions errors:
```

Caused by: com.google.api.client.googleapis.json.GoogleJsonResponseException: 403 Forbidden
--
  | POST https://toolresults.googleapis.com/toolresults/v1beta3/projects/wf-gcp-us-testenablement-prod:initializeSettings
  | {
  | "code" : 403,
  | "errors" : [ {
  | "domain" : "global",
  | "message" : "The caller does not have permission",
  | "reason" : "forbidden"
  | } ],
  | "message" : "The caller does not have permission",
  | "status" : "PERMISSION_DENIED"
  | }


```

With help from the Fladle plugin author, we found that this was because projectId wasn't specified in the fladle config because Flank should automatically grab it from the JSON keyfile.

What's weird is that this works fine when run locally (tested on multiple developers' macbooks)

**To Reproduce**

set up a project with minimal fladle config:
```
apply plugin: 'com.osacky.fladle'

fladle {
    serviceAccountCredentials = project.layout.projectDirectory.file('../ftl-dev.json')
    variant = "debug"
    devices = [
            // run on minSdk and targetSdk devices
            [ "model": "NexusLowRes", "version": "29" ],
            [ "model": "Nexus5", "version": "23" ]
    ]
}

```
Run locally with "gradle runFlank" and it works fine.
Do the same on CI and you'll see the permissions error.

Add projectId and it works fine.

**Expected behavior**
Should not need to specify projectId...it should be able to read this from the JSON file.

```
> Task :app:execFlank
--
  | version: v20.09.3
  | revision: 06454ae331697f57aa682b0da273fe8ab8a0e823
```

**Additional context**
Fladle author said Flank should autodetect properly.  It should also log what projectId it's trying to use or if it can't autodetect one from the JSON file.

pawelpasterz:
Fixes #1407 

Flank will now use `GOOGLE_APPLICATION_CREDENTIALS` env first, to parse and get project id. (previously the first one was `GOOGLE_CLOUD_PROJECT`

1. build flank
1. remember to use the fat jar to launch test runs `java -jar ./test_runner/build/libs/flank.jar`
1. create dummy `echo '{"project_id":"any funny project id here"}' >> test.json`
1. export variable `export GOOGLE_APPLICATION_CREDENTIALS="$(pwd)/test.json"`
1. export variable `export GOOGLE_CLOUD_PROJECT="another project id"`
1. run any flank test run (within the same terminal window)
1. If you are creative enough you should get 404 (interesting fact: I used `ddd` project id during local tests and it turned out there is such project, I received 403 instead 😄 ) with a message 
    1. 404 -> `Flank was unable to find project [id from step 4.]`
    2. 403 -> `Flank encountered a 403 error when running on project [id from 2.] (from [path to file from 2.])`
1. clear `export GOOGLE_APPLICATION_CREDENTIALS=`
1. similar to 7. except project id should be from 5
1. clear `export GOOGLE_CLOUD_PROJECT=`
1. run tests, flank should proceed normally

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
+++ b/test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
@@ -30,6 +30,7 @@ import ftl.shard.createShardsByShardCount
 import ftl.shard.shardCountByTime
 import ftl.util.FlankTestMethod
 import ftl.util.assertNotEmpty
+import ftl.util.getGACPathOrEmpty
 import java.io.File
 import java.net.URI
 import java.nio.file.Files
@@ -43,6 +44,8 @@ object ArgsHelper {
     val yamlMapper: ObjectMapper by lazy {
         YamlObjectMapper().registerKotlinModule()
     }
+
+    private var projectIdSource: String? = null
 
     fun assertFileExists(file: String, name: String) {
         if (!file.exist()) throw FlankGeneralError("'$file' $name doesn't exist")
@@ -137,7 +140,7 @@ object ArgsHelper {
     // Make best effort to list/create the bucket.
     // Due to permission issues, the user may not be able to list or create buckets.
     fun createGcsBucket(projectId: String, bucket: String): String {
-        if (bucket.isBlank()) return GcToolResults.getDefaultBucket(projectId)
+        if (bucket.isBlank()) return GcToolResults.getDefaultBucket(projectId, projectIdSource)
             ?: throw FlankGeneralError("Failed to make bucket for $projectId")
         if (useMock) return bucket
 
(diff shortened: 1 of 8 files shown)


pawelpasterz:
@flank-it

test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
```diff
@@ -178,29 +181,30 @@ object ArgsHelper {
...
-
-        // Allow users control over project by checking using Google's logic first before falling back to JSON.
-        return ServiceOptions.getDefaultProjectId() ?: serviceAccountProjectId()
-    }
+    fun getDefaultProjectId(): String = if (useMock) "mockProjectId"
+    // Allow users control over project by checking using Google's logic first before falling back to JSON.
+    else fromUserProvidedCredentials()
+        ?: ServiceOptions.getDefaultProjectId()?.let { if (it.isBlank()) null else it }
+        ?: fromDefaultCredentials()
+        ?: ""
```
Sloox, line 189: orEmpty?

Sloox (approved):

test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
pawelpasterz: That would require wrapping with parenthesis or usage of scope functions which, IMHO, is less readable.
But I'll change `getDefaultProjectId` signature to return `null` which seems more reasonable

test_runner/src/main/kotlin/ftl/util/Utils.kt
```diff
@@ -102,3 +102,5 @@ fun <T> KMutableProperty<T?>.require() =
     getter.call() ?: throw FlankGeneralError(
         "Invalid value for [${setter.annotations.filterIsInstance<JsonProperty>().first().value}]: no argument value found"
     )
+
+fun getGACPathOrNull(): String? = System.getenv("GOOGLE_APPLICATION_CREDENTIALS")
```
piotradamczyk5, line 106: Basically, it does not return path, but value of env variable

test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
```diff
@@ -44,6 +45,8 @@ object ArgsHelper {
         YamlObjectMapper().registerKotlinModule()
     }
 
+    private var projectIdSource: String? = null
```
piotradamczyk5, line 48: `projectIdSource` -> `projectId`

piotradamczyk5: and maybe we should move it to `FtlContstans` or make it lazy withing this file

test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
```diff
@@ -178,29 +181,29 @@ object ArgsHelper {
...
+    fun getDefaultProjectIdOrNull(): String? = if (useMock) "mockProjectId"
+    // Allow users control over project by checking using Google's logic first before falling back to JSON.
+    else fromUserProvidedCredentials()
+        ?: ServiceOptions.getDefaultProjectId()?.let { if (it.isBlank()) null else it }
+        ?: fromDefaultCredentials()
+
+    private fun fromDefaultCredentials() = getProjectIdFromJson(defaultCredentialPath)
+
+    private fun fromUserProvidedCredentials() =
+        getProjectIdFromJson(Paths.get(getGACPathOrNull() ?: ""))
```
piotradamczyk5, line 193: `getGACPathOrNull().orEmpty()` ?

test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
pawelpasterz: Unfortunately not, it's actually the path where project id was read from.

pawelpasterz: > and maybe we should move it to FtlContstans 

It's not used anywhere outside of this class. That would create unnecessary couplings

> or make it lazy val inside this file

`null` is desired here. `projectIdSource` is used when error message is created. Since we do not control where `ServiceOptions.getDefaultProjectId()` found (or from where) project id we should not add `(from ....)` string in that case. I don't know, how `by lazy { }` could be implemented in a reasonable way to provide expected functionality. `lateinit` could work then we would always assign a value (even null) otherwise we would have `UninitializedPropertyAccessException` thrown.

test_runner/src/main/kotlin/ftl/util/Utils.kt
pawelpasterz: True, but I wanted to be explicit that path (as a string) is the expected value here

piotradamczyk5 (approved):

PR #1439, opened 2020-12-31
ci: Remove sanity checks and change full suite IT OS to macOS

piotradamczyk5:
1. The old integration tests are not needed anymore because we have daily/comment/manual trigger for full suite. They just consume our quote, so they were removed
2. Full suite integration tests will use `macos-latest` instead of `ubuntu-latest` to prevent occurring issue with sharding performance

IT works as previous

## Checklist

- [x] Full suite IT OS changed from Ubuntu to MacOS
- [x] Sanity check removed


Sloox (approved):
Looks fine to me. 
For that sharding issue can we not just increase the time for sharding if its linux or something similar? @piotradamczyk5

adamfilipow92 (approved):
Looks good 👍

piotradamczyk5:
> Looks fine to me.
> For that sharding issue can we not just increase the time for sharding if its linux or something similar? @piotradamczyk5

I am thinking about adding retry when CI==true

PR #1448, opened 2021-01-04
feat: Filtering test configurations (iOS TestPlans)

### LINKED ISSUE #1379: iOS: Support for filtering Test Configurations
Currently Flank supports .xctestrun file from Test Plan #685, but user is not able to specify which test configuration Flank should run. 
For example if .xctestrun file contains 2 configurations Flank will run both of them.

FTL doesn't support specifying test configuration yet, so Flank needs to handle this on his side.

Example: We have .xctestrun file containing two test configurations: *pl* and *en*:

```
{
  "en": [ // test configuration: en
    [
      "UITestsClass/test1_ENLocale"
    ]
  ],
  "pl": [ // test configuration: pl
    [
      "UITestsClass/test1_ENLocale"
    ]
  ]
}
```

### Actual result:
Flank will generate two .xctestrun files for each configuration and run two separate test runs on FTL.

### Expected result:
User should be able to specify which configuration Flank should run by passing argument: (proposal - to be discussed)
```
 -only-test-configuration: en
```
or
```
-skip-test-configuration: pl
```

___
Proposed argument names are from actual `xcodebuild test-without-building` command. Reference:
`man xcodebuild`

>      -skip-test-configuration  test-configuration-name,
        -only-test-configuration test-configuration-name
           Constrain      test     configurations     in     test     actions.
           -only-test-configuration constrains a test action to  only  test  a
           specified  test  configuration  within a test plan, and exclude all
           other test configurations.  -skip-test-configuration  constrains  a
           test action to skip a specified test configuration, but include all
           other test configurations. Each test configuration name must  match
           the  name  of a configuration specified in a test plan and is case-
           sensitive. An xcodebuild command can  combine  multiple  constraint
           options,   but   -only-test-configuration   has   precedence   over
           -skip-test-configuration.

axelzuziak-gogo:
Fixes #1379 

This PR adds to new options for flank: 
- `skip-test-configuration`
- `only-test-configuration`

Flank users can filter test configurations (inside of a test plan).

From project root:
```
. .env
cd ./test_runner/
flankScripts shell buildFlank 
flank ios run -c ./src/test/kotlin/ftl/fixtures/ios_test_plan.yml
```
additionally, edit `os_test_plan.yml` for testing different configurations.

## Checklist

- [x] Documented - updated main docs 
- [x] Unit tested

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -21,9 +21,7 @@ plugins {
 tasks {
     "lintKotlinMain"(LintTask::class) {
         exclude(
-            "**/*Generated.kt",
-            "**/*Test.kt",
-            "**/Test*.kt" // we can expand this list
+            "**/*Generated.kt" // we can expand this list
         )
     }
 }
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -21,7 +21,9 @@ plugins {
 tasks {
     "lintKotlinMain"(LintTask::class) {
         exclude(
-            "**/*Generated.kt" // we can expand this list
+            "**/*Generated.kt",
+            "**/*Test.kt",	
+            "**/Test*.kt" // we can expand this list
         )
     }
 }
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -22,7 +22,7 @@ tasks {
     "lintKotlinMain"(LintTask::class) {
         exclude(
             "**/*Generated.kt",
-            "**/*Test.kt",	
+            "**/*Test.kt",
             "**/Test*.kt" // we can expand this list
         )
     }
--- a/test_runner/src/main/kotlin/ftl/args/CreateIosArgs.kt
+++ b/test_runner/src/main/kotlin/ftl/args/CreateIosArgs.kt
@@ -33,7 +33,9 @@ private fun createIosArgs(
     testTargets = flank.testTargets?.filterNotNull().orEmpty(),
     obfuscateDumpShards = obfuscate,
     app = gcloud.app?.normalizeFilePath().orEmpty(),
-    testSpecialEntitlements = gcloud.testSpecialEntitlements ?: false
+    testSpecialEntitlements = gcloud.testSpecialEntitlements ?: false,
+    onlyTestConfiguration = flank.onlyTestConfiguration,
+    skipTestConfiguration = flank.skipTestConfiguration
 )
 
 private fun convertToShardCount(inputValue: Int) =
--- a/test_runner/src/main/kotlin/ftl/args/CreateIosArgs.kt
+++ b/test_runner/src/main/kotlin/ftl/args/CreateIosArgs.kt
@@ -34,8 +34,8 @@ private fun createIosArgs(
     obfuscateDumpShards = obfuscate,
     app = gcloud.app?.normalizeFilePath().orEmpty(),
     testSpecialEntitlements = gcloud.testSpecialEntitlements ?: false,
-    onlyTestConfiguration = flank.onlyTestConfiguration,
-    skipTestConfiguration = flank.skipTestConfiguration
+    onlyTestConfiguration = flank.onlyTestConfiguration.orEmpty(),
+    skipTestConfiguration = flank.skipTestConfiguration.orEmpty()
 )
 
 private fun convertToShardCount(inputValue: Int) =
--- a/test_runner/src/main/kotlin/ftl/args/ValidateIosArgs.kt
+++ b/test_runner/src/main/kotlin/ftl/args/ValidateIosArgs.kt
@@ -102,9 +102,12 @@ private fun IosArgs.validType() {
 }
 
 private fun IosArgs.assertXcTestRunVersion() {
-    if (onlyTestConfiguration.isNotBlank() or skipTestConfiguration.isNotBlank())
-        if (xcTestRunData.version == XcTestRunVersion.V1) throw FlankConfigurationError("Specified [xctestrun-file] doesn't contain test plans. Options: [only-test-configuration] or [skip-test-configuration] are not valid for this [xctestrun-file]")
+    if (filterTestConfiguration && xcTestRunData.version == XcTestRunVersion.V1)
+        throw FlankConfigurationError("Specified [xctestrun-file] doesn't contain test plans. Options: [only-test-configuration] or [skip-test-configuration] are not valid for this [xctestrun-file]")
 }
+
+private val IosArgs.filterTestConfiguration
+    get() = onlyTestConfiguration.isNotBlank() or skipTestConfiguration.isNotBlank()
 
 private fun IosArgs.assertXcTestRunData() =
     takeIf { isXcTest && !disableSharding && testTargets.isNotEmpty() }
--- a/test_runner/src/main/kotlin/ftl/args/ValidateIosArgs.kt
+++ b/test_runner/src/main/kotlin/ftl/args/ValidateIosArgs.kt
@@ -5,6 +5,7 @@ import ftl.args.yml.Type
 import ftl.ios.IosCatalog
 import ftl.ios.IosCatalog.getSupportedVersionId
 import ftl.ios.xctest.XcTestRunData
+import ftl.ios.xctest.common.XcTestRunVersion
 import ftl.ios.xctest.common.mapToRegex
 import ftl.run.exception.FlankConfigurationError
 import ftl.run.exception.FlankGeneralError
@@ -20,6 +21,7 @@ fun IosArgs.validate() = apply {
     checkResultsDirUnique()
     assertAdditionalIpas()
     validType()
+    assertXcTestRunVersion()
     assertGameloop()
     assertXcTestRunData()
 }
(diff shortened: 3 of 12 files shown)


test_runner/src/main/kotlin/ftl/config/ios/IosFlankConfig.kt
```diff
@@ -25,16 +25,42 @@ data class IosFlankConfig @JsonIgnore constructor(
...
+    )
+    @set:JsonProperty("only-test-configuration")
+    var onlyTestConfiguration: String? by data
+
+    @set:CommandLine.Option(
+        names = ["--skip-test-configuration"],
+        description = [
+            "Constrains a test action to skip a specified test configuration " +
+                    "and include all other test configurations. (default: run all test configurations). " +
+                    "Flank can combine  multiple constraint options, but -only-test-configuration has precedence over -skip-test-configuration. " +
```
jan-gogo, line 45: IMO those two options are opposite and should be mutually exclusive on validation.

test_runner/src/main/kotlin/ftl/config/ios/IosFlankConfig.kt
```diff
@@ -25,16 +25,42 @@ data class IosFlankConfig @JsonIgnore constructor(
...
+                    "Each test configuration name must match the name of a configuration specified in a test plan and is case-sensitive."
+        ]
+    )
+    @set:JsonProperty("skip-test-configuration")
+    var skipTestConfiguration: String? by data
```
jan-gogo, line 50: Isn't be better to allow a list of configurations instead of just one?
In other words:
As a user, If I have a collection of configuration, I would like to skip part of them, not only one.

axelzuziak-gogo: You're right IMO it's better to allow list of configurations. We can change that. 
But, I tried to match Xcode's interface - if you try to run iOS tests from command line, xcodebuild tool is exposing those two parameters: skip-test-configuration, only-test-configuration - that's why I introduced them here

jan-gogo: OK, If iOS developers are familiar with those options we should keep convention.

piotradamczyk5:
@Mergifyio rebase

test_runner/src/main/kotlin/ftl/ios/xctest/XcTestData.kt
```diff
@@ -54,15 +54,23 @@ private fun emptyXcTestRunData() = XcTestRunData(
     nsDict = NSDictionary(),
     version = V1
 )
+private fun IosArgs.filterTestConfigurationsIfNeeded(
+    configurations: Map<String, Map<String, List<String>>>
+): Map<String, Map<String, List<String>>> {
+    val only = onlyTestConfiguration ?: return configurations.filterKeys { it != skipTestConfiguration }
+    return configurations.filterKeys { it == only }
+}
```
jan-gogo, line 62:
```suggestion
private fun IosArgs.filterTestConfigurationsIfNeeded(
    configurations: Map<String, Map<String, List<String>>>
): Map<String, Map<String, List<String>>> = when {
    onlyTestConfiguration != null -> configurations.filterKeys { it == onlyTestConfiguration }
    skipTestConfiguration != null -> configurations.filterKeys { it != skipTestConfiguration }
    else -> configurations
}
```

In most cases, even one return is too much :).

jan-gogo (commented):
Implementation looks great 👍

axelzuziak-gogo:

piotradamczyk5 (commented):
Please update `flank.ios.yml`

test_runner/src/main/kotlin/ftl/args/ValidateIosArgs.kt
```diff
@@ -99,6 +101,11 @@ private fun IosArgs.validType() {
         throw FlankConfigurationError("Type should be one of ${validIosTypes.joinToString(",")}")
 }
 
+private fun IosArgs.assertXcTestRunVersion() {
+    if (!onlyTestConfiguration.isNullOrEmpty() or !skipTestConfiguration.isNullOrEmpty())
+        if (xcTestRunData.version == XcTestRunVersion.V1) throw FlankConfigurationError("Specified [xctestrun-file] doesn't contain test plans. Options: [only-test-configuration] or [skip-test-configuration] are not valid for this [xctestrun-file]")
+}
```
piotradamczyk5, line 107: I think that you could write it in single if

test_runner/src/main/kotlin/ftl/args/CreateIosArgs.kt
```diff
@@ -33,7 +33,9 @@ private fun createIosArgs(
     testTargets = flank.testTargets?.filterNotNull().orEmpty(),
     obfuscateDumpShards = obfuscate,
     app = gcloud.app?.normalizeFilePath().orEmpty(),
-    testSpecialEntitlements = gcloud.testSpecialEntitlements ?: false
+    testSpecialEntitlements = gcloud.testSpecialEntitlements ?: false,
+    onlyTestConfiguration = flank.onlyTestConfiguration,
```
Sloox, line 37: should we not have an .orEmpty here?

Sloox (changes requested):

pawelpasterz:
Please provide a detailed Test Plan. Thanks! 👍

test_runner/src/main/kotlin/ftl/args/ValidateIosArgs.kt
jan-gogo: IMO complex conditions are always not convenient for reading. What is the difference between 

```
private fun IosArgs.assertXcTestRunVersion() {
    if (!onlyTestConfiguration.isNullOrEmpty() or !skipTestConfiguration.isNullOrEmpty())
        if (xcTestRunData.version == XcTestRunVersion.V1) throw FlankConfigurationError(
            "Specified [xctestrun-file] doesn't contain test plans. Options: [only-test-configuration] or [skip-test-configuration] are not valid for this [xctestrun-file]"
        )
}
```

and

```
private fun IosArgs.assertXcTestRunVersion2() {
    if ((!onlyTestConfiguration.isNullOrEmpty() or !skipTestConfiguration.isNullOrEmpty())
        and xcTestRunData.version == XcTestRunVersion.V1) throw FlankConfigurationError(
            "Specified [xctestrun-file] doesn't contain test plans. Options: [only-test-configuration] or [skip-test-configuration] are not valid for this [xctestrun-file]"
        )
}

It almost the same ;P
```

test_runner/src/main/kotlin/ftl/args/CreateIosArgs.kt
jan-gogo: Agreed, empty and null means the same in this context, so we don't need both, so `onlyTestConfiguration` and `skipTestConfiguration` in IosArgs shouldn't be nullable.

test_runner/src/main/kotlin/ftl/args/ValidateIosArgs.kt
piotradamczyk5: so my proposition is to move complex condition into a separate function, IMO nested ifs are not so readable

jan-gogo:

build.gradle.kts
```diff
@@ -21,9 +21,7 @@ plugins {
 tasks {
     "lintKotlinMain"(LintTask::class) {
         exclude(
-            "**/*Generated.kt",
-            "**/*Test.kt",
-            "**/Test*.kt" // we can expand this list
+            "**/*Generated.kt" // we can expand this list
```
piotradamczyk5, line 24:
```suggestion
            "**/*Generated.kt",
            "**/*Test.kt",	
            "**/Test*.kt" // we can expand this list
```

jan-gogo:
**Test results**
* [only-test-configuration](https://gist.github.com/jan-gogo/051d36ad6327d05b5ec0df4d7c0cb03e)
* [skip-test-configuration](https://gist.github.com/jan-gogo/e022826dc6003e82050638c7561d2c9b)
* [both](https://gist.github.com/jan-gogo/26a0219875915a8a351c282c15a77658)
* [none](https://gist.github.com/jan-gogo/1f5dccc2c26923ec81f3b6613ed8a749)

IMO looks ok. @axelzuziak-gogo pls verify the results.

Sloox (approved):
LGTM

piotradamczyk5 (changes requested):
Still valid comment about missing documentation in `flank.ios.yml`

build.gradle.kts
```diff
@@ -22,7 +22,7 @@ tasks {
     "lintKotlinMain"(LintTask::class) {
         exclude(
             "**/*Generated.kt",
-            "**/*Test.kt",
+            "**/*Test.kt",
```
adamfilipow92, line 25: Do we need this spacing?

adamfilipow92 (commented):
-[X] tested with default [```ios_test_plan.yml```](https://gist.github.com/adamfilipow92/24f24b82c7ea60d88b6cc309ac780ba0)
- [X] Code looks good

Pls remove this spacing and update docs

build.gradle.kts
jan-gogo: Nope, probably it's a mistake.

build.gradle.kts
```diff
@@ -21,7 +21,9 @@ plugins {
 tasks {
     "lintKotlinMain"(LintTask::class) {
         exclude(
-            "**/*Generated.kt" // we can expand this list
+            "**/*Generated.kt",
+            "**/*Test.kt",
+            "**/Test*.kt" // we can expand this list
```
adamfilipow92, line 26: I think we enable linter in this pr: #1515 
https://github.com/Flank/flank/commit/88b6c52a5e10cee38ce5931e7d9859fa3c1a70f8

jan-gogo: KK, something goes wrong during rebase

adamfilipow92 (approved):
👍

piotradamczyk5 (approved):

pawelpasterz:
IT are not updated with new options...

PR #1468, opened 2021-01-07
ci: Use pointer workflows to preserve run date

pawelpasterz:
Due to GH API limitation workflow is unable to fetch info about itself. Dates of previous runs are used in scripts. An empty array is returned instead.
As a workaround to overcome those limitations two new workflows are introduced.
Their intention is "just to be triggered" and therefore preserve the date which is used in dependencies and IT workflows.

There is no easy way to test the implementation. This PR introduced 2 new workflows which can be triggered after the merge to master.

### DIFF
--- a/.github/workflows/integration_tests_pointer.yml
+++ b/.github/workflows/integration_tests_pointer.yml
@@ -1,0 +1,13 @@
+# Workflow is unable to retrieve data about previous workflows (the same file).
+# This workflow is a way to preserve last run date for full_suite_integration_tests.yml workflow
+# which is available from within that workflow
+name: Integration Tests Pointer
+on:
+  repository_dispatch:
+    types: [ integration-pointer]
+jobs:
+  point:
+    runs-on: ubuntu-latest
+    steps:
+      - run: |
+          echo "Integration tests pointer updated: $(date +'%Y-%m-%d')"
--- a/.github/workflows/update_dependencies_and_client.yml
+++ b/.github/workflows/update_dependencies_and_client.yml
@@ -115,3 +115,18 @@ jobs:
           flankScripts shell firebase checkForSdkUpdate \
           --github-token=${{ secrets.GITHUB_TOKEN }} \
           --zenhub-token=${{ secrets.ZENHUB_API_KEY }}
+
+  trigger-pointer:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: tibdex/github-app-token@v1
+        id: generate-token
+        with:
+          app_id: ${{ secrets.FLANK_RELEASE_APP_ID }}
+          private_key: ${{ secrets.FLANK_RELEASE_PRIVATE_KEY }}
+
+      - name: Dispatch repository event
+        uses: peter-evans/repository-dispatch@v1
+        with:
+          token: ${{ steps.generate-token.outputs.token }}
+          event-type: dependencies-pointer
--- a/.github/workflows/update_dependencies_pointer.yml
+++ b/.github/workflows/update_dependencies_pointer.yml
@@ -1,0 +1,13 @@
+# Workflow is unable to retrieve data about previous workflows (the same file).
+# This workflow is a way to preserve last run date for update_dependencies_and_client.yml workflow
+# which is available from within that workflow
+name: Dependencies Update Pointer
+on:
+  repository_dispatch:
+    types: [ dependencies-pointer ]
+jobs:
+  point:
+    runs-on: ubuntu-latest
+    steps:
+      - run: |
+          echo "Dependencies pointer updated: $(date +'%Y-%m-%d')"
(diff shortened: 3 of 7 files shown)


pawelpasterz:
recheck

piotradamczyk5 (approved):

.github/workflows/integration_tests_pointer.yml
```diff
@@ -0,0 +1,13 @@
...
+name: Integration Tests Pointer
+on:
+  repository_dispatch:
+    types: [ integration-pointer]
+jobs:
+  point:
+    runs-on: ubuntu-latest
+    steps:
+      - run: |
+          echo "Integration tests pointer updated: $(date +'%Y-%m-%d')"
```
Sloox, line 13: how is it exactly preserving the date via an echo?

.github/workflows/update_dependencies_and_client.yml
```diff
@@ -115,3 +115,18 @@ jobs:
...
           --zenhub-token=${{ secrets.ZENHUB_API_KEY }}
+
+  trigger-pointer:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: tibdex/github-app-token@v1
+        id: generate-token
+        with:
+          app_id: ${{ secrets.FLANK_RELEASE_APP_ID }}
+          private_key: ${{ secrets.FLANK_RELEASE_PRIVATE_KEY }}
```
Sloox, line 126: I truly get nervous about sharing private keys with opensource repositories.
Are we sure its trusted.

.github/workflows/update_dependencies_pointer.yml
```diff
@@ -0,0 +1,13 @@
...
+name: Dependencies Update Pointer
+on:
+  repository_dispatch:
+    types: [ dependencies-pointer ]
+jobs:
+  point:
+    runs-on: ubuntu-latest
+    steps:
+      - run: |
+          echo "Dependencies pointer updated: $(date +'%Y-%m-%d')"
```
Sloox, line 13: Really need an explanation here 🤔

flank-scripts/build.gradle.kts
```diff
@@ -28,7 +28,7 @@ shadowJar.apply {
     }
 }
 // <breaking change>.<feature added>.<fix/minor change>
-version = "1.3.2"
+version = "1.3.3"
```
Sloox, line 31: hmmm check if this is not the same on develop

Sloox (changes requested):

.github/workflows/integration_tests_pointer.yml
pawelpasterz: Hm... I thought the description is just enough. Maybe I should add additional info...
Long story short, let's say we have workflow described in `a.yml`. During `a` run we want to fetch some data about previous `a` runs (like a time when it was previously finished). We use https://docs.github.com/en/free-pro-team@latest/rest/reference/actions#list-workflow-runs. Due to limitations, we get an empty array as a response.
But we can use the same endpoint to get info about another workflow (so inside `a` workflow we can get info about `b`). Those pointers are "just to be triggered" meaning they do nothing like preserves (by being triggered) timestamp, date etc.
For dependency update workflow looks like:
1. start dependency update workflow
2. proceed with all the jobs/steps but in parallel trigger (with repository dispatch) event "pointer" workflow
3. next time dependency update will be launched it will look for the latest date of "pointer" workflow (because it can not fetch data about itself)

> how is it exactly preserving the date via an echo?

Echo is just for debugging purposes. It could be nothing there actually

.github/workflows/update_dependencies_and_client.yml
pawelpasterz: The alternative is to create PAT and add it to secret. We use this logic in a couple of places inside the project. Actually, @piotradamczyk5 has configured this app and may has more detailed info

.github/workflows/update_dependencies_pointer.yml
pawelpasterz: https://github.com/Flank/flank/pull/1468#discussion_r553992274

flank-scripts/build.gradle.kts
pawelpasterz: Yup, it wasn't but the master is changing all the time, will update it

pawelpasterz: Hm...master is still having 1.3.2 (for now at least)

jan-gogo (commented):
Our CI is going to be a complicated system. I am not talking only about the impact of this PR. So you know what I mean. Will be helpful to have a documentation page for each workflow, with the described purpose and results.

piotradamczyk5:
> Our CI is going to be a complicated system. I am not talking only about the impact of this PR. So you know what I mean. Will be helpful to have a documentation page for each workflow, with the described purpose and results.

good point maybe let's create ticket for it

.github/workflows/update_dependencies_and_client.yml
piotradamczyk5: This is the most popular workflow for this purpose and this is secure.
Github apps are better to manage than PAT, so that's the reason why we use them

Sloox (approved):

PR #1470, opened 2021-01-07
fix: Resolve Fails on Windows

### LINKED ISSUE #1434: Resolve Artifacts Fails on Windows
**Describe the bug**
resolveArtifacts for windows fails due to the failure of privilege's that are required to create a symbolic link.
Files.kt
```
 Exception in thread "main" java.nio.file.FileSystemException:  A required privilege is not held by the client.
        at sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:86)
        at sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:97)
        at sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:102)
        at sun.nio.fs.WindowsFileSystemProvider.createSymbolicLink(WindowsFileSystemProvider.java:585)
        at java.nio.file.Files.createSymbolicLink(Files.java:1043)
        at flank.common.FilesKt.createSymbolicLink(Files.kt:17)
```

**To Reproduce**
It was tested on two different windows 10 machines. Windows 10 Home(2004) & Pro(1903).
Both fail. 
Reproduce by running it on Windows 10 native. Not a virtual machine.

Tested on commit - a7380ff0f3f2a71251593932daefdbbd53291917

Activating developer mode on Windows 10 Pro does not alleviate the problem.
We may need to elevate the privilege's of the java process or investigate a way around this issue.
Ideas.
1. Do an actual File copy for windows instead of a symbolic link
2. Checkout the conditions of this flag `SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE` and which OpenJDK supports this
3. Elevate the user process so it can complete successfully for windows. (Maybe execute a separate bat file or some other way.)

Sloox:
Fixes #1434 

Simple solution. Increases space used but works without process elevation.
works on Pro + Home.

### DIFF
--- a/.github/workflows/full_suite_integration_tests.yml
+++ b/.github/workflows/full_suite_integration_tests.yml
@@ -161,7 +161,7 @@ jobs:
 
   trigger-pointer:
     runs-on: ubuntu-latest
-    if: github.ref == 'refs/heads/master'
+    if: github.event_name != 'issue_comment'
     steps:
       - uses: tibdex/github-app-token@v1
         id: generate-token
--- a/.github/workflows/full_suite_integration_tests.yml
+++ b/.github/workflows/full_suite_integration_tests.yml
@@ -161,6 +161,7 @@ jobs:
 
   trigger-pointer:
     runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/master'
     steps:
       - uses: tibdex/github-app-token@v1
         id: generate-token
--- a/.github/workflows/full_suite_integration_tests.yml
+++ b/.github/workflows/full_suite_integration_tests.yml
@@ -161,7 +161,8 @@ jobs:
 
   trigger-pointer:
     runs-on: ubuntu-latest
-    if: github.event_name != 'issue_comment'
+    needs: [ process-results ]
+    if: always() && github.event_name != 'issue_comment'
     steps:
       - uses: tibdex/github-app-token@v1
         id: generate-token
--- a/common/src/main/kotlin/flank/common/Files.kt
+++ b/common/src/main/kotlin/flank/common/Files.kt
@@ -3,12 +3,12 @@ package flank.common
 import com.github.kittinunf.fuel.Fuel
 import com.github.kittinunf.fuel.core.awaitUnit
 import com.github.kittinunf.fuel.httpDownload
+import com.sun.org.apache.xpath.internal.operations.Bool
 import java.io.File
 import java.io.IOException
 import java.nio.file.Files
 import java.nio.file.Path
 import java.nio.file.Paths
-
 
 val userHome: String by lazy {
     if (isWindows) System.getenv("HOMEPATH") else System.getProperty("user.home")
@@ -19,7 +19,7 @@ fun String.deleteFile() = Paths.get(this).delete()
 fun linkFiles(
     link: String,
     target: String
-) = if (isWindows) copyDirectory(link, target) else createSymbolicLink(link, target)
+) = if (isWindows) copyDirectory(target, link) else createSymbolicLink(link, target)
 
 fun copyDirectory(sourceDirectoryLocation: String, destinationDirectoryLocation: String) {
     Files.walk(Paths.get(sourceDirectoryLocation))
@@ -86,6 +86,8 @@ fun String.fileCopyTo(directory: String, overwrite: Boolean = true): Boolean {
         true
     } else false
 }
+
+fun String.fileExists(): Boolean = Paths.get(this).exists()
 
 private fun Path.exists(): Boolean = Files.exists(this)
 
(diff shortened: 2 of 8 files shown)


Sloox:

Sloox:
recheck

Sloox:
@flank-it

common/src/main/kotlin/flank/common/Files.kt
```diff
@@ -14,6 +15,24 @@ val userHome: String by lazy {
...
+                Files.copy(source, destination)
+            } catch (e: IOException) {
+                logLn(e.localizedMessage)
+            }
+        }
+}
+
 fun createSymbolicLink(
     link: String,
     target: String
```
piotradamczyk5, line 38: Should not we apply similar logic for this method?

Sloox: We can but then its no longer creating a SymbolicLink and the name should change.

jan-gogo: Maybe this function signature should have a `typealias` so we could switch reference to function for specific linking strategy depending on OS type.

common/src/main/kotlin/flank/common/Files.kt
```diff
@@ -14,6 +15,24 @@ val userHome: String by lazy {
...
+        .forEach { source: Path ->
+            val destination =
+                Paths.get(destinationDirectoryLocation, source.toString().substring(sourceDirectoryLocation.length))
+            try {
+                Files.copy(source, destination)
+            } catch (e: IOException) {
+                logLn(e.localizedMessage)
+            }
+        }
+}
```
piotradamczyk5, line 34: Could we add some os based tests to it to just make sure that it creates symbolic link?

Sloox: @piotradamczyk5 there is already test for symbolic linking done previously.

Sloox: its also fixed for windows
i will add tests for new functions added

Sloox: @piotradamczyk5 added

common/src/main/kotlin/flank/common/Files.kt
Sloox: As mentioned we are copying for windows and symlinking for everyone else. I dont think there should be a typealias as its fairly self explanatory.

piotradamczyk5 (approved):

adamfilipow92:
I try to test your solution but when I try to build Flank I get
```
Execution failed for task ':flank-scripts:checkIfVersionUpdated'.
> Flank scripts version is not updated, but files changed.
  Please update version according to schema: <breaking change>.<feature added>.<fix/minor change>

```

pawelpasterz:
> I try to test your solution but when I try to build Flank I get
> 
> ```
> Execution failed for task ':flank-scripts:checkIfVersionUpdated'.
> > Flank scripts version is not updated, but files changed.
>   Please update version according to schema: <breaking change>.<feature added>.<fix/minor change>
> ```

Hey @adamfilipow92 I've implemented some changes, could you verify again?

adamfilipow92:
> > I try to test your solution but when I try to build Flank I get
> > ```
> > Execution failed for task ':flank-scripts:checkIfVersionUpdated'.
> > > Flank scripts version is not updated, but files changed.
> >   Please update version according to schema: <breaking change>.<feature added>.<fix/minor change>
> > ```
> 
> Hey @adamfilipow92 I've implemented some changes, could you verify again?

Sure!

adamfilipow92:
I think it's not related to this pr but two tests fail
```should create new epic with sub issues``` && ```should update existing epic``` from SDKUpdateTest, outputs looks the same.  Maybe we could skip these tests on windows and fix it in another ticket.

**Comment by pawelpasterz** [comment]
@adamfilipow92 thanks!
Fixed

**Comment by pawelpasterz** [comment]
@flank-it

**Comment by pawelpasterz** [comment]
@flank-it

File: .github/workflows/full_suite_integration_tests.yml [hunk 3]
```diff
@@ -161,6 +161,7 @@ jobs:
 
   trigger-pointer:
     runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/master'
```
piotradamczyk5, line 164: why this kind of check?

piotradamczyk5 (changes requested):

.github/workflows/full_suite_integration_tests.yml
pawelpasterz: Noticed during my work on this PR. Pointers should be only triggered by the master branch. We don't want comment trigger to change date.
If you want I can makanother PR for this change

piotradamczyk5:
```suggestion
    if: ${{ github.event_name != 'issue_comment'}}
```

pawelpasterz: Hm...same results, but sure, I can change it that way

piotradamczyk5: same, but we do not depend on the branch name.

piotradamczyk5 (approved):

adamfilipow92 (approved):
- [X] Works on windows
- [X] It passes on windows

PR #1471, opened 2021-01-07
feat: Migrate from bugsnag to sentry

### LINKED ISSUE #1318: Evaluate moving to Sentry from BugSnag

As a Flank developer, I want to understand crashes in Flank so I can fix them.

**Describe the solution**

Evaluate [Sentry](http://sentry.io/) as a BugSnag alternative.

adamfilipow92:
Fixes #1318 

1. Flank should display session id below ```revision```
2. Should save session id to ```session_id.txt``` in matrix path
3. Should upload ```session_id.txt``` when ```disable-results-upload``` not set
4. Unit & Integration tests should pass

## Checklist

- [X] Save and upload session id
- [X] Unit tested


buildSrc/src/main/kotlin/Versions.kt
```diff
@@ -2,6 +2,9 @@ object Versions {
     // https://github.com/bugsnag/bugsnag-java/releases
     const val BUGSNAG = "3.6.2"
```
axelzuziak-gogo, line 3: Can we remove it?

adamfilipow92:

adamfilipow92:
recheck

test_runner/src/main/kotlin/ftl/util/CreashReporter.kt
```diff
@@ -0,0 +1,60 @@
...
+internal fun initCrashReporter(
+    useMock: Boolean,
+    rootPath: String = System.getProperty("user.home")
+) = when {
+    useMock -> null
+    analyticsFileExistAndIsDisabled(rootPath) -> null
+    else -> initializeCrashReportWrapper()
+}
+
+private fun analyticsFileExistAndIsDisabled(rootPath: String) =
```
pawelpasterz, line 34: Hm...this name is not fully self-explanatory (for me). The logic here is to check if a user had disabled google analytics, flank should respect those settings.
How about `isGoogleAnalyticsDisabled`? Maybe not exactly the same but something explains why we check that file

test_runner/src/main/kotlin/ftl/util/CreashReporter.kt
```diff
@@ -0,0 +1,60 @@
+package ftl.util
```
pawelpasterz, line 1: Small typo in the file name `CreashReporter` -> `CrashReporter`

test_runner/src/main/kotlin/ftl/gc/GcStorage.kt
```diff
@@ -101,6 +103,14 @@ object GcStorage {
         )
     }
 
+    fun IArgs.uploadSessionId() = takeUnless { disableResultsUpload }?.let {
+        upload(
+            file = Path.of(localResultDir, SESSION_ID_FILE).toString(),
```
pawelpasterz, line 108: We use `Paths` in other places in the project. Your implementation is not wrong! I am just wondering if it is possible to use `Paths` here as well

test_runner/src/main/kotlin/ftl/util/CreashReporter.kt
adamfilipow92: agree

test_runner/src/main/kotlin/ftl/gc/GcStorage.kt
adamfilipow92: Changed! Thanks! 👍

pawelpasterz (approved):

piotradamczyk5 (approved):

PR #1534, opened 2021-01-25
ci: Added Windows workflow

### LINKED ISSUE #1528: Re-introduce Windows as a build platform

As a developer I want to ensure Windows is supported for flank. That includes tests.

**Describe the solution**
Lets include a Windows instance that can run a flank test suite (Unit tests and gradle clean build) to ensure we test it correctly

piotradamczyk5:
Fixes #1528

Windows workflow runs and passes

## Checklist

- [x] Added Windows workflow

### DIFF
--- a/.github/workflows/windows_workflow.yml
+++ b/.github/workflows/windows_workflow.yml
@@ -1,0 +1,32 @@
+name: windows-workflow
+
+on:
+  pull_request:
+    branches:
+      - '*'
+
+jobs:
+  build:
+    runs-on: windows-latest
+
+    steps:
+    - name: Cancel Previous Runs
+      uses: styfle/cancel-workflow-action@0.5.0
+      with:
+        access_token: ${{ secrets.GITHUB_TOKEN }}
+    - uses: actions/checkout@v2
+    - uses: gradle/wrapper-validation-action@v1
+    - uses: actions/cache@v2
+      with:
+        path: ~/.gradle/caches
+        key: ${{ runner.os }}-2-gradle-${{ hashFiles('**/*.gradle*') }}
+        restore-keys: |
+          ${{ runner.os }}-2-gradle-
+
+    - name: Gradle clean build
+      uses: eskatos/gradle-command-action@v1
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        HEAD_REF: ${{ github.head_ref }}
+      with:
+        arguments: "clean build"
--- a/common/src/main/kotlin/flank/common/Files.kt
+++ b/common/src/main/kotlin/flank/common/Files.kt
@@ -13,6 +13,10 @@ val userHome: String by lazy {
     if (isWindows) System.getenv("HOMEPATH") else System.getProperty("user.home")
 }
 
+val appDataDirectory: String by lazy {
+    if (isWindows) System.getenv("APPDATA") else System.getProperty("user.home")
+}
+
 fun linkFiles(
     link: String,
     target: String
--- a/test_runner/src/main/kotlin/ftl/args/CreateCommonArgs.kt
+++ b/test_runner/src/main/kotlin/ftl/args/CreateCommonArgs.kt
@@ -15,7 +15,7 @@ fun CommonConfig.createCommonArgs(
     // gcloud
     devices = gcloud::devices.require(),
     resultsBucket = ArgsHelper.createGcsBucket(
-        projectId = flank.project.orEmpty(),
+        projectId = flank::project.require(),
         bucket = gcloud::resultsBucket.require()
     ),
     resultsDir = gcloud.resultsDir ?: uniqueObjectName(),
@@ -41,7 +41,7 @@ fun CommonConfig.createCommonArgs(
     testTargetsAlwaysRun = flank::testTargetsAlwaysRun.require(),
     runTimeout = flank::runTimeout.require(),
     fullJUnitResult = flank::fullJUnitResult.require(),
-    project = flank.project.orEmpty(),
+    project = flank::project.require(),
     outputStyle = outputStyle,
     keepFilePath = flank::keepFilePath.require(),
     ignoreFailedTests = flank::ignoreFailedTests.require(),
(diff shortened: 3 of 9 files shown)


.github/workflows/windows_workflow.yml
```diff
@@ -0,0 +1,45 @@
+name: macOS-workflow
```
axelzuziak-gogo, line 1: Am I missing something here?

piotradamczyk5: I commited bad file, fixed. Thanks

axelzuziak-gogo (approved):

.github/workflows/windows_workflow.yml
```diff
@@ -0,0 +1,35 @@
+name: windows-workflow
+
+on:
+  pull_request:
+    branches:
+      - '*'
+
+jobs:
+  build:
+    runs-on: windows-latest
```
Sloox, line 10: Is there a set windows version? Do we know it? 
Do they update it often?
Windows is notoriously unstable across versions so I want to be sure we are constant.

Sloox (changes requested):

.github/workflows/windows_workflow.yml
```diff
@@ -0,0 +1,35 @@
...
+        path: ~/.gradle/caches
+        key: ${{ runner.os }}-2-gradle-${{ hashFiles('**/*.gradle*') }}
+        restore-keys: |
+          ${{ runner.os }}-2-gradle-
+
+    - name: Gradle clean build
+      uses: eskatos/gradle-command-action@v1
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        HEAD_REF: ${{ github.head_ref }}
```
Sloox, line 33: I think you need to add this to get it to be succesful:
```suggestion
        HEAD_REF: ${{ github.head_ref }}
        FLANK_PROJECT_ID: flank-open-source
```
or possibily 
```suggestion
        HEAD_REF: ${{ github.head_ref }}
        GOOGLE_CLOUD_PROJECT: flank-open-source
```

piotradamczyk5: this task run only local unit tests, so I do not think we need to specify project id

.github/workflows/windows_workflow.yml
piotradamczyk5: Version is specified at https://github.com/actions/virtual-environments

GitHub team make a version as `-latest` only when it is stable (ex. ubuntu still use 18.04 as latest)

.github/workflows/windows_workflow.yml
Sloox: @piotradamczyk5 the run is failing as the project field is null. I assume the project field requires this correct?
https://github.com/Flank/flank/pull/1534/checks?check_run_id=1767430233

Sloox: @piotradamczyk5 I see a lot of errors actually nevermind.

piotradamczyk5: I will check them, they could be related to other things

.github/workflows/windows_workflow.yml
Sloox: 👍

test_runner/src/main/kotlin/ftl/args/CreateCommonArgs.kt
```diff
@@ -15,7 +15,7 @@ fun CommonConfig.createCommonArgs(
     // gcloud
     devices = gcloud::devices.require(),
     resultsBucket = ArgsHelper.createGcsBucket(
-        projectId = flank::project.require(),
```
pawelpasterz, line 18: Why this change? Project id must never be empty

pawelpasterz (changes requested):

test_runner/src/main/kotlin/ftl/args/CreateCommonArgs.kt
```diff
@@ -41,7 +41,7 @@ fun CommonConfig.createCommonArgs(
     testTargetsAlwaysRun = flank::testTargetsAlwaysRun.require(),
     runTimeout = flank::runTimeout.require(),
     fullJUnitResult = flank::fullJUnitResult.require(),
-    project = flank::project.require(),
+    project = flank.project.orEmpty(),
```
piotradamczyk5, line 44:
```suggestion
    project = flank::project.require(),
```

test_runner/src/main/kotlin/ftl/args/CreateCommonArgs.kt
```diff
@@ -15,7 +15,7 @@ fun CommonConfig.createCommonArgs(
     // gcloud
     devices = gcloud::devices.require(),
     resultsBucket = ArgsHelper.createGcsBucket(
-        projectId = flank::project.require(),
+        projectId = flank.project.orEmpty(),
```
piotradamczyk5, line 18:
```suggestion
        projectId = flank::project.require(),
```

test_runner/src/main/kotlin/ftl/args/CreateCommonArgs.kt
piotradamczyk5: reverted

common/src/main/kotlin/flank/common/Files.kt
```diff
@@ -13,6 +13,10 @@ val userHome: String by lazy {
     if (isWindows) System.getenv("HOMEPATH") else System.getProperty("user.home")
 }
 
+val appDataDirectory: String by lazy {
+    if (isWindows) System.getenv("APPDATA") else System.getProperty("user.home")
```
Sloox, line 17: I had a feeling something related to HOMEPATH was going to be the issue. I assume that HOMEPATH you dont have write permissions on the VM'S right?

Sloox (approved):

common/src/main/kotlin/flank/common/Files.kt
piotradamczyk5: exactly. User directory is for users 🎉  Just one working day to figure this out

piotradamczyk5: and basically this is good practice on Windows to keep app data into app data folder instead of user one

PR #1537, opened 2021-01-27
ci: Integration Test for all OS's

### LINKED ISSUE #1527: Re-introduce integration tests for windows

I would like to ensure Windows as a platform is supported.
Integration tests need to run successfully on a windows platform

**Describe the solution**
@flank-it should run a successful windows instance.

Sloox:
Fixes #1527 

Need it to be merged before its actual testing

## Checklist

### DIFF
--- a/.github/workflows/full_suite_integration_tests.yml
+++ b/.github/workflows/full_suite_integration_tests.yml
@@ -72,7 +72,10 @@ jobs:
   run-it-full-suite:
     needs: [ should_run_it ]
     if: needs.should_run_it.outputs.run_integration_tests == 'true'
-    runs-on: windows-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ macos-latest, windows-latest, ubuntu-latest ]
     outputs:
       job_status: ${{ job.status }}
       build-scan-url: ${{ steps.run-it.outputs.build-scan-url }}
--- a/.github/workflows/full_suite_integration_tests.yml
+++ b/.github/workflows/full_suite_integration_tests.yml
@@ -72,10 +72,7 @@ jobs:
   run-it-full-suite:
     needs: [ should_run_it ]
     if: needs.should_run_it.outputs.run_integration_tests == 'true'
-    runs-on: ${{ matrix.os }}
-      strategy:
-        matrix:
-          os: [ ubuntu-latest, windows-latest, macos-latest ]
+    runs-on: windows-latest
     outputs:
       job_status: ${{ job.status }}
       build-scan-url: ${{ steps.run-it.outputs.build-scan-url }}
--- a/.github/workflows/full_suite_integration_tests.yml
+++ b/.github/workflows/full_suite_integration_tests.yml
@@ -72,7 +72,10 @@ jobs:
   run-it-full-suite:
     needs: [ should_run_it ]
     if: needs.should_run_it.outputs.run_integration_tests == 'true'
-    runs-on: [ubuntu-latest, macos-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+      strategy:
+        matrix:
+          os: [ ubuntu-latest, windows-latest, macos-latest ]
     outputs:
       job_status: ${{ job.status }}
       build-scan-url: ${{ steps.run-it.outputs.build-scan-url }}


Sloox:
@flank-it

axelzuziak-gogo (approved):

piotradamczyk5 (changes requested):
it does not work good. IT failed, but no message was posted

Sloox:
@piotradamczyk5 if i am correct this needs to be merged into master to actually be executable for the IT tests ?

.github/workflows/full_suite_integration_tests.yml
```diff
@@ -72,7 +72,10 @@ jobs:
...
     needs: [ should_run_it ]
     if: needs.should_run_it.outputs.run_integration_tests == 'true'
-    runs-on: [ubuntu-latest, macos-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+      strategy:
+        matrix:
+          os: [ ubuntu-latest, windows-latest, macos-latest ]
     outputs:
       job_status: ${{ job.status }}
       build-scan-url: ${{ steps.run-it.outputs.build-scan-url }}
```
piotradamczyk5, line 81: I am not sure how to handle this `build-scan-url`, because previously we expect 1 link, but now we will have 3 of them

Sloox: @piotradamczyk5 hmmm, maybe we should chat about it and attempt another solution. I personally feel that the IT should run on all platforms for stability. A requirement of flank is to provide support for all 3 platforms so it would be great. 
We can however approach it differently.

piotradamczyk5: sure we could figure out the best solution, we should involve @pawelpasterz because he is a father of integration tests

pawelpasterz:
> @piotradamczyk5 if i am correct this needs to be merged into master to actually be executable for the IT tests ?

AFAIK it is true for newly created workflows. If you just modify an existing one you should be able to launch 'new' version from the branch

.github/workflows/full_suite_integration_tests.yml
pawelpasterz: > @piotradamczyk5 hmmm, maybe we should chat about it and attempt another solution. I personally feel that the IT should run on all platforms for stability. A requirement of flank is to provide support for all 3 platforms so it would be great.
We can however approach it differently.

IMO that would be great! 
Modifying `integration processResult` script won't be difficult. Question is how to pass 3 build scan URLs?
As I see it there are a couple of action items here:
1. add more OS ✅ 
2. figure out how to pass 3 URLs to the another jobs
3. change `integration processResult` to handle 3 URLs (and 3 results actually)
4. change `post-comment_with_results` to handle 3 urls

pawelpasterz: I am wondering how job is behaving with os matrix? Is it run sequentially/in parallel with different os each? How about `outputs` - are they somehow aggregated? Maybe 2) from my previous comment won't be so difficult. We could set URL, for each os, with suffix, like: 
```
echo "::set-output name=scan-url-${{ matrix.os }}::${{ steps.run-it.outputs.build-scan-url }}"
```
and then set explicit job output (example):
```
(...)

outputs:
  build-scan-url-macos: ${{ steps.run-it.outputs.build-scan-url-macos }}
  build-scan-url-ubuntu: ${{ steps.run-it.outputs.build-scan-url-ubuntu }}

(etc...)
```
Then fetching those data will be easy.
Of course, it's my guess and it should be verified

pawelpasterz: Anyway, we should decide if we want to do it within this PR scope? IMO enabling new os brings value itself so we can handle additional logic in separated issue/pr.
I think we could create a separate job, running for ubuntu and windows. It will spin in parallel to the existing macOS job and will not interfere with the existing 'processing result logic`. Simultaneously let's create a new issue for updating logic and handle multiple build scan URLs. I would be happy to work on it

Sloox:
@flank-it

Sloox:
@flank-it

Sloox:
@pawelpasterz @piotradamczyk5 as far as i can see it seems like it does need to be merged into master for changes to apply.
The most recent change only runs on windows yet it still has the same OS error as before, "Cannot find windows, linux or mac"
https://github.com/Flank/flank/actions/runs/517401129

piotradamczyk5:
> > @piotradamczyk5 if i am correct this needs to be merged into master to actually be executable for the IT tests ?
> 
> AFAIK it is true for newly created workflows. If you just modify an existing one you should be able to launch 'new' version from the branch

if you listen for comment trigger, you must merge it first

pawelpasterz:
> > > @piotradamczyk5 if i am correct this needs to be merged into master to actually be executable for the IT tests ?
> > 
> > 
> > AFAIK it is true for newly created workflows. If you just modify an existing one you should be able to launch 'new' version from the branch
> 
> if you listen for comment trigger, you must merge it first

Didn't know, thanks!

Sloox:
@piotradamczyk5 @pawelpasterz Shall i try fix the ubuntu workflow apart of this branch before we merge. it appears to be broken again :/

.github/workflows/full_suite_integration_tests.yml
Sloox: captured here: https://github.com/Flank/flank/issues/1539

PR #1557, opened 2021-02-02
fix: Move shards json to result directory

### LINKED ISSUE #1475: Move shards.json to results directory
**Describe the bug**

Currently, flank always saves shards file in the current directory. This behavior can cause overriding this file when user run flank multiple times at the same time.

**Expected behavior**

When we run tests we should save shards file to result directory. This behavior can cause overriding this file when user run flank multiple times at the same time.

adamfilipow92:
Fixes #1475 

1. Run flank with android or ios test
1. Wait until ```results-directory``` is created
1. Open the results directory and check ```*_shards.json``` exists
1. If   ```disable-results-upload``` not set, check shards file exists in gcloud bucket 
1. Flank shouldn't upload shards file to gcs if ```disable-results-upload``` set

## Checklist

- [X] Documented [(SDD in ticket comment)](https://github.com/Flank/flank/issues/1475#issuecomment-769227112)
- [x] Unit tests updated


adamfilipow92:
@flank-it

Sloox (approved):

test_runner/src/main/kotlin/ftl/args/AndroidArgs.kt
```diff
@@ -151,5 +153,8 @@ val AndroidArgs.isSanityRobo
 val AndroidArgs.isGameLoop
     get() = if (type == null) false else (type == Type.GAMELOOP)
 
+val AndroidArgs.shardsFilePath
+    get() = Paths.get(localResultDir, resultsDir, ANDROID_SHARD_FILE).toAbsolutePath().toString()
```
piotradamczyk5, line 157: if `localResultsDir` is different than default, we should not add file to `resultsDir`. We have similar logic in the whole app

piotradamczyk5 (approved):

PR #1560, opened 2021-02-03
fix: Fix IT workflow for windows

Sloox:
Fixes #

Can only be tested once its merged due to IT structure.


jan-gogo (approved):

.github/workflows/full_suite_integration_tests.yml
```diff
@@ -99,6 +99,8 @@ jobs:
 
       - name: Prepare Google Service Account
         env:
+          FLANK_PROJECT_ID: flank-open-source
+          GOOGLE_CLOUD_PROJECT: flank-open-source
```
piotradamczyk5, line 103: Hey @Sloox I think that is a bad workaround for the real problem, project id should be mocked for testing

Sloox: @piotradamczyk5 do we have an example for this? I have looked and my windows machines looks for it this way.

Sloox: its also not needed on ubuntu and mac so its confused me a bit.

piotradamczyk5: If we need to have it set, then make it happen in integration tests Gradle, but I am pretty sure that it could be taken from the service account file. It works that way previously

adamfilipow92: @Sloox 
Check this doc please: https://flank.github.io/flank/#authenticate-with-a-service-account

.github/workflows/full_suite_integration_tests.yml
```diff
@@ -99,12 +99,14 @@ jobs:
...
           GCLOUD_KEY: ${{ secrets.GCLOUD_KEY }}
         run: |
             if [ "$RUNNER_OS" == "Windows" ]; then
                 set GCLOUD_DIR="%HOMEPATH%/.config/gcloud/"
                 mkdir %GCLOUD_DIR%
-                echo certutil -decode gcloud_key.txt %GCLOUD_DIR%application_default_credentials.json
+                echo certutil -decode %GCLOUD_KEY% %GCLOUD_DIR%application_default_credentials.json
```
piotradamczyk5, line 109: if this work, let's remove
```
FLANK_PROJECT_ID: flank-open-source
GOOGLE_CLOUD_PROJECT: flank-open-source
```

adamfilipow92: We don't need these additional variables because ```application_default_credentials.json``` contains this data

piotradamczyk5:
I guess that PR title should be also changed

.github/workflows/full_suite_integration_tests.yml
Sloox: @piotradamczyk5 removed. @adamfilipow92

.github/workflows/full_suite_integration_tests.yml
Sloox: removed.

.github/workflows/full_suite_integration_tests.yml
```diff
@@ -85,7 +85,7 @@ jobs:
 
       - name: Checkout Pull Request
         if: github.event_name == 'issue_comment'
-        env:
+        env:Invalid value for [
```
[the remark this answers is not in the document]
Sloox: removed.

piotradamczyk5 (approved):

adamfilipow92 (approved):

PR #1564, opened 2021-02-04
fix: windows issues

Sloox:
Fixes Nothing for now due to more changes to come

Need to test if these changes fix IT tests.

### DIFF
--- a/.github/workflows/full_suite_integration_tests.yml
+++ b/.github/workflows/full_suite_integration_tests.yml
@@ -99,7 +99,9 @@ jobs:
 
       - name: Prepare Google Service Account
         env:
+          GOOGLE_CLOUD_PROJECT: flank-open-source
           GCLOUD_KEY: ${{ secrets.GCLOUD_KEY }}
+
         run: |
             if [ "$RUNNER_OS" == "Windows" ]; then
                 set GCLOUD_DIR="%HOMEPATH%/.config/gcloud/"
--- a/.github/workflows/full_suite_integration_tests.yml
+++ b/.github/workflows/full_suite_integration_tests.yml
@@ -99,7 +99,7 @@ jobs:
 
       - name: Prepare Google Service Account
         env:
-          GOOGLE_CLOUD_PROJECT: flank-open-source
+          GOOGLE_CLOUD_PROJECT: flank-open-source #needs to be removed.
           GCLOUD_KEY: ${{ secrets.GCLOUD_KEY }}
 
         run: |
--- a/test_runner/src/main/kotlin/ftl/ios/xctest/common/InstallParseBinaries.kt
+++ b/test_runner/src/main/kotlin/ftl/ios/xctest/common/InstallParseBinaries.kt
@@ -3,8 +3,7 @@ package ftl.ios.xctest.common
 import flank.common.OutputLogLevel
 import flank.common.appDataDirectory
 import flank.common.createDirectoryIfNotExist
-import flank.common.createFileCopy
-import flank.common.createSymbolicLinkToFile
+import flank.common.createLinkToFile
 import flank.common.downloadFile
 import flank.common.hasAllFiles
 import flank.common.isMacOS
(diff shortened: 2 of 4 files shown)


.github/workflows/full_suite_integration_tests.yml
```diff
@@ -99,7 +99,9 @@ jobs:
 
       - name: Prepare Google Service Account
         env:
+          GOOGLE_CLOUD_PROJECT: flank-open-source
```
Sloox, line 102: I know this is a hated solution but for github integration test I want to test if this actually makes a difference.

test_runner/src/main/kotlin/ftl/ios/xctest/common/InstallParseBinaries.kt
```diff
@@ -54,8 +55,16 @@ private fun downloadAndUnzip(osname: String) {
             it.setExecutable(true)
         }
     Files.delete(destinationFile)
-    createSymbolicLinkToFile(
-        link = Paths.get(flankBinariesDirectory.toString(), "libatomic.so.1"),
-        target = Paths.get(flankBinariesDirectory.toString(), "libatomic.so.1.2.0")
-    )
+
+    if (isWindows) {
```
Sloox, line 59: we cannot create a symobolic link on native windows without admin privalages.

test_runner/src/main/kotlin/ftl/ios/xctest/common/InstallParseBinaries.kt
```diff
@@ -54,8 +55,16 @@ private fun downloadAndUnzip(osname: String) {
...
+        createFileCopy(
+            "$flankBinariesDirectory\\libatomic.so.1.2.0",
+            "$flankBinariesDirectory\\libatomic.so.1"
+        )
+    } else {
+        createSymbolicLinkToFile(
+            link = Paths.get(flankBinariesDirectory.toString(), "libatomic.so.1"),
+            target = Paths.get(flankBinariesDirectory.toString(), "libatomic.so.1.2.0")
+        )
+    }
```
piotradamczyk5, line 69: I think that we should hide files copy inside `createSymbolicLinkToFile`

.github/workflows/full_suite_integration_tests.yml
piotradamczyk5: so please add it in `:integration_test build.gradle.kts` using gradle extension `environment(key, value)`

Sloox: @piotradamczyk5 Why do you want it there? This change is temporary and only meant to be for the github actions not for the project. 
I therefore dont see why it must be in gradle?
I honestly am not expecting it to work anyways so will remove it afterwards.

WDYT?

piotradamczyk5: Ok, maybe leave a comment to remove it afterwards

test_runner/src/main/kotlin/ftl/ios/xctest/common/InstallParseBinaries.kt
piotradamczyk5: yes, but we could make a copy  `if (windows)` so there will be just single method
`createSymbolicLink` which has inside copy implementation for windows and symbolic link for others

test_runner/src/main/kotlin/ftl/ios/xctest/common/InstallParseBinaries.kt
Sloox: we should then change createSymbolicLinkToFile to be complient as it will not createSymbolicLinkToFile for windows and therefore the name is not true to its meaning.

I am happy to rename it and do as mentioned but leaving it as createSymbolicLinkToFile is deceptive.

piotradamczyk5: yeah, we could think about other name, but using everwhere 
```
if(isWindows) copy else copySymbolicLink 
```

is not good choice IMO

test_runner/src/main/kotlin/ftl/ios/xctest/common/InstallParseBinaries.kt
Sloox: 👍

test_runner/src/main/kotlin/ftl/ios/xctest/common/InstallParseBinaries.kt
Sloox: I changed the name have a look :)

piotradamczyk5 (approved):

adamfilipow92 (approved):

PR #1572, opened 2021-02-08
refactor: Flank scripts features not depends on each other

### LINKED ISSUE #1543: Make flank-scripts features not depends on each other

As a contributor, I want to keep features not dependent on each other so I don't have to be afraid of the hidden complexity of the domain layer.

**Describe the solution**

1. Identify relations between different `flank-scripts` features.
1. Consider identified code as a standalone utility(ies).
1. Move common implementation to the separated package(s) outside `ops`.

adamfilipow92:
Fixes #1543 

The tests should pass. Scripts should work as before.

## Checklist

- [X] Unit tested

### DIFF
--- a/flank-scripts/build.gradle.kts
+++ b/flank-scripts/build.gradle.kts
@@ -28,7 +28,7 @@ shadowJar.apply {
     }
 }
 // <breaking change>.<feature added>.<fix/minor change>
-version = "1.5.1"
+version = "1.5.2"
 group = "com.github.flank"
 
 application {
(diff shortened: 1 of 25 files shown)


Sloox (changes requested):
Is this only a package name change?

adamfilipow92:
> Is this only a package name change?

I was moving some functions to ```common``` package like ```AppendReleaseNotes.kt``` because is used in two other packages.

piotradamczyk5 (approved):
It needs some adjustments after merge #1574 , but for now the main goal is achieved 👍

Sloox (approved):
Understand it now and approved: 🚀

PR #1585, opened 2021-02-10
fix: Print Flank version only on run command

### LINKED ISSUE #1561: Only print flank version on Run command
**Describe the bug**

Every command starts by printing the flank version. I think only the run command should start by printing the version.

The workflow is someone runs Flank and then wants to know what version of flank was used to execute the tests. I don't think it makes sense to print the flank version on `flank provided-software list` for example.

**To Reproduce**

1. Run `flank provided-software list`
2. See the version is printed

```
$ flank provided-software list
version: local_snapshot
revision: ae09391820ab7facaaa5b7679876e9ae2c0edaed
session id: 39f08992-c5c8-4f61-9fb2-e61aea7f7a97

┌──────────────────────┐
│ ORCHESTRATOR VERSION │
├──────────────────────┤
│ 1.0.2                │
└──────────────────────┘
```

**Expected behavior**

No version printed unless we are invoking a `flank run ...` command or the version command directly.

```
$ flank provided-software list

┌──────────────────────┐
│ ORCHESTRATOR VERSION │
├──────────────────────┤
│ 1.0.2                │
└──────────────────────┘
```

>

```
flank --version
version: local_snapshot
revision: ae09391820ab7facaaa5b7679876e9ae2c0edaed
session id: 35f6c8a1-2f32-44b7-963d-62b7ff1c6933
```

adamfilipow92:
Fixes #1561 

## I case

1. Run android or ios test
1. Flank should print version, revision, and sessionId 

## II case

1. Run Flank with ```--version```
1. Flank should print version, revision, and sessionId 

## III case

1. Run a command like ```flank provided-software list````
1. Version information should not be printed.

### DIFF
--- a/test_runner/src/main/kotlin/ftl/cli/firebase/test/android/AndroidRunCommand.kt
+++ b/test_runner/src/main/kotlin/ftl/cli/firebase/test/android/AndroidRunCommand.kt
@@ -60,7 +60,6 @@ class AndroidRunCommand : CommonRunCommand(), Runnable {
     }
 
     override fun run() {
-
         printVersionInfo()
 
         if (dryRun) {
--- a/test_runner/src/main/kotlin/ftl/cli/firebase/test/android/AndroidRunCommand.kt
+++ b/test_runner/src/main/kotlin/ftl/cli/firebase/test/android/AndroidRunCommand.kt
@@ -20,6 +20,7 @@ import ftl.run.dumpShards
 import ftl.run.newTestRun
 import ftl.util.DEVICE_SYSTEM
 import ftl.util.TEST_TYPE
+import ftl.util.printVersionInfo
 import ftl.util.setCrashReportTag
 import kotlinx.coroutines.runBlocking
 import picocli.CommandLine
@@ -59,6 +60,9 @@ class AndroidRunCommand : CommonRunCommand(), Runnable {
     }
 
     override fun run() {
+
+        printVersionInfo()
+
         if (dryRun) {
             MockServer.start()
         }
--- a/test_runner/src/main/kotlin/ftl/util/Utils.kt
+++ b/test_runner/src/main/kotlin/ftl/util/Utils.kt
@@ -3,6 +3,7 @@
 package ftl.util
 
 import com.fasterxml.jackson.annotation.JsonProperty
+import flank.common.logLn
 import ftl.run.exception.FlankGeneralError
 import java.io.InputStream
 import java.time.Instant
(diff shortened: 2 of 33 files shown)


test_runner/src/main/kotlin/ftl/util/Utils.kt
```diff
@@ -68,6 +69,13 @@ private fun getResource(name: String): InputStream {
         ?: throw FlankGeneralError("Unable to find resource: $name")
 }
 
+fun printVersionInfo() {
+    logLn("version: " + readVersion())
```
Sloox, line 73:
```suggestion
    logLn("version:${readVersion}")
```
?
The same for all 2 lines below?

test_runner/src/main/kotlin/ftl/cli/firebase/test/android/AndroidRunCommand.kt
```diff
@@ -59,6 +60,9 @@ class AndroidRunCommand : CommonRunCommand(), Runnable {
     }
 
     override fun run() {
+
```
Sloox, line 63: hmmm do you have git pre commit working? :)

Sloox (changes requested):

test_runner/src/main/kotlin/ftl/util/Utils.kt
adamfilipow92: Thanks! Changed!

test_runner/src/main/kotlin/ftl/cli/firebase/test/android/AndroidRunCommand.kt
adamfilipow92: Yes, but pre commit hook not catch that. I removed this line! Thanks!

pawelpasterz:
It would be great to have a couple of tests to verify it?

adamfilipow92:
> It would be great to have a couple of tests to verify it, do you think it requires lots of work to implement it?

Right! Tests added! 👍

piotradamczyk5 (approved):
Tested few commands (`flank ios versions list`, `flank android models list`, `flank network-profiles list`) and middle command (help of groups) - version is not printed
Tested (`flank -v `, `flank android run`, `flank ios run`) - Version is printed

Code looks good 👍

jan-gogo (changes requested):
I love that now we are absolutely sure that the version will not be printed in other commands than run, but IMO from a maintenance perspective, it's a bad idea to have a test for checking that something wasn't occurring. Do you know what I mean? There is an infinite number of things that cannot occur in flank runtime, we shouldn't focus on them.

jan-gogo (approved):

Sloox (approved):

PR #1657, opened 2021-03-03
fix: Remove physical device from integration tests

### LINKED ISSUE #1632: Remove physical devices (android) from IT configs
Physical devices burn flank's quota extremely fast. Let's get rid of them!

Currently, there is no support for virtual iPhone instances (on FTL) so iOS tests must remain as they are (at least for now).
What we can do is to remove `HUR` device from https://github.com/Flank/flank/blob/c8d0ff4300af5285b197afaaf29ed124d88fc8f2/integration_tests/src/test/resources/cases/flank_android_multiple_devices.yml#L10

What needs to be done?
* remove `HUR` device (it can't be replaced with another virtual device, our artifacts are compiled with SDK version 30, and currently only 2 virt dev support it, we already use both)
* integration test to be updated (assertions and compare file)

adamfilipow92:
Fixes #1632 

IT should pass.

## Checklist

### DIFF
--- a/integration_tests/src/test/kotlin/integration/MultipleDevicesIT.kt
+++ b/integration_tests/src/test/kotlin/integration/MultipleDevicesIT.kt
@@ -34,7 +34,6 @@ class MultipleDevicesIT {
             resOutput,
             "app-multiple-success-debug-androidTest.apk",
             "app-multiple-error-debug-androidTest.apk",
-            "performanceMetrics.json"
         )
         assertContainsOutcomeSummary(resOutput) {
             success = 6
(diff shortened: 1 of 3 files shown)


piotradamczyk5 (approved):

piotradamczyk5:
@flank-it

piotradamczyk5:
@adamfilipow92 please check IT

piotradamczyk5 (approved):

piotradamczyk5:
@flank-it

Sloox (approved):

PR #1661, opened 2021-03-05
test: Implement logic to clear LocalGcs after each test

### LINKED ISSUE #1578: LocalGcs should be cleaned after each test
This is a follow-up issue of #1577 

Here is the copy-pasted description:

_After some digging it turned out it's due to `GcStorage` client and `FlankTestRunner`. Here are some of my thoughts and conclusions:_
* _When FlankTestRun is initiated it uploads some artifacts needed during tests. GcStorage uses FakeStorageRpc (during test run) under the hood and one preserves all artifatcs/files/and other files as String - byte[] map. #1350 added two new artifacts that are uploaded and one of them is EarlGreyExample.zip which is quite heavy ~47MB._
* _There are a couple of tests which uploads those zip file but with different key (which expected). Yet, it leads to the point when we have few big byte[] in memory and referenced by GcStorage (so gc can not free the memory)_
* _I checked and the default xmx for gradle test task on GH action runner is 512 MB._
* _In general, it is not so small number but...if we run mockedIosTestRun gcsAndHistoryName and mockedIosTestRun local we already have ~141MB of memory occupied only by those zip files, one for the FlankTestRunner and each of tests 'uploads' zip on it's own. And we need to remember there are other lots of other tests that 'uploads' files..._
* _All above can cause OOM error being thrown (at least I think so)_

#### Story:
As a developer I want storage to be cleaned after each test so that I am sure there will be no memory leak during tests.

#### Solution:
There are two possible approaches:
1. clear entire `LocalGcs` after each test:
    1. check if it impacts overall test performance
    2. implement logic to clear `LocalGcs`
    3. find tests that require test artifacts to be uploaded and add logic to upload required files manually (currently they are uploaded during `FlankTestRunner` initialization 
        https://github.com/Flank/flank/blob/master/test_runner/src/test/kotlin/ftl/test/util/FlankTestRunner.kt#L13
        https://github.com/Flank/flank/blob/master/test_runner/src/test/kotlin/ftl/test/util/LocalGcs.kt#L32
    4. clear storage after each test
2. clear file uploaded during tests only:
    1. implement logic to clear storage only for specific files (all pre-uploaded files are held in `tmp_bucket_2`, `FakeStorageRpc` holds files(as byte[]) in hashmap with string key consisting of `[bucket_name]/[filename]`, we could use that info)
    2. clear storage after each test that requires it (maybe we do not clear after all of them)

#### Additional thoughts:
Maybe we can use `FlankTestRunner`(junit runner) to handle some work? (like automatic storage cleaning ) -- needs to be investigated.

Solutions are just proposals, we do not need to strictly follow them.

pawelpasterz:
Fixes #1578 

Change description:
* there are only 2 tests that require pre-uploaded files -- now they upload filed on their own (`ArgsHelperTest#assertGcsFileExists succeeds` & `TestRunnerTest#mockedAndroidTestRun gcsAndHistoryName`)
* Google API provides `FakeStorageRpc` which is created with `LocalStorageHelper.getOptions()`
* `FakeStorageRpc` can be cleaned/reseted only by calling `LocalStorageHelper.getOptions()` (both its hashmaps are cleared)
* Above is, unfortunately, the only way to clear `FakeStorageRpc` -- storage has `reset` method but it's not exposed
* The `TestStorageProvider` is introduced to wrap additional test logic
* Intention behind `TestStorageProvider` is `GcStorage` should not be aware of how it is used (tests or production )
* The `TestStorageProvider` has nullable `Storage` filed. It can be cleared (set null) with `#clearStorage`
* If storage is null `TestStorageProvider` will invoke `LocalStorageHelper.getOptions()` and assign reference to storage field.
* `FlankTestRunner` has a custom listener that will invoke `LocalGcs#clear()` after each test run

Unfortunately, there is no straightforward way to verify it since OOM error occurred only on GH Actions machines, randomly.

### DIFF
--- a/test_runner/src/main/kotlin/ftl/gc/GcStorage.kt
+++ b/test_runner/src/main/kotlin/ftl/gc/GcStorage.kt
@@ -224,7 +224,6 @@ object TestStorageProvider {
         }
 
     fun clearStorage() {
-        println("storage null")
         backingStorage = null
     }
 }
--- a/test_runner/src/main/kotlin/ftl/gc/GcStorage.kt
+++ b/test_runner/src/main/kotlin/ftl/gc/GcStorage.kt
@@ -51,7 +51,7 @@ object GcStorage {
     val storage: Storage by lazy {
         GoogleApiLogger.silenceComputeEngine()
         if (FtlConstants.useMock) {
-            LocalStorageHelper.getOptions().service
+            TestStorageProvider.storage
         } else {
             storageOptions.service
         }
@@ -209,4 +209,22 @@ object GcStorage {
     }
 
     private fun String.dropLeadingSlash() = drop(1)
+}
+
+object TestStorageProvider {
+    init {
+        require(FtlConstants.useMock) { "Storage provider can be used only during tests" }
+    }
+
+    private var backingStorage: Storage? = null
+
+    val storage: Storage
+        get() = backingStorage ?: LocalStorageHelper.getOptions().service.apply {
+            backingStorage = this
+        }
+
+    fun clearStorage() {
+        println("storage null")
+        backingStorage = null
+    }
 }
--- a/test_runner/src/test/kotlin/ftl/args/ArgsHelperTest.kt
+++ b/test_runner/src/test/kotlin/ftl/args/ArgsHelperTest.kt
@@ -16,6 +16,7 @@ import ftl.shard.TestMethod
 import ftl.shard.TestShard
 import ftl.shard.stringShards
 import ftl.test.util.FlankTestRunner
+import ftl.test.util.LocalGcs
 import ftl.test.util.TestHelper.absolutePath
 import ftl.test.util.assertThrowsWithMessage
 import io.mockk.every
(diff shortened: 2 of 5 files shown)


Sloox (approved):

pawelpasterz:
@flank-it

test_runner/src/main/kotlin/ftl/gc/GcStorage.kt
```diff
@@ -210,3 +210,20 @@ object GcStorage {
...
+    init {
+        require(FtlConstants.useMock) { "Storage provider can be used only during tests" }
+    }
+
+    private var backingStorage: Storage? = null
+
+    val storage: Storage
+        get() = backingStorage ?: LocalStorageHelper.getOptions().service.apply {
+            backingStorage = this
+        }
```
piotradamczyk5, line 224: looks like `backingStorage` is just lazy field

piotradamczyk5 (approved):

PR #1680, opened 2021-03-11
fix: Fix output report error with dump shards

pawelpasterz:
When a user tries to dump shards AND the output report is enabled an error is thrown. This is due to the fact flank is trying to create `outputReport.json` file inside result directory which does not exist for dump shard run, to reproduce run on the current master:
1. add `output-report: json` to the config file
2. start flank with `--dump-shards`
3. see `FileNotFound` error

1. add `output-report: json` to the config file
2. start flank with `--dump-shards`
3. no error is thrown and flank finishes run normally

## Checklist

- [x] Unit tested

### DIFF
--- a/common/src/main/kotlin/flank/common/Files.kt
+++ b/common/src/main/kotlin/flank/common/Files.kt
@@ -111,7 +111,7 @@ fun File.hasAllFiles(fileList: List<String>): Boolean {
 
 fun String.fileExists(): Boolean = Paths.get(this).exists()
 
-fun osPathSeperator() = (if (isWindows) "\\" else "/")
+fun osPathSeparator() = (if (isWindows) "\\" else "/")
 
 private fun Path.exists(): Boolean = Files.exists(this)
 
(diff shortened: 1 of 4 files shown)


pawelpasterz:
@flank-it

common/src/main/kotlin/flank/common/Files.kt
```diff
@@ -111,7 +111,7 @@ fun File.hasAllFiles(fileList: List<String>): Boolean {
 
 fun String.fileExists(): Boolean = Paths.get(this).exists()
 
-fun osPathSeperator() = (if (isWindows) "\\" else "/")
+fun osPathSeparator() = (if (isWindows) "\\" else "/")
```
Sloox, line 114: Am i blind what actually changed on this line?

Sloox (approved):

common/src/main/kotlin/flank/common/Files.kt
pawelpasterz: There was a typo `sepErator`. I've changed it to `sepArator`

piotradamczyk5: is it different from `File.pathSeparator` ?

adamfilipow92 (approved):
- [X] Code looks good
- [X] Tested and works
👍

Issue #1710, opened 2021-03-17
Allow device & shard count validation to be skipped


inktomi:
As a firebase test lab user, I may be offered access from time to time above the normal limits. As a flank & fladle user, I would like to be able to pass in customized config values to use my account to the fullest - but the configuration validation prevents it. 

I would like the ability to skip config validation by passing in a flag at runtime. This flag could even be part of the config, it doesn't need to be a command line option. In the event of having an invalid config but passing the skip validation flag, Flank should crash (unparsable config) or pass the values provided (eg, 1000 shards, or `FluffyBunny` device type, etc). 

As this is part of Flank that's (correctly!) built in, it's not currently possible to work around this when passing in otherwise invalid parameters.

jan-gogo:
**For estimation**
1. Add new option `skip-config-validation`
2. Update implementation for skipping o args validation if this option is used.

PR #1711, opened 2021-03-17
feat: Add support for emulator devices

pawelpasterz:
In the scope of #1710

PR adds support for `EMULATOR` devices

1. build flank locally with `./gradlew flankFullRun`
2. ensure you are logged to `flank-open-source` project
3. run `flank firebase test android models list`
4. you should see new devices of type `EMULATOR`
5. run a simple test with
    ```
      device:
         - model: NexusLowResEmulator
           version: 28
    ```
6. tests should finish normally

## Checklist

- [x] Unit tested


pawelpasterz:
@flank-it

inktomi:
How will this work with the existing checks for max shards? They check specifically physical and virtual devices - as best I know there are not limits for these new emulators - so we should ensure that they don't fall into a default bucket of some sort in the validation logic.

pawelpasterz:
> How will this work with the existing checks for max shards? They check specifically physical and virtual devices - as best I know there are not limits for these new emulators - so we should ensure that they don't fall into a default bucket of some sort in the validation logic.

We will introduce skipping validation option in a future PR. AFAIK emulators are considered as virtual devices so you can use up to 500 shards. Just wanted to keep PR scope related to one topic :)

Sloox (approved):
10k lines made me cry at first

piotradamczyk5 (approved):

PR #1715, opened 2021-03-21
feat: Add option to skip config validation

### LINKED ISSUE #1710: Allow device & shard count validation to be skipped

pawelpasterz:
Fixes #1710

2. launch test run with config:
    ```
    flank:
      max-test-shards: 5000000
    ```
3. you should see a validation error message `max-test-shards must be ...`
4. launch test run with config:
    ```
    flank:
      max-test-shards: 5000000
      skip-config-validation: true
    ```
5. test run proceeds normally

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/test_runner/src/main/kotlin/ftl/args/ValidateAndroidArgs.kt
+++ b/test_runner/src/main/kotlin/ftl/args/ValidateAndroidArgs.kt
@@ -15,7 +15,7 @@ import ftl.run.exception.FlankGeneralError
 import ftl.run.exception.IncompatibleTestDimensionError
 import java.io.File
 
-fun AndroidArgs.validate() = apply {
+fun AndroidArgs.validate() = if (shouldValidateConfig) apply {
     commonArgs.validate()
     assertDevicesSupported()
     assertShards()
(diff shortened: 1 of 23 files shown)


pawelpasterz:
@flank-it

Sloox (approved):

Sloox:
Ended up being more simple than we predicted during planning :)

test_runner/src/main/kotlin/ftl/args/ValidateAndroidArgs.kt
```diff
@@ -15,7 +15,7 @@ import ftl.run.exception.FlankGeneralError
 import ftl.run.exception.IncompatibleTestDimensionError
 import java.io.File
 
-fun AndroidArgs.validate() = apply {
+fun AndroidArgs.validate() = if (shouldValidateConfig) apply {
```
piotradamczyk5, line 18: `run` fits better in my opinion

piotradamczyk5 (approved):

PR #1724, opened 2021-03-23
refactor: Separate main function from MainCommand

### LINKED ISSUE #1723: Separate main function from MainCommand
In an abstract way,  the `main` function is a typical entry point for any `jvm` application. The CLI is one of the possible options for how the user can interact with the application. Those are two different things and should be separated.

As a developer, I want to have separated the main function from the picocli MainComand due to the SRP rule.

**Describe the solution**

the `ftl.Main.kt` file should contain only the entry point function:

```kotlin
fun main(args: Array<String>) {
    withGlobalExceptionHandling {
        CommandLine(Main()).execute(*args)
    }
}
```

Everything else, related to picocli command should go to the `ftl.cli.MainCommand.kt`

jan-gogo:
Fixes #1723 

## Test Plan

It's just refactoring, all unit tests should pass.

### DIFF
--- a/test_runner/src/main/kotlin/ftl/Main.kt
+++ b/test_runner/src/main/kotlin/ftl/Main.kt
@@ -1,56 +1,11 @@
 package ftl
 
-import ftl.cli.AuthCommand
-import ftl.cli.FirebaseCommand
-import ftl.cli.firebase.CancelCommand
-import ftl.cli.firebase.RefreshCommand
-import ftl.cli.firebase.test.AndroidCommand
-import ftl.cli.firebase.test.IPBlocksCommand
-import ftl.cli.firebase.test.IosCommand
-import ftl.cli.firebase.test.NetworkProfilesCommand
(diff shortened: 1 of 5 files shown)


Sloox (approved):

Sloox (dismissed):
Does it require any updates for tests? or atleast a documentaiton of what the change introduces (If anybesides neater code)

jan-gogo:
> Does it require any updates for tests? or atleast a documentaiton of what the change introduces (If anybesides neater code)

This is refactoring, so it doesn't bring any changes in logic, so it also doesn't require any new test. But it brings some updates in tests according to refactor changes.

The reason for this change is specified here:

* https://github.com/Flank/flank/issues/1723 (in the description)

The documentation already takes this change into account it is specified on the diagram:

* https://github.com/Flank/flank/tree/master/docs/hld/presentation/cli

pawelpasterz (approved):

piotradamczyk5 (approved):

Issue #1650, opened 2021-03-02
Add IT for #1649


pawelpasterz:
This is a follow-up issue.

We should implement IT that will cover case fixed in https://github.com/Flank/flank/pull/1649

IT should cover:
* test run where at least one shard has more than 50 tests
* since this is an issue related to test result API usage android test should be enough (iOS uses legacy, XML based approach)

PR #1730, opened 2021-03-23
test: Added app with many tests

piotradamczyk5:
Relates to #1650

Tests app with many tests could be built.

It is needed for example to add IT in #1650


test_projects/android/ops.sh
```diff
@@ -23,4 +23,8 @@ function cucumber_sample_app() {
   $DIR/../../flank-scripts/bash/flankScripts assemble android app --copy --generate --artifacts=buildCucumberSampleApp
 }
 
+function many_tests_apk() {
```
jan-gogo, line 26: Do we have any good reasons for supporting shell scripts, now? Wasn't a `flank-scripts` a replacement for the shell? 

This is not a change request, I am just thinking about adding an issue for removing any `ops.sh`

jan-gogo (approved):

flank-scripts/src/main/kotlin/flank/scripts/ops/assemble/android/BuildManyTestsApk.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    if (copy) copyManyTestsApk()
+}
+
+private fun copyManyTestsApk() {
+    val outputDir = Paths.get(flankFixturesTmpPath, "apk", "app_many_tests.apk")
+
+    if (!outputDir.parent.toFile().exists()) Files.createDirectories(outputDir.parent)
+
+    val assembleDirectory = Paths.get(androidTestProjectsPath, "app_many_tests", "build", "outputs", "apk", "debug", "app_many_tests.apk")
```
jan-gogo, line 27:
```suggestion
    val assembleDirectory = Paths.get(androidTestProjectsPath, "app_many_tests", "build", "outputs", "apk", "debug", "app_many_tests-debug.apk")
```

test_projects/android/ops.sh
piotradamczyk5: `ops.sh` just run `flankScripts` with the proper command it was added for backward compatibility, but sure we could remove it

test_projects/android/app_many_tests/src/androidTest/java/com/example/app/many/tests/ManyTests.kt
```diff
@@ -0,0 +1,308 @@
...
+    }
+
+    @Test
+    fun test58() {
+        assertEquals(true, "true".toBoolean())
+    }
+
+    @Test
+    fun test59() {
+        assertEquals(true, "true".toBoolean())
```
Sloox, line 305: Welldone what a pain in the ....

Sloox (approved):

test_projects/android/ops.sh
bootstraponline: +1 to not having shell scripts if possible

PR #1739, opened 2021-03-26
feat: Windows IOS integration tests

### LINKED ISSUE #1686: iOS IT Windows: Fix Windows Paths & Linking issues
Fix the issue with linking files (currently flank makes a symbolic link for the nm.exe file that cannot be used. We need an alternative)
**Describe the solution**
It is possible to simply copy the file and rename it appropriately

It needs to be tested that it works as intended.

Sloox:
Fixes #1686 

IT tests will work

## Checklist

### DIFF
--- a/test_runner/src/main/kotlin/ftl/ios/xctest/common/ParseSwiftTests.kt
+++ b/test_runner/src/main/kotlin/ftl/ios/xctest/common/ParseSwiftTests.kt
@@ -5,6 +5,7 @@ import flank.common.isLinux
 import flank.common.isMacOS
 import flank.common.isWindows
 import ftl.util.Bash
+import ftl.util.ShellEnvironment
 
 internal fun parseSwiftTests(binary: String): List<String> {
     installBinaries
@@ -28,7 +29,9 @@ internal fun parseSwiftTests(binary: String): List<String> {
     } else emptyList()
 
     // https://github.com/linkedin/bluepill/blob/37e7efa42472222b81adaa0e88f2bd82aa289b44/Source/Shared/BPXCTestFile.m#L17-18
-    val demangledOutput = Bash.execute(cmd, path)
+    val shell = if (isWindows) ShellEnvironment.Cmd else ShellEnvironment.Default
+
+    val demangledOutput = Bash.execute(cmd, path, shell)
     demangledOutput.lines().forEach { line ->
         // _T025EarlGreyExampleTestsSwift0abceD0C10testLayoutyyF ---> EarlGreyExampleTestsSwift.EarlGreyExampleSwiftTests.testLayout() -> ()
         // _T025EarlGreyExampleTestsSwift0abceD0C16testCustomActionyyF ---> EarlGreyExampleTestsSwift.EarlGreyExampleSwiftTests.testCustomAction() -> ()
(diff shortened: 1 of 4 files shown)


Sloox:
@flank-it

Sloox:
@flank-it

Sloox:
@flank-it

Sloox:
@flank-it

test_runner/src/main/kotlin/ftl/util/Bash.kt
```diff
@@ -6,16 +6,20 @@ import ftl.run.exception.FlankGeneralError
...
-
-        val process = ProcessBuilder(bashPath, "-c", cmd)
-            .redirectOutput(PIPE)
-            .redirectError(PIPE)
-            .start()
+        val process = ProcessBuilder(shellEnvironment.execution, "-c", cmd).also {
+            if (additionalPath.isNotEmpty()) {
+                val envs: MutableMap<String, String> = it.environment()
+                additionalPath.forEach { extra ->
+                    envs[extra.component1()] = extra.component2()
```
piotradamczyk5, line 19: why using `component()` you could use it directly
`val (key, value) = extra`

test_runner/src/main/kotlin/ftl/ios/xctest/common/ParseSwiftTests.kt
```diff
@@ -10,16 +14,24 @@ internal fun parseSwiftTests(binary: String): List<String> {
...
 
+    val path = if (isWindows) {
+        listOf(Pair("Path", "$appDataDirectory\\.flank\\"), Pair("LD_LIBRARY_PATH", "$appDataDirectory\\.flank\\"))
+    } else emptyList()
+
     // https://github.com/linkedin/bluepill/blob/37e7efa42472222b81adaa0e88f2bd82aa289b44/Source/Shared/BPXCTestFile.m#L17-18
-    val demangledOutput = Bash.execute(cmd)
+    val shell = if (isWindows) ShellEnvironment.Cmd else ShellEnvironment.Default
+
+    val demangledOutput = Bash.execute(cmd, path, shell)
```
piotradamczyk5, line 34: would not be better to have executed function inside `ShellEnvironment` ?

piotradamczyk5:
@flank-it

piotradamczyk5 (changes requested):
IT failed even for MacOS https://github.com/Flank/flank/actions/runs/705041576

test_runner/src/main/kotlin/ftl/ios/xctest/common/ParseSwiftTests.kt
Sloox: I think it is unnecessary as then we might as well create an entire new hierarchy. The sealed class is to help encapsulate the different shell environments and the Bash class simple executes what its been provided with. 
I understand that its the next logical step but if we go that far the entire class structure should be replaced entirely and reworked. :)
WDYT?

piotradamczyk5: we could leave it as it is, but for me 
`shell.execute(cmd, path)`
is much more readable than
`Bash.execute(cmd, path, shell)`

if you would like to leave it just resolve my comment

Sloox:
@flank-it

test_runner/src/main/kotlin/ftl/ios/xctest/common/ParseSwiftTests.kt
Sloox: @piotradamczyk5 I will create a ticket for refactoring this entire section. 
We need to look into how to handle commands and different shell environments.
https://github.com/Flank/flank/issues/1770

pawelpasterz (approved):

piotradamczyk5 (approved):

PR #1760, opened 2021-03-29
build: Fix dependabot security vulnerabilities

piotradamczyk5:
Fix dependabot security vulnerabilities


Sloox (approved):
can we decide on what title to use here either build or deps not both? It doesnt adhere to our standards.

piotradamczyk5:
> can we decide on what title to use here either build or deps not both? It doesnt adhere to our standards.

change to `build`

pawelpasterz (approved):

PR #1793, opened 2021-04-09
fix: Added workaround for uppercase project name

### LINKED ISSUE #1789: Non-lowercase project names not supported
Flank does not support Firebase project names that are not entirely lowercase. 
While tests start and finish just fine, at the end of the run Flank outputs: `Matrix is ERROR`. And the following:

```
  No cost. 0m


MatrixResultsReport
  1 / 1 (100.00%)
┌─────────┬───────────┬──────────┬─────────────────┬──────────────┐
│ OUTCOME │ MATRIX ID │ APP NAME │ TEST AXIS VALUE │ TEST DETAILS │
├─────────┼───────────┼──────────┼─────────────────┼──────────────┤
└─────────┴───────────┴──────────┴─────────────────┴──────────────┘
```

**To Reproduce**
1. Create Firebase project with non-lowercase name 
2. Set project name in `flank.yml`
3. Run flank

**Expected behavior**
Print matrix output as normal.
Normal behavior can be reproduced by setting the project name as all lowercase (even if the project is not entirely lowercase in Firebase). This executes just fine.

Test on latest flank release and snapshot. Also tested against gcloud (where it worked fine). Also tested by setting project name to all lower case (where it also worked fine).

piotradamczyk5:
Fixes #1789 

FTL API does not work properly with upper case letters in the project name, it does not report any issues, but results are empty. The project name could be just lower case, so this PR makes all projects names lower case, to make them work properly with FTL API

1. Run tests with the real project name, but make some letters upper case (ex. `Flank-Open-Source`)
2. All tests should run normally without any problems

## Checklist

- [x] Unit tested

### DIFF
--- a/test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
+++ b/test_runner/src/main/kotlin/ftl/args/ArgsHelper.kt
@@ -181,9 +181,10 @@ object ArgsHelper {
         return bucket
     }
 
-    fun getDefaultProjectIdOrNull(): String? = if (useMock) "mockProjectId"
+    fun getDefaultProjectIdOrNull() = (if (useMock) "mock-project-id" else getUserProjectId())?.toLowerCase()
+
     // Allow users control over project by checking using Google's logic first before falling back to JSON.
-    else fromUserProvidedCredentials()
+    private fun getUserProjectId(): String? = fromUserProvidedCredentials()
         ?: ServiceOptions.getDefaultProjectId()?.let { if (it.isBlank()) null else it }
         ?: fromDefaultCredentials()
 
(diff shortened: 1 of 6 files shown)


pawelpasterz:
According to https://cloud.google.com/sdk/gcloud/reference#--project `gcloud` does not consume project name but `project_id`. This PR makes it possible to use project name with flank but ONLY when project name is unique (meaning there is no `-ea123` or sth suffix in project id).
Based on that I think it would be great to add small note in documentation.

Sloox (approved):

axelzuziak-gogo (approved):

PR #1813, opened 2021-04-16
ci: Disabled estimations copying

piotradamczyk5:
Estimations are not copied when creating a new pull request.


piotradamczyk5:
I left zenhub data package in case we need it in future

Sloox (approved):

jan-gogo (approved):

PR #1825, opened 2021-04-21
test: Optimize quota usage in IT

pawelpasterz:
PR introduces a category approach in IT. Tests are divided into `AndroidTest` and `IosTest` category. It can be achieved with annotating tests with either `@Category(AndroidTest::class)` or `@Category(IosTest::class)` 

#### Changes:
There are slightly different rules applied to each category.
`AndroidTest` category:
* runs in parallel
* no fail-fast
* runs first

`IosTest` category:
* runs tests sequentially
* with fail-fast
* depends on `AndroidTest` category

For each category 2 new gradle tasks were created `instrumentationTestsAnroid` and `instrumentationTestsIos` -- each runs corresponding category.

`instrumentationTests` task was refactored and now it depends on android and ios tasks, forcing android category tests to run before iOS

#### Great, but what does it give us actually?
Flank project keeps facing exceeded quota problem, mainly due to iOS tests that use physical devices. 

With proposed changes, flank first runs android tests. If 'common' logic (meaning, shared by both android and ios) tests will fail, only virtual devices will be used. Since flank has like 2-3x more android tests we keep parallel settings to make it faster.

Once android category is done, ios starts. A combination of sequential run and `fail-fast: true` will terminate the test run with a minimal number of devices used to detect a problem.

To sum up:
:heavy_plus_sign: test coverage is the same
:heavy_plus_sign: we burn less quota if a problem is found
:heavy_minus_sign: full IT run takes more time, locally, on my machine its `16m 40s` old vs `19m 12s` new --  so it's about +15% longer time run

### DIFF
--- a/integration_tests/build.gradle.kts
+++ b/integration_tests/build.gradle.kts
@@ -52,7 +52,7 @@ tasks.register<Test>("integrationTestsAndroid") {
         events("skipped", "failed")
         exceptionFormat = TestExceptionFormat.FULL
     }
-    maxParallelForks = Runtime.getRuntime().availableProcessors() / 2
+    maxParallelForks = Runtime.getRuntime().availableProcessors()
 }
 
 tasks.register<Test>("integrationTestsIos") {
--- a/integration_tests/build.gradle.kts
+++ b/integration_tests/build.gradle.kts
@@ -38,17 +38,43 @@ tasks.test {
     }
 }
 
-tasks.register<Test>("integrationTests") {
+tasks.register<Test>("integrationTestsAndroid") {
     group = "Verification"
-    description = "Runs flank integration tests"
+    description = "Runs flank integration tests -- Android"
     filter {
         includeTestsMatching("*IT")
+    }
+    useJUnit {
+        includeCategories = setOf("integration.config.AndroidTest")
+        excludeCategories = setOf("integration.config.IosTest")
     }
     testLogging {
         events("skipped", "failed")
         exceptionFormat = TestExceptionFormat.FULL
     }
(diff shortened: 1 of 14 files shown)


integration_tests/build.gradle.kts
```diff
@@ -38,19 +38,45 @@ tasks.test {
...
     }
+    useJUnit {
+        includeCategories = setOf("integration.config.AndroidTest")
+        excludeCategories = setOf("integration.config.IosTest")
+    }
     testLogging {
         events("skipped", "failed")
         exceptionFormat = TestExceptionFormat.FULL
     }
     maxParallelForks = Runtime.getRuntime().availableProcessors() / 2
```
Sloox, line 55: Is this necessary? Lets remove this and see if it speeds it up.

Sloox: i.e.:
```suggestion
    maxParallelForks = Runtime.getRuntime().availableProcessors()
```

Sloox (approved):
Good idea 👍

integration_tests/build.gradle.kts
pawelpasterz: Why not, let's give it a try 👍🏻

pawelpasterz:
As a follow-up task we can think of implementing additional comment triggers for each category (i.e.: `@flank-it-android`, `@flank-it-ios`)

pawelpasterz:
> **Integration tests were triggered at 2021-04-21 08:26:16, you can track progress [here](https://github.com/Flank/flank/actions/runs/769976298)**

Damn 😅

PR #1835, opened 2021-04-22
feat: Add sharding implementation for corellium

### LINKED ISSUE #1801: Multi-module sharding algorithm
[README](https://github.com/Flank/flank/blob/1801_Multi-module_sharding_algorithm/corellium/shard/README.md)

jan-gogo:
Fixes #1801 

The implementation of multi-module sharding

[README](https://github.com/Flank/flank/blob/1801_Multi-module_sharding_algorithm/corellium/shard/README.md)

Unit test passes

## Checklist

- [x] Documented
- [x] Unit tested
- [ ] Update diagram URL in README file after approvals

### DIFF
--- a/corellium/shard/README.md
+++ b/corellium/shard/README.md
@@ -4,7 +4,7 @@ Depending on provided test cases duration, sharding algorithm will:
 
 * Split test cases from one test apk to run on many devices.
 * Group test cases from many test apks to run on a single device.
-* Mix both of the above options.
+* A mix both of the above options.
 
 ## Diagram
 
--- a/corellium/shard/README.md
+++ b/corellium/shard/README.md
@@ -3,7 +3,7 @@
 Depending on provided test cases duration, sharding algorithm will:
 
 * Split test cases from one test apk to run on many devices.
-* Group test cases from many test apks to run on a single device.
+* Group the test cases from many test apks to run on a single device.
 * A mix both of the above options.
 
 ## Diagram
--- a/corellium/shard/README.md
+++ b/corellium/shard/README.md
@@ -1,7 +1,9 @@
 # Sharding
 
-* Splitting test cases from one test apk to run on many devices.
-* Grouping test cases from many test apks to run on a single device.
+Depending on provided test cases duration, sharding algorithm will:
+
+* Split test cases from one test apk to run on many devices.
+* Group test cases from many test apks to run on a single device.
 * Mix both of the above options.
 
 ### Input
--- a/corellium/shard/README.md
+++ b/corellium/shard/README.md
@@ -6,6 +6,12 @@ Depending on provided test cases duration, sharding algorithm will:
 * Group test cases from many test apks to run on a single device.
 * Mix both of the above options.
 
+## Diagram
+
+![sharding_class_diagram](http://www.plantuml.com/plantuml/proxy?cache=no&fmt=svg&src=https://raw.githubusercontent.com/Flank/flank/1801_Multi-module_sharding_algorithm/docs/corellium/sharding-class.puml)
+
+## Example
+
 ### Input
 
 ```yaml
--- a/corellium/shard/README.md
+++ b/corellium/shard/README.md
@@ -1,6 +1,6 @@
 # Sharding
 
-Depending on provided test cases duration, sharding algorithm will:
+Depending on the provided test cases duration, the sharding algorithm will:
 
 * Split the test cases from one test apk to run on many devices.
 * Group the test cases from many test apks to run on a single device.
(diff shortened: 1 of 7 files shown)


corellium/shard/src/main/kotlin/flank/corellium/shard/Internal.kt
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * Create flat and easy to iterate list of [Chunk] without loosing info about app and test related to test case.
+ * @receiver The [List] of [Shard.App].
+ * @return The [List] of [Chunk] which is just different representation of input data.
+ */
+internal fun List<Shard.App>.mapToInternalChunks(): List<Chunk> =
+    mutableListOf<Chunk>().also { result ->
+        forEach { app ->
+            app.tests.forEach { test ->
+                test.cases.forEach { case ->
```
piotradamczyk5, line 16: triple `forEach` nested 😄 
could not we avoid this ?

corellium/shard/README.md
```diff
@@ -0,0 +1,100 @@
+# Sharding
+
+Depending on provided test cases duration, sharding algorithm will:
```
Sloox, line 3:
```suggestion
Depending on the provided test cases duration, the sharding algorithm will:
```

corellium/shard/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+* Split test cases from one test apk to run on many devices.
```
Sloox, line 5:
```suggestion
* Split the test cases from one test apk to run on many devices.
```

corellium/shard/README.md
```diff
@@ -0,0 +1,100 @@
...
+* Group test cases from many test apks to run on a single device.
```
Sloox, line 6:
```suggestion
* Group the test cases from many test apks to run on a single device.
```

corellium/shard/README.md
```diff
@@ -0,0 +1,100 @@
...
+* Mix both of the above options.
```
Sloox, line 7:
```suggestion
* A mix both of the above options.
```

corellium/shard/src/main/kotlin/flank/corellium/shard/Internal.kt
```diff
@@ -0,0 +1,105 @@
+package flank.corellium.shard
+
+import kotlin.math.min
+
+// Stage 1
+
+/**
+ * Create flat and easy to iterate list of [Chunk] without loosing info about app and test related to test case.
```
Sloox, line 8:
```suggestion
 * Create flat and easy to iterate list of [Chunk] without losing info about app and test related to test case.
```

corellium/shard/src/main/kotlin/flank/corellium/shard/Internal.kt
Sloox: It does look chaotic agreed.

corellium/shard/src/main/kotlin/flank/corellium/shard/Internal.kt
```diff
@@ -0,0 +1,105 @@
...
+        }.mapValues { (_, chunks) ->
+            chunks.groupBy { chunk ->
+                chunk.test
+            }
+        }
+    }
+
+    // Convert grouped chunks into the output structures.
+    return list.map { map ->
+        map.map { (app, tests) ->
```
Sloox, line 88: I know it might be necessary but 4 nested maps is crazy...

Sloox (changes requested):
Minor changes :)

corellium/shard/src/main/kotlin/flank/corellium/shard/Internal.kt
jan-gogo: It's not possible to avoid nested loops in this case.

The input structure `List<Shard.App>` is designed to not store redundant information, but for sharding-calculation is necessary to change the data representation to flat.

Anyway, It's just a simple mapper so I cannot understand why you are complaining, there is no single ugly condition here, just pure function.

corellium/shard/src/main/kotlin/flank/corellium/shard/Internal.kt
jan-gogo: Nope, it's not crazy, it's just mapping one data representation to another without any condition. 
The crazy will be adding the conditions inside these mappings or creating improper data structures just because of some additional mappings and loops in private or internal functions.

corellium/shard/src/main/kotlin/flank/corellium/shard/Internal.kt
piotradamczyk5: We are complaining because in our opinion this code is not so readable. I think that code review is not just about reviewing conditions.

Sloox (approved):

axelzuziak-gogo (approved):

PR #1875, opened 2021-04-26
refactor: Fetching artifacts

### LINKED ISSUE #1758: Data scratch - artifacts
# Plan

According to the specification described in epic [#1727](https://github.com/Flank/flank/issues/1727)
add abstraction and implementation for the following part of the data layer:

## Abstraction

`ftl/api/Artifacts.kt`

```kotlin
package ftl.api

val fetchArtifacts: Artifacts.Fetch get() = TODO()

object Artifacts {
    
    data class Identity(
        val gcsPathWithoutRootBucket: String,
        val gcsRootBucket: String,
        val regex: List<Regex>,
        val blobPath: String,
        val downloadPath: DownloadPath,
    )
    
    data class DownloadPath(
        val localResultDir: String,
        val useLocalResultDir: Boolean,
        val keepFilePath: Boolean,
    )
    
    interface Fetch: (Identity) -> List<String> 
}
```

## Target

* `FetchArtifacts.kt`

## Adapter

`ftl/adapter/GoogleArtifactsFetch.kt`

pawelpasterz:
Fixes #1758 

- Code is refactored according to the description in #1758
- File reference has been replaced and is acting as expected.

### DIFF
--- a/test_runner/build.gradle.kts
+++ b/test_runner/build.gradle.kts
@@ -345,4 +345,10 @@ tasks.test {
     maxHeapSize = "3072m"
     minHeapSize = "512m"
     dependsOn(":resolveArtifacts")
+
+    // deletes 'results' directory which contains remainings from unit tests
+    // think about better solution
+    doLast {
+        project.delete("results")
+    }
 }
--- a/test_runner/build.gradle.kts
+++ b/test_runner/build.gradle.kts
@@ -229,6 +229,7 @@ buildscript {
 
 tasks.withType<KotlinCompile> {
     kotlinOptions.jvmTarget = "1.8"
+    kotlinOptions.freeCompilerArgs = listOf("-Xinline-classes")
 }
 
 // https://github.com/gradle/kotlin-dsl/blob/master/samples/task-dependencies/build.gradle.kts#L41
--- a/test_runner/build.gradle.kts
+++ b/test_runner/build.gradle.kts
@@ -229,7 +229,6 @@ buildscript {
 
 tasks.withType<KotlinCompile> {
     kotlinOptions.jvmTarget = "1.8"
-    kotlinOptions.freeCompilerArgs = listOf("-Xinline-classes")
 }
 
 // https://github.com/gradle/kotlin-dsl/blob/master/samples/task-dependencies/build.gradle.kts#L41
--- a/test_runner/src/main/kotlin/ftl/adapter/google/FetchArtifacts.kt
+++ b/test_runner/src/main/kotlin/ftl/adapter/google/FetchArtifacts.kt
@@ -7,11 +7,11 @@ import ftl.client.google.GcStorage
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
-import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.runBlocking
 import java.nio.file.Path
 import java.nio.file.Paths
 
-suspend fun fetchArtifacts(identity: Identity): Pair<String, List<String>> = coroutineScope {
+fun fetchArtifacts(identity: Identity): Pair<String, List<String>> = runBlocking {
     val prefix = Storage.BlobListOption.prefix(identity.gcsPathWithoutRootBucket)
     val result = GcStorage.storage.list(identity.gcsRootBucket, prefix, fields)
 
--- a/test_runner/src/main/kotlin/ftl/adapter/google/FetchArtifacts.kt
+++ b/test_runner/src/main/kotlin/ftl/adapter/google/FetchArtifacts.kt
@@ -14,6 +14,7 @@ import java.nio.file.Paths
 
 suspend fun fetchArtifacts(identity: Identity): Pair<MatrixId, List<String>> = coroutineScope {
     val prefix = Storage.BlobListOption.prefix(identity.gcsPathWithoutRootBucket)
+    val fields = Storage.BlobListOption.fields(Storage.BlobField.NAME)
     val result = GcStorage.storage.list(identity.gcsRootBucket, prefix, fields)
 
     MatrixId(identity.matrixId) to result.iterateAll().mapNotNull { blob ->
@@ -45,4 +46,4 @@ internal fun getFilePathToDownload(downloadPath: DownloadPath, blobPath: String)
     return Paths.get("$localDir/$objName/$matrixName/$deviceName/$filePathName/$fileName")
 }
 
-private val fields = Storage.BlobListOption.fields(Storage.BlobField.NAME)
+// private val fields = Storage.BlobListOption.fields(Storage.BlobField.NAME)
--- a/test_runner/src/main/kotlin/ftl/adapter/google/FetchArtifacts.kt
+++ b/test_runner/src/main/kotlin/ftl/adapter/google/FetchArtifacts.kt
@@ -1,0 +1,48 @@
+package ftl.adapter.google
+
+import com.google.cloud.storage.Storage
+import ftl.api.Artifacts.DownloadPath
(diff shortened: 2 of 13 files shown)


test_runner/src/main/kotlin/ftl/api/Artifacts.kt
```diff
@@ -0,0 +1,34 @@
...
+
+val fetchArtifacts: Artifacts.Fetch get() = GcFetchArtifacts
+
+object Artifacts {
+
+    data class Identity(
+        val gcsPathWithoutRootBucket: String,
+        val gcsRootBucket: String,
+        val regex: List<Regex>,
+        val blobPath: String, // todo not needed?
```
pawelpasterz, line 13: I think we do not need it. Every `Identity` can have multiple blobs
@jan-gogo

test_runner/src/main/kotlin/ftl/api/Artifacts.kt
```diff
@@ -0,0 +1,34 @@
...
+        val downloadPath: DownloadPath,
+        val matrixId: String
```
pawelpasterz, line 15: Added `matrixId` to persists info about matrix from which files are downloaded. It is used to update `download` flag once all files are fetched. Let me know if this violates any assumptions.
https://github.com/Flank/flank/pull/1875/files#diff-f91686b92f886f88f9b047c859da24c40bc6ec3ee8686ef5eec4bd5259f91de2R45

@jan-gogo

test_runner/src/main/kotlin/ftl/json/SavedMatrix.kt
```diff
@@ -28,7 +28,9 @@ data class SavedMatrix(
     val state: String = "",
     val gcsPath: String = "",
     val webLink: String = "",
-    val downloaded: Boolean = false,
+    // this is variable intentionally, this is workaround to pass changes while fetching artifacts
+    // todo think about different solution
+    var downloaded: Boolean = false,
```
pawelpasterz, line 33: Changed to `var`. The old implementation had a bug and none of `SavedMatrix` had `downloaded` updated during artifact fetch. This is a quick fix for this issue. We should think about a different solution but in a separate issue. 

Might be fixed within another refactor task (?)

@jan-gogo

test_runner/src/main/kotlin/ftl/api/Artifacts.kt
```diff
@@ -0,0 +1,34 @@
...
+     * Fetches all files for provided [Identity].
+     * @return Matrix ID of which files were downloaded
+     * associated with list of downloaded files
+     */
+    interface Fetch {
+        suspend operator fun invoke(input: Identity): Pair<MatrixId, List<String>>
+    }
+}
+
+inline class MatrixId(val value: String)
```
pawelpasterz, line 34: Inline classes as an experiment, let me know what do you think. I can change it to simple string ofc.

@jan-gogo

test_runner/build.gradle.kts
```diff
@@ -229,6 +229,7 @@ buildscript {
 
 tasks.withType<KotlinCompile> {
     kotlinOptions.jvmTarget = "1.8"
+    kotlinOptions.freeCompilerArgs = listOf("-Xinline-classes")
```
Sloox, line 232: To get rid of warnings?

test_runner/src/main/kotlin/ftl/adapter/google/FetchArtifacts.kt
```diff
@@ -0,0 +1,48 @@
...
+
+internal fun getFilePathToDownload(downloadPath: DownloadPath, blobPath: String): Path {
+    val localDir = downloadPath.localResultDir
+    val parsed = Paths.get(blobPath)
+    val objName = if (downloadPath.useLocalResultDir) "" else parsed.getName(0).toString()
+    // for iOS it is shardName, remove this comment after FTL introduce server side sharding for iOS
+    val matrixName = parsed.getName(1).toString()
+    val fileName = parsed.fileName.toString()
+    val deviceName = parsed.getName(2).toString().takeUnless { it == fileName }.orEmpty()
+    val filePathName = if (downloadPath.keepFilePath) parsed.parent.drop(3).joinToString("/") else ""
```
Sloox, line 43: Do you think its worth it to make these int variables constants with names for a little more clarity?

test_runner/src/main/kotlin/ftl/api/Artifacts.kt
Sloox: @pawelpasterz besides the idea of a performance increase(which i am unsure is necessary in our case) what benefit do they bring?

pawelpasterz: @Sloox Intention is rather code robustness. The inline class brings more information on what a primitive (string in that case) represents. It will be unwrapped to `String` (during compilation) so I think there is no benefit in performance.

It's definitely not something that I need to add, I'm just curious about your opinion.

test_runner/build.gradle.kts
pawelpasterz: It's required to compile inline classes

test_runner/src/main/kotlin/ftl/adapter/google/FetchArtifacts.kt
pawelpasterz: I think I could change it to extension functions with self-explanatory names, WDYT?

jan-gogo: Those ints are explained by the value names so adding constants for them will be an unneeded redundancy. IMO this implementation looks ok.

test_runner/src/main/kotlin/ftl/api/Artifacts.kt
jan-gogo: Inline classes could be useful for further issues when it will come to wrapping direct structured output.

But for this case, there is no need for a dedicated type for matrix id, because there is no type checking on it. The `private typealias` is sufficient.

test_runner/src/main/kotlin/ftl/run/common/FetchArtifacts.kt
```diff
@@ -1,76 +1,50 @@
...
+        .filter {
+            val finished = it.state == MatrixState.FINISHED
+            val notDownloaded = !it.downloaded
+            finished && notDownloaded
+        }
+        .map {
+            Identity(
+                gcsPathWithoutRootBucket = it.gcsPathWithoutRootBucket,
+                gcsRootBucket = it.gcsRootBucket,
+                regex = Artifacts.regexList(args),
```
jan-gogo, line 37:
```suggestion
    val artifactRegexList = Artifacts.regexList(args)

    matrixMap.map.values
        .filter {
            val finished = it.state == MatrixState.FINISHED
            val notDownloaded = !it.downloaded
            finished && notDownloaded
        }
        .map {
            Identity(
                gcsPathWithoutRootBucket = it.gcsPathWithoutRootBucket,
                gcsRootBucket = it.gcsRootBucket,
                regex = artifactRegexList,
```

The value returned by `Artifacts.regexList(args)` is immutable so it can be reused over all the all identities

pawelpasterz: oh nice catch, thanks!

pawelpasterz:
@flank-it

test_runner/src/main/kotlin/ftl/adapter/google/FetchArtifacts.kt
Sloox: I am fine with it as it is too :)

Sloox (approved):

adamfilipow92 (approved):

PR #1897, opened 2021-05-05
feat: Domain layer implementation for android test run Corellium MVP

### LINKED ISSUE #1802: Corellium domain layer
Input: shards and corresponding apks structure
Body: Invoking low-level functions for: initializing devices, running tests
Output: log streams

jan-gogo:
Fixes #1802 

This PR adds a domain logic for `RunTestCorelliumAndroid`.

List of features:
* Parsing test cases & other info from apk.
* Creating and/or invoking android instances on demand.
* Advanced sharding.
  * Many apps and tests in one shard.
  * Tests from one app in many shards.
* Dumping shards to file.
* Executing tests
* Generating JUnitReport.xml

What is currently not supported:
* Smart sharding (using test case durations from previous runs).
* Flaky tests.
* Cleaning up devices after tests.
* Some other not listed stuff.

Module [README.md](https://github.com/Flank/flank/blob/corellium-domain/corellium/domain/README.md)

Unit tests pass.
`$ flank corellium test android run -c="flank-corellium-android.yml"` is invoking a fully integrated feature.

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/corellium/domain/README.md
+++ b/corellium/domain/README.md
@@ -1,0 +1,34 @@
+# Flank-Corellium Domain Layer
+
+This module is specifying public API and internal implementation of flank-corellium use-cases.
+
+## Public API
+
+Functions and structures which are specified directly inside [flank.corellium.domain](./src/main/kotlin/flank/corellium/domain) package.
+
+## Internal implementation
+
+Everything inside nested packages of [flank.corellium.domain](./src/main/kotlin/flank/corellium/domain).
+
+## Design
+
+### Stateful execution
+
+Following specification is suitable for complicated long-running use-cases,
+when becomes convenient to split execution into smaller atomic chunks of work.
+
+#### Definition:
+
+* `Execution` is process which is creating initial `State`, and is running the set of `steps` on it in specific `Context`.
+* `Execution` can pass `Context` to `Step` if needed.
+* `Step` is a suspendable operation that is receiving `State` as the only argument.
+* `Step` must return received or new `State`.
+* `Step` can generate side effects.
+* `State` is a structure used for sharing data between preceding and following `steps`.
+* `Context` is providing arguments and functions for `step`.
+
+#### Utility:
+
+`Transform.kt`
+* [local link](src/main/kotlin/flank/corellium/domain/util/Transform.kt)
+* [github link](https://github.com/Flank/flank/blob/master/corellium/domain/src/main/kotlin/flank/corellium/domain/util/Transform.kt)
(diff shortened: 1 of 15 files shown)


corellium/adapter/src/test/kotlin/flank/corellium/adapter/Example.kt
```diff
@@ -85,8 +86,8 @@ fun main() {
         )
 
         println("* Executing tests")
-        api.executeTest(testPlan).collect { line ->
-            print(line)
+        api.executeTest(testPlan).forEach { flow ->
+            launch { flow.collect { line -> println(line) } }
         }
```
piotradamczyk5, line 91: is there any reason why you did not merge flow? multiple collects is not so common way to do such operation

corellium/adapter/src/test/kotlin/flank/corellium/adapter/Example.kt
```diff
@@ -85,8 +86,8 @@ fun main() {
         )
 
         println("* Executing tests")
-        api.executeTest(testPlan).collect { line ->
-            print(line)
+        api.executeTest(testPlan).forEach { flow ->
+            launch { flow.collect { line -> println(line) } }
```
piotradamczyk5, line 90: is there any reason why you did not merge flow? multiple collects is not so common way to do such operation

piotradamczyk5 (dismissed):

corellium/adapter/src/test/kotlin/flank/corellium/adapter/Example.kt
jan-gogo: Each flow from the list is representing output from a different android instance (virtual device) which could be treated as a single test suite with many test case results. Each console is available immediately but the results not, so that is why `List<Flow<String>>`. But if you are asking just about printing lines in this example, it's not production code. Just an example use of the current implementation that will be removed soon before MVP finish.

jan-gogo: But of course, you are right that technically could be a merge of flows instead of collect calls in the loop.

piotradamczyk5: so could you change it to 
`api.executeTest(testPlan).merge().collect { line -> println(line) }`?

jan-gogo: Do not afraid, I will remove it soon, it's not a production code, just an example for testing purposes.

piotradamczyk5: I do not see a point to make any code review for this code if all my propositions (then and now) are dropped, because "it is not production code"

jan-gogo: Pls, notice that this PR was and still is a draft. Drafts by design are not ready to review.

Besides, that here is PR which is removing that example code https://github.com/Flank/flank/pull/1933.

pawelpasterz (approved):

adamfilipow92 (approved):

PR #1920, opened 2021-05-12
refactor: Move shards calculation to separated module

jan-gogo:
Related to #1818 

This will allow distinguishing shard calculation and shard dumping as different modules that share the same data scheme.
To meeting SRP on module level.

Unit test passes

## Checklist

- [x] Documented
- [x] Unit tested


Sloox (approved):
:+1:

pawelpasterz (approved):

PR #1923, opened 2021-05-12
feat: Obfuscate shards

jan-gogo:
Related to #1818

Unit tests pass.

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/corellium/shard/obfuscate/src/main/kotlin/flank/corellium/shard/Obfuscate.kt
+++ b/corellium/shard/obfuscate/src/main/kotlin/flank/corellium/shard/Obfuscate.kt
@@ -10,13 +10,18 @@ package flank.corellium.shard
 fun Shards.obfuscate(): Shards =
     map { shard ->
         shard.map { app ->
-            app.copy(tests = app.tests.map { test ->
-                test.copy(cases = test.cases.map { case ->
-                    case.copy(name = obfuscationMappings.obfuscateAndroidTestName(case.name))
-                })
-            })
+            app.copy(
+                tests = app.tests.map { test ->
+                    test.copy(
+                        cases = test.cases.map { case ->
+                            case.copy(
+                                name = obfuscationMappings.obfuscateAndroidTestName(case.name)
+                            )
+                        }
+                    )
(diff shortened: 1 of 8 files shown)


corellium/shard/obfuscate/src/main/kotlin/flank/corellium/shard/Obfuscate.kt
```diff
@@ -0,0 +1,27 @@
+package flank.corellium.shard
+
+/**
+ * Obfuscate each [Shard.Test.Case.name] value.
+ * Use for security purpose.
+ *
+ * @receiver calculated shards
+ * @return obfuscated shards
+ */
+fun Shards.obfuscate(): Shards =
```
pawelpasterz, line 10: Hm... this code is quite hard to read and might be difficult to maintain in the future

jan-gogo: I agree that our lint rules are annoying. The first version was:
```
fun Shards.obfuscate(): Shards =
    map { shard ->
        shard.map { app ->
            app.copy(tests = app.tests.map { test ->
                test.copy(cases = test.cases.map { case ->
                    case.copy(name = obfuscationMappings.obfuscateAndroidTestName(case.name))
                })
            })
        }
    }
```

@pawelpasterz Any propositions?

pawelpasterz: TBH I never thought about linter. What I meant - code in this form may be error prone. I'm aware it's partially enforced by structure of `Shard` (and that is fine).
That's why it's just a comment, not a change request -- I point there might be a problem here in the future, but don't have better solution on my mind

jan-gogo: Ok. I have added explicit types and some comments that explain the meaning of this function.

pawelpasterz (approved):

Sloox (approved):

PR #1934, opened 2021-05-13
ci: Make IT results independent of os versions

pawelpasterz:
Processing result logic is now independent of OS version


pawelpasterz:
@flank-it

adamfilipow92 (approved):

piotradamczyk5 (approved):

Issue #1869, opened 2021-04-26
Structural output - RunDoctorIos


jan-gogo:
# Plan

According to the specification described in epic #1728, implement the structural output for the command and move the console logging to layer CLI.

# Command

`RunDoctorIos`

pawelpasterz:
Closed in https://github.com/Flank/flank/pull/1937

PR #1937, opened 2021-05-16
refactor: Run doctor command

### LINKED ISSUE #1868: Structural output - RunDoctorAndroid
# Plan

According to the specification described in epic #1728, implement the structural output for the command and move the console logging to layer CLI.

# Command

`RunDoctorAndroid`

pawelpasterz:
Fixes #1868 #1869

There is no better way to test it than breaking the config yml. Then run:
* `flank firebase test android doctor -c [path to your yml]`
* `flank firebase test ios doctor -c [path to your yml]`
* add `--fix`
* 🎉


test_runner/src/main/kotlin/ftl/doctor/Doctor.kt
```diff
@@ -5,43 +5,49 @@ import com.google.common.annotations.VisibleForTesting
...
+    }.filter { it.second.isNotEmpty() }.toMap(),
+    invalidDevices = validateDevices().orEmpty()
+)
 
-private fun JsonNode.validateTopLevelKeys(args: IArgs.ICompanion) =
+private fun JsonNode.validateTopLevelKeys(args: IArgs.ICompanion): List<String> =
     (parseArgs().keys - args.validArgs.keys)
         .takeIf { it.isNotEmpty() }
-        ?.let { unknownKeys -> "Unknown top level keys: $unknownKeys\n" }
+        ?.map { it }
```
piotradamczyk5, line 50: what is the purpose of this `map`?

test_runner/src/main/kotlin/ftl/domain/DoctorErrors.kt
```diff
@@ -0,0 +1,45 @@
...
+        else buildString {
+            parsingErrors.forEach { appendLine(it) }
+            invalidDevices
+                .map { "Warning: Version should be string $GCLOUD_NODE -> $DEVICES_NODE[${it.model}] -> $VERSION_NODE[${it.version}]" }
+                .forEach { appendLine(it) }
+            if (topLevelUnknownKeys.isNotEmpty()) appendLine("Unknown top level keys: $topLevelUnknownKeys")
+            nestedUnknownKeys.forEach { (topLevelKey, keys) ->
+                appendLine("Unknown keys in $topLevelKey -> $keys")
+            }
+        }.trim()
```
piotradamczyk5, line 33: it should be an extension, `data class` should not have functions (linter should complain also)

pawelpasterz: Good point 👍🏻 Thanks

test_runner/src/main/kotlin/ftl/doctor/Doctor.kt
pawelpasterz: That is an excellent question...and I don't know 😅 will be removed, Thanks!

piotradamczyk5 (changes requested):
some problem with double printing under investigation

test_runner/src/main/kotlin/ftl/doctor/Doctor.kt
adamfilipow92: We expecting ```List``` as result of function but ```(parseArgs().keys - args.validArgs.keys)``` returns ```Set```. I will change this map to ```.toList()```

piotradamczyk5 (approved):

piotradamczyk5 (approved):

jan-gogo (changes requested):
`DoctorErrors.kt` cannot be placed directly inside `domain` package because it's not a top-level function or scope. But as I see `RunDoctorAndroid` and `RunDoctorIos` looks similar, the only difference is the first argument for `validateYaml`. So my proposition is to merge `RunDoctorIos` and `RunDoctorAndroid` into one `RunDoctor` create parameterized `RunDoctor.invoke(args: IArgs.ICompanion)`, and put `DoctorErrors` inside `RunDoctor` under `Errors` name, so in result there will be a `RunDoctor.Errors` structure. Be aware that `RunDoctor.Errors.summary()` mapper is a part of `presentation.cli`, not a `domain`.

Sloox (approved):

test_runner/src/main/kotlin/ftl/domain/RunDoctor.kt
```diff
@@ -0,0 +1,48 @@
...
+    if (fix) processValidation(Paths.get(configPath))
+    validationResult.out()
+    if (!validationResult.isEmpty()) throw YmlValidationError()
+}
+
+fun RunDoctor.Error.isEmpty() =
+    parsingErrors.isEmpty() && topLevelUnknownKeys.isEmpty() &&
+        nestedUnknownKeys.isEmpty() && invalidDevices.isEmpty()
+
+operator fun RunDoctor.Error.plus(right: RunDoctor.Error) = RunDoctor.Error(
```
jan-gogo, line 43: This is not a top-level function, it shouldn't be here.

test_runner/src/main/kotlin/ftl/domain/RunDoctor.kt
```diff
@@ -0,0 +1,48 @@
...
+    val validationResult = validateYaml(args, path = Paths.get(configPath))
+    if (fix) processValidation(Paths.get(configPath))
+    validationResult.out()
+    if (!validationResult.isEmpty()) throw YmlValidationError()
+}
+
+fun RunDoctor.Error.isEmpty() =
+    parsingErrors.isEmpty() && topLevelUnknownKeys.isEmpty() &&
+        nestedUnknownKeys.isEmpty() && invalidDevices.isEmpty()
+
```
jan-gogo, line 42:
```suggestion
```
This function is redundant due to `error == RunDoctor.Error.EMPTY` operation.

jan-gogo (changes requested):

test_runner/src/main/kotlin/ftl/domain/RunDoctor.kt
adamfilipow92: Thanks! Fixed! 👍

PR #1948, opened 2021-05-19
refactor: Move am instrument commad formatter to dedicated module

jan-gogo:
This patch fixes inconsequence in the `AndroidTestPlan` API where `AndroidTestPlan.Execute` is receiving structured shards as input but producing raw `am instrument -r -w` log lines as output.

As a result of this fix:
* API of `AndroidTestPlan.Execute` is normalized to operate on raw input and output.
* `prepareRunCommand` is renamed to `formatAmInstrumentCommand` and moved to `:corellium:instrument:log` module so the adapter layer is a slightly thinner.
* `:corellium:log` is moved to `:corellium:instrument:log` for consistency.
* The responsibility for calling `formatAmInstrumentCommand` is moved from the `ExecuteAndroidTestPlan` adapter to the `ExecuteTests` step, so the domain logic is controlling a little bit more.
* Overall SRP ratio increases 🎉


adamfilipow92 (approved):

Sloox (approved):
:+1:

PR #1961, opened 2021-05-24
docs: Add implementation section to architecture doc

### LINKED ISSUE #1960: Add implementation section to architecture doc

jan-gogo:
Fixes #1960 

This PR is adding a section about how to meet architecture requirements in implementation.

added section

## Checklist

- [x] Update diagram links before merge


pawelpasterz (approved):

docs/architecture.md
```diff
@@ -263,3 +270,46 @@ one of or many:
...
+# Implementation <a name="implementation"/>
+
+For convenience and clarity, the code should be written in a functional programming style. It's mandatory to avoid the OOP style which almost always makes things much more complicated than should be.
+
+## Public API <a name="implementation_public_api"/>
+
+Any application or library always have a public API and internal/private part. For convenience keep public functions and structures in the root package, so the API will be easy to find. Additionally, if the `component`:
+
+* is providing reach public API with additional structures. - Is mandatory to distinct the public structures and functions from internal implementation which should be kept in nested package(s).
+* is just a simple tool with a compact implementation that is not specifying many structures. - Private implementation can be kept in the same file, just behind the public API or even whole tool can be delivered as one public function if the implementation is simple enough.
```
Sloox, line 283:
```suggestion
* is just a simple tool with a compact implementation that is not specifying many structures. - Private implementations can be kept in the same file, just behind the public API or even the whole tool can be delivered as one public function if the implementation is simple enough.
```

docs/architecture.md
```diff
@@ -263,3 +270,46 @@ one of or many:
...
+
+DO NOT keep many public functions along with internal implementation in the same file or package, because it is messing up the public API, which makes code harder to analyze and navigate.
```
Sloox, line 285:
```suggestion
DO NOT keep multiple public functions along with internal implementations in the same file or package, because it is messes up the public API, which makes code harder to analyze and navigate.
```

docs/architecture.md
```diff
@@ -263,3 +270,46 @@ one of or many:
...
+
+## Components composition <a name="components_composition"/>
+
+Business logic shouldn't implement complicated tools on its own because it is messing up crucial high-level implementation making it harder to understand. Instead of this, it should be decomposed into high-level use-case implementation that is operating on tools provided by specialized components.
```
Sloox, line 289:
```suggestion
Business logic shouldn't implement complicated tools on its own because it is can mess up crucial high-level implementations making it harder to understand. Instead, it should be decomposed into high-level use-case implementations that operate on tools provided by specialized components.
```

docs/architecture.md
```diff
@@ -263,3 +270,46 @@ one of or many:
...
+
+## Code composition <a name="code_composition"/>
+
+Typically, when huge features are divided into smaller functions and one of those functions is a (public) root, the functions can be composed in two different ways.
+
+### Vertical <a name="code_composition_vertical"/>
+
+The preceding function is calling the following, so the composition of functions is similar to the linked list.
```
Sloox, line 297:
```suggestion
The preceding function is calling the following, so the composition of functions is similar to a linked list.
```

docs/architecture.md
```diff
@@ -263,3 +270,46 @@ one of or many:
...
+
+Try to **AVOID** this pattern where possible, especially in business logic. In some situations it can be even worse than one huge monolithic function with comments, for example when internal functions are not ordered correctly. Understanding the feature composed in vertical style, almost always require analyzing the whole chain of functions which typically is not efficient.
+
+### Horizontal <a name="code_composition_horizontal"/>
+
+Root function is controlling independent internal and specialized functions.
+
+![horizontal-composition](http://www.plantuml.com/plantuml/proxy?cache=no&fmt=svg&src=https://raw.githubusercontent.com/Flank/flank/1960_Add_implementation_section_to_architecture_doc/docs/hld/horizontal-composition.puml)
+
+This approach is giving a fast overview of high-level implementation but is hiding the details not important from the high-level perspective. Comparing to `vertical` composition where the cost of manual access to internal functions (jumping on references in IDE) in the worst-case scenario is `n`, the horizontal composition almost always gives `1` on the same layer (or `2` taking private functions into account if exist).
```
Sloox, line 309:
```suggestion
This approach gives a fast overview of high-level implementation but is hiding the details not important from the high-level perspective. Comparing to `vertical` composition where the cost of manual access to internal functions (jumping on references in IDE) in the worst-case scenario is `n`, the horizontal composition almost always gives `1` on the same layer (or `2` taking private functions into account if exist).
```

Sloox (changes requested):
👍

jan-gogo:
> 👍

@Sloox Thx for the corrections ❤️. All applied

Sloox (approved):

PR #1967, opened 2021-05-25
fix: Filter devices without supported versions

### LINKED ISSUE #1892: Flank displays different list of devices compared to `gcloud`
Flank seems not to perform any filtering when it comes to the model list. There is no verification if a device has a version that can be used. That leads to differences in the devices list displayed to the user comparing to `gcloud`.

##### flank

https://github.com/Flank/flank/blob/b3c4a7616f3ac346e151ed65b9fff940ceaf4aeb/test_runner/src/main/kotlin/ftl/client/google/AndroidCatalog.kt#L24-L30

#### What needs to be done?
* verify if this is a real bug or a feature (maybe it was implemented that way intentionally -- reach out Matt)
* implement fix
* update JSONs in fixture directory [ANDROID](https://github.com/Flank/flank/blob/master/test_runner/src/test/kotlin/ftl/fixtures/android_catalog.json) [iOS](https://github.com/Flank/flank/blob/master/test_runner/src/test/kotlin/ftl/fixtures/ios_catalog.json) (you can use https://github.com/Flank/flank/blob/master/test_runner/src/test/kotlin/task/UpdateCatalogFixtures.kt)

pawelpasterz:
Fixes #1892 

2. run `flank firebase test ios models list` and compare with `gcloud firebase test ios models list` -- there should be no different
3. run `flank firebase test android models list` and compare with `gcloud firebase test android models list` -- there should be no different

## Checklist

- [x] Unit tested

### DIFF
--- a/test_runner/src/main/kotlin/ftl/client/google/AndroidCatalog.kt
+++ b/test_runner/src/main/kotlin/ftl/client/google/AndroidCatalog.kt
@@ -26,19 +26,26 @@ object AndroidCatalog {
             .setProjectId(projectId)
             .executeWithRetry()
             .androidDeviceCatalog
+            .filterDevicesWithoutSupportedVersions()
     }
+
+    private fun AndroidDeviceCatalog.filterDevicesWithoutSupportedVersions() =
+        setModels(models.filterNotNull().filter { it.supportedVersionIds?.isNotEmpty() ?: false })
 
     fun getModels(projectId: String): List<AndroidModel> = deviceCatalog(projectId).models.orEmpty()
 
     fun supportedVersionsAsTable(projectId: String) = fetchAndroidOsVersion(projectId).toCliTable()
 
-    fun describeSoftwareVersion(projectId: String, versionId: String) = fetchAndroidOsVersion(projectId).getDescription(versionId)
+    fun describeSoftwareVersion(projectId: String, versionId: String) =
+        fetchAndroidOsVersion(projectId).getDescription(versionId)
 
     private fun getVersionsList(projectId: String) = deviceCatalog(projectId).versions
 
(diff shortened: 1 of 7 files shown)


Sloox (approved):

test_runner/src/main/kotlin/ftl/client/google/AndroidCatalog.kt
```diff
@@ -26,19 +26,26 @@ object AndroidCatalog {
...
+    private fun AndroidDeviceCatalog.filterDevicesWithoutSupportedVersions() =
+        setModels(models.filterNotNull().filter { it.supportedVersionIds?.isNotEmpty() ?: false })
+
     fun getModels(projectId: String): List<AndroidModel> = deviceCatalog(projectId).models.orEmpty()
 
     fun supportedVersionsAsTable(projectId: String) = fetchAndroidOsVersion(projectId).toCliTable()
 
-    fun describeSoftwareVersion(projectId: String, versionId: String) = fetchAndroidOsVersion(projectId).getDescription(versionId)
+    fun describeSoftwareVersion(projectId: String, versionId: String) =
+        fetchAndroidOsVersion(projectId).getDescription(versionId)
```
piotradamczyk5, line 40: I would prefer:
```
fun describeSoftwareVersion(
    projectId: String, 
    versionId: String
) = fetchAndroidOsVersion(projectId).getDescription(versionId)
```
but this is just a suggestion, please check other places

piotradamczyk5 (approved):

test_runner/src/main/kotlin/ftl/client/google/AndroidCatalog.kt
pawelpasterz: Ah sorry, it was merged before I implemented changes. Will do it within another PR

PR #2004, opened 2021-06-03
feat: Additional config options for test pairs

### LINKED ISSUE #1815: Allow fine turning configurations per APK pair whose tests are ran

As a consumer of the Flank project, I want to be able to override configures per test APK in `additional-app-test-apks` so I can run tests for different modules with different configurations.

My personal use case is:

1. I want to specify which tests to run so Flank doesn't need to download and parse the APKs from the gcloud bucket.
2. We'd like to runs different modules on different devices.

**Describe the solution**

I'll attach a SDD describing a solution shortly.

Again see the SDD.

pawelpasterz:
Fixes #1815

This PR introduces a couple of changes:
1. from now on each android test context will have its own copy of `AndroidConfig`
2. each copy will be overwritten with values defined by the user
3. each copy will be validated

With the above enhancements, we can extend `additional-test-app-pairs` easier in the future. For now, only community-requested options are implemented.

1. Run `./gradlew flankFullRun`
2. Example config:
    ```yaml
    gcloud:
      app: ...
      test: ...
    flank:
      max-test-shards: 2
      additional-app-test-apks:
        - test: ...
          max-test-shards: 5
        - test: ...
          device:
            - model: Nexus6P
              version: 27
    ```
3. Yous should receive 3 matrices with different set up:
    1. 2 shards, NexusLowRes device (default one)
    2. 5 shards, NexusLowRes device
    3. 2 shards, Nexus6P device
## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/test_runner/src/main/kotlin/ftl/args/ArgsToString.kt
+++ b/test_runner/src/main/kotlin/ftl/args/ArgsToString.kt
@@ -58,7 +58,7 @@ object ArgsToString {
                         .map { it.replace("\n", "\n  ") }
                         .forEach { append(it) }
                 }
-            }
+            }.trimEnd()
         }
     }
 }
--- a/test_runner/src/main/kotlin/ftl/args/ArgsToString.kt
+++ b/test_runner/src/main/kotlin/ftl/args/ArgsToString.kt
@@ -3,35 +3,62 @@ package ftl.args
 import ftl.args.yml.AppTestPair
 
 private const val NEW_LINE = '\n'
+private const val SPACESx8 = "        "
+private const val SPACESx10 = "          "
 
 object ArgsToString {
 
-    fun mapToString(map: Map<String, String>?): String {
+    fun mapToString(
+        map: Map<String, String>?,
+        transform: (Map.Entry<String, String>) -> String = { (key, value) -> "$SPACESx8$key: $value" }
+    ): String {
         if (map.isNullOrEmpty()) return ""
-        return NEW_LINE + map.map { (key, value) -> "        $key: $value" }
+        return NEW_LINE + map.map(transform)
             .joinToString(System.lineSeparator())
     }
 
-    fun listToString(list: List<String?>?): String {
+    fun listToString(list: List<String?>?, transform: (String) -> String = { "$SPACESx8- $it" }): String {
         if (list.isNullOrEmpty()) return ""
         return NEW_LINE + list.filterNotNull()
-            .joinToString(System.lineSeparator()) { dir -> "        - $dir" }
+            .joinToString(System.lineSeparator(), transform = transform)
     }
 
-    fun objectsToString(objects: List<Any?>?): String {
+    fun objectsToString(objects: List<Any?>?, transform: (Any) -> String = { "$it" }): String {
         if (objects.isNullOrEmpty()) return ""
         return NEW_LINE + objects.filterNotNull()
-            .joinToString(System.lineSeparator()) { "$it" }
+            .joinToString(System.lineSeparator(), transform = transform)
     }
 
     fun listOfListToString(listOfList: List<List<String?>>?): String {
         if (listOfList.isNullOrEmpty()) return ""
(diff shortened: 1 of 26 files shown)


test_runner/src/main/kotlin/ftl/client/google/run/android/GcAndroidTestMatrix.kt
```diff
@@ -73,18 +70,11 @@ private fun createAndroidTestMatrix(
     }.getOrElse { e -> throw FlankGeneralError(e) }
 }
 
-fun TestMatrixAndroid.Config.clientInfo(matrix: TestMatrixAndroid.Type): ClientInfo {
-    return if (matrix is TestMatrixAndroid.Type.Instrumentation && matrix.clientDetails.isNotEmpty()) {
```
Sloox, line 77: @pawelpasterz I am sure that there were changes done here for a fix by @asadsalman. Can we confirm that it wont be effected by this change as it seems like a revert to the previous version of the code.

test_runner/src/main/kotlin/ftl/run/model/AndroidTestContext.kt
```diff
@@ -2,10 +2,15 @@ package ftl.run.model
 
 import ftl.api.FileReference
 import ftl.api.ShardChunks
+import ftl.args.AndroidArgs
 import ftl.args.IgnoredTestCases
 import ftl.shard.Chunk
 
-sealed class AndroidTestContext
+interface WithArgs {
```
Sloox, line 9: Why an interface at this point instead of simply:

```suggestion
sealed class AndroidTestContext(open val args: AndroidArgs)
```

Sloox (commented):
It looks alright for now. Lets get another pair of eyes on it aswell.
@asadsalman Also have a look at the effect it might have for your changes a while back and if it fixes the issues.

asadsalman:
@pawelpasterz this PR looks great! Thanks for adding more fields to additional test APKs!

@Sloox, I had added unit tests that that covered the changes I had made previously. They're still passing so I don't think there's issue or regression here. I'll manually verify with an end-to-end run today too but I don't expect any issues.

test_runner/src/main/kotlin/ftl/client/google/run/android/GcAndroidTestMatrix.kt
pawelpasterz: You are right - this will revert those changes. But this is intentional, previous implementation merges client details (from main and additional test pair). All the other, 'additional' configs overwrite 'main' configs. I think we should be consistent here. Saying that merging is not the best option and `clientDetails` should be overwritten as well.

test_runner/src/main/kotlin/ftl/run/model/AndroidTestContext.kt
pawelpasterz: Hm...why not, I do not have a strong opinion about it 👍🏻

bootstraponline:
>I had added unit tests that that covered the changes I had made previously. They're still passing so I don't think there's issue or regression here. 

I love this! That's one reason why tests are valuable. We're less likely to break business logic when refactoring.

test_runner/src/main/kotlin/ftl/args/ArgsToString.kt
```diff
@@ -3,35 +3,62 @@ package ftl.args
...
+                }
+                pair.devices?.let { list ->
+                    appendLine("$SPACESx8  device:")
+                    list
+                        .map { "  $it" }
+                        .map { it.replace("\n", "\n  ") }
+                        .forEach { append(it) }
+                }
+            }.trimEnd()
+        }
```
piotradamczyk5, line 62: do we have tests for these functions?

pawelpasterz: We do, in `AndroidArgsTest`

Sloox (approved):

piotradamczyk5 (approved):

PR #2010, opened 2021-06-07
fix: Dumpshards fix #1

### LINKED ISSUE #1: per test sharding?
> Instead of creating one shard per test case Flank can create shards that contain tests that add up to a given duration

Is it not possible to have one shard per test case? In terms of fault isolation/video/logs, there are some use cases where this is desirable.

Sloox:
No gaurentee for a fix here as xargs is causing issues and can only be tested once merged :( 

We test.

## Checklist
We may need to get some testing occuring for the different platforms.

### DIFF
--- a/integration_tests/src/test/resources/cases/flank_android_single_legacy.yml
+++ b/integration_tests/src/test/resources/cases/flank_android_single_legacy.yml
@@ -5,4 +5,5 @@ gcloud:
 flank:
   disable-sharding: true
   output-report: json
+  disable-usage-statistics: true
   legacy-junit-result: true
--- a/integration_tests/src/test/resources/cases/flank_ios_single_legacy.yml
+++ b/integration_tests/src/test/resources/cases/flank_ios_single_legacy.yml
@@ -5,3 +5,4 @@ gcloud:
 flank:
   output-report: json
   legacy-junit-result: true
+  disable-usage-statistics: true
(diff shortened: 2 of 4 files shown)


integration_tests/src/test/resources/cases/flank_android_single_legacy.yml
```diff
@@ -5,4 +5,5 @@ gcloud:
 flank:
   disable-sharding: true
   output-report: json
+  disable-usage-statistics: true
```
pawelpasterz, line 8: ❤️

pawelpasterz (commented):
> No gaurentee for a fix here as xargs is causing issues and can only be tested once merged :(

Hm...I might be missing something but your changes are mainly in `test_runner` module so those should be 'testable'

Anyway, I think it would be worth to run IT

test_runner/src/test/kotlin/Debug.kt
```diff
@@ -21,17 +21,17 @@ fun main() {
...
             "run",
+//            "--debug",
 //            "--dry",
 //            "--dump-shards",
-            "--output-style=single",
+//           "--output-style=single",
 //            "--full-junit-result",
 //            "--legacy-junit-result",
-            "-c=./test_runner/src/test/kotlin/ftl/fixtures/test_app_cases/flank-single-success.yml",
+            "-c=./integration_tests/src/test/resources/cases/all_test_filtered_ios.yml",
```
piotradamczyk5, line 34: should we commit this changes

Sloox:
@pawelpasterz I have run it locally and they pass. Will attach some logs if i can shortly.

piotradamczyk5 (approved):

pawelpasterz (approved):

PR #2026, opened 2021-06-10
fix: Fix flank freezes when large number of matrices is launched

### LINKED ISSUE #2021: Large number of matrices causes flank to block indefinitely
**Describe the bug**
If we use flank to launch 70+ matrices in parallel (using `additional-app-test-apks`), flank blocks after uploading the APKs and starting test execution on FTL. The test execution starts and finishes on FTL just fine, but flank does not report their status, or download the artifacts etc.

Flank will upload all the APKs, start the execution, print "Matrices webLink", but does not proceed beyond this point. It does not actually print any of the the matrix links.

This problem does not happen if I have a smaller number of matrices.

It's worth nothing that none of this actually effects test execution itself. The tests start and end on FTL just fine, no unexpected failures or crashes. In fact, if I kill flank after it blocks and then run flank with `refresh` command, flank picks up the matrices, tracks their execution, pulls artifacts etc just fine.

**To Reproduce**

Start a flank invocation with many matrices. 70+ matrices consistently blocked for me. For me it does not always hang, but it hangs most of the time.

**Expected behavior**
I expect the links to the matrices to be printed, then flank to track the status of matrix executions and fetch artifact etc.

> Tested on latest release and latest snapshot.

**Additional context**

I managed to track down the blocking code to [Line 62 of AfterRunTests.kt](https://github.com/Flank/flank/blob/0143dc15ca570482883d01f55d888d595cd115ca/test_runner/src/main/kotlin/ftl/run/platform/common/AfterRunTests.kt#L62). Commenting out line 60-64 in this file actually prevents flank from blocking indefinitely and execution proceeds as normal.

Upon further investigation, I found that block ends up happening inside the [refresh method of GcTestMatrix.kt](https://github.com/Flank/flank/blob/60be7a8187a756eb6c19c3f02d27a706c94e97a6/test_runner/src/main/kotlin/ftl/client/google/GcTestMatrix.kt#L38). The execution never reaches line 42. I had a hunch that flank might be spawning too many coroutines on IO dispatcher. I changed the `Dispatchers.IO` on [line 39](https://github.com/Flank/flank/blob/60be7a8187a756eb6c19c3f02d27a706c94e97a6/test_runner/src/main/kotlin/ftl/client/google/GcTestMatrix.kt#L39) to `Dispatchers.Default`, and everything started working fine.

I'm not sure if doing GET requests on the default dispatcher is the recommended approach, that seems like the perfect job for the IO dispatcher. Maybe, instead of spawning all of the coroutines off at the same time, we can throttle the number of coroutines flank spawns to get matrix links? I'd really appreciate suggestions on how to fix this.

pawelpasterz:
Fixes #2021 

It is impossible to test whole flow in sensible and reliable way. It would require mocking lots of methods and since the problem is related with multithreading I think it's not the best idea. What's more, test might be unreliable due to differences in CPU configurations on different machines.

@asadsalman was asked to verify fix, as a proof (thanks!)

### DIFF
--- a/test_runner/src/main/kotlin/ftl/adapter/TestMatrixRefresh.kt
+++ b/test_runner/src/main/kotlin/ftl/adapter/TestMatrixRefresh.kt
@@ -3,12 +3,10 @@ package ftl.adapter
 import ftl.adapter.google.toApiModel
 import ftl.api.TestMatrix
 import ftl.client.google.refreshMatrix
-import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.coroutineScope
 
-object TestMatrixRefresh :
-    TestMatrix.Refresh,
-    (TestMatrix.Identity) -> TestMatrix.Data by { identity ->
-        runBlocking {
-            refreshMatrix(identity.matrixId, identity.projectId).toApiModel(identity)
-        }
+object TestMatrixRefresh : TestMatrix.Refresh {
+    override suspend fun invoke(identity: TestMatrix.Identity): TestMatrix.Data = coroutineScope {
+        refreshMatrix(identity.matrixId, identity.projectId).toApiModel(identity)
     }
+}
--- a/test_runner/src/main/kotlin/ftl/api/TestMatrix.kt
+++ b/test_runner/src/main/kotlin/ftl/api/TestMatrix.kt
@@ -94,5 +94,7 @@ object TestMatrix {
     )
 
     interface Cancel : (Identity) -> Unit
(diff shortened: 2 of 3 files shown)


asadsalman:
Tried this fix out and it works great! Thank you for the quick help!

pawelpasterz:
@flank-it

adamfilipow92 (approved):

Sloox (approved):
one line? @pawelpasterz

PR #2043, opened 2021-06-23
fix: Enhance table builder logic

### LINKED ISSUE #1893: Flank models list ugly `OS_VERSION_IDS` column format
Flank's `OS_VERSION_IDS` is rather ugly and inconsistent, let's print it in one format:

|  flank | gcloud |
:-------:|:-------:
  |

pawelpasterz:
Fixes #1893 

1. Run `./gradlew flankFullRun`
2. Run `flank firebase test android models list`
3. You should see `OS_VERSION_IDS` column aligned to the left
4. You should see `RESOLUTION` column centered by `x` mark

### DIFF
--- a/test_runner/src/main/kotlin/ftl/environment/TestEnvironmentInfo.kt
+++ b/test_runner/src/main/kotlin/ftl/environment/TestEnvironmentInfo.kt
@@ -1,6 +1,6 @@
 package ftl.environment
 
-import ftl.util.Alignment
+import ftl.util.Align
 import ftl.util.SystemOutColor
 import ftl.util.TableColumn
 
@@ -8,7 +8,7 @@ typealias TestEnvironmentInfo = MutableMap<String, MutableList<String>>
 
 internal fun TestEnvironmentInfo.getOrCreateList(key: String) = getOrPut(key) { mutableListOf() }
 
-internal fun TestEnvironmentInfo.createTableColumnFor(key: String, alignment: Alignment) = TableColumn(key, getValue(key), alignment = alignment)
+internal fun TestEnvironmentInfo.createTableColumnFor(key: String, align: Align) = TableColumn(key, getValue(key), align = align)
 
 internal val tagToSystemOutColorMapper: (String) -> SystemOutColor = {
     when {
--- a/test_runner/src/main/kotlin/ftl/environment/TestEnvironmentInfo.kt
+++ b/test_runner/src/main/kotlin/ftl/environment/TestEnvironmentInfo.kt
@@ -8,7 +8,7 @@ typealias TestEnvironmentInfo = MutableMap<String, MutableList<String>>
 
 internal fun TestEnvironmentInfo.getOrCreateList(key: String) = getOrPut(key) { mutableListOf() }
 
-internal fun TestEnvironmentInfo.createTableColumnFor(key: String, align: Align) = TableColumn(key, getValue(key), align = align)
+internal fun TestEnvironmentInfo.createTableColumnFor(key: String, align: Align = Align.LEFT) = TableColumn(key, getValue(key), align = align)
 
 internal val tagToSystemOutColorMapper: (String) -> SystemOutColor = {
     when {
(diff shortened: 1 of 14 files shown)


jan-gogo (commented):
My proposition is to change `Alignment` to `Align` just because is shorter.

test_runner/src/main/kotlin/ftl/util/LogTableBuilder.kt
```diff
@@ -145,3 +157,11 @@ private fun DataWithSize.center() = String.format(
 )
 
 inline fun TableColumn.applyColorsUsing(mapper: (String) -> SystemOutColor) = copy(dataColor = data.map(mapper))
+
+// RESOLUTION column needs dedicated alignment logic and data pre-processing
+fun TableColumn.alignToTheXMark() = copy(
```
jan-gogo, line 162: This function will look better with the following signature:

```
fun TableColumn.alignToThe(char: Char): TableColumn
```
or

```
fun TableColumn.alignToThe(string: String): TableColumn
```

test_runner/src/main/kotlin/ftl/util/LogTableBuilder.kt
```diff
@@ -145,3 +157,11 @@ private fun DataWithSize.center() = String.format(
 )
 
 inline fun TableColumn.applyColorsUsing(mapper: (String) -> SystemOutColor) = copy(dataColor = data.map(mapper))
+
+// RESOLUTION column needs dedicated alignment logic and data pre-processing
```
Sloox, line 161: Is this a todo or something that has been achieved?

Sloox (approved):
Minor comment, else looks good!

test_runner/src/main/kotlin/ftl/environment/TestEnvironmentInfo.kt
```diff
@@ -1,13 +1,14 @@
...
+import ftl.util.Align
 import ftl.util.SystemOutColor
 import ftl.util.TableColumn
 
 typealias TestEnvironmentInfo = MutableMap<String, MutableList<String>>
 
 internal fun TestEnvironmentInfo.getOrCreateList(key: String) = getOrPut(key) { mutableListOf() }
 
-internal fun TestEnvironmentInfo.createTableColumnFor(key: String) = TableColumn(key, getValue(key))
+internal fun TestEnvironmentInfo.createTableColumnFor(key: String, align: Align) = TableColumn(key, getValue(key), align = align)
```
piotradamczyk5, line 11: could we set `Align.LEFT` as default ?

test_runner/src/main/kotlin/ftl/util/LogTableBuilder.kt
```diff
@@ -12,16 +12,19 @@ data class TableColumn(
     val header: String,
     val data: List<String>,
     val dataColor: List<SystemOutColor> = listOf(),
-    val columnSize: Int = ((data + header).maxByOrNull { it.length }?.length ?: 0) + DEFAULT_COLUMN_PADDING
+    val columnSize: Int = ((data + header).maxByOrNull { it.length }?.length ?: 0) + DEFAULT_COLUMN_PADDING,
+    val align: Align
```
jan-gogo, line 16: As @piotradamczyk5 mentioned, Align.LEFT could set as default, also here.

test_runner/src/main/kotlin/ftl/util/LogTableBuilder.kt
pawelpasterz: It's not a todo, it's just a short info why we have additional method here

test_runner/src/main/kotlin/ftl/util/LogTableBuilder.kt
pawelpasterz: @jan-gogo @piotradamczyk5 
IMO it's clearer to force explicit alignment set up and I do not fully agree with you.
On the other hand I don't have very strong opinion about it and I'm in minority in that case, so sure, I'll change it 👍🏻

piotradamczyk5 (approved):

PR #2045, opened 2021-06-24
feat: Added Flank wrapper

### LINKED ISSUE #2041: Create Flank wrapper
Create MVP for a Flank Evergreen using the wrapper.

1. Create a script that downloads the latest version of Flank(we could use flank-scripts solution for it)
1. Create wrapper for both Windows and Unix
1. Cache flank version

piotradamczyk5:
Fixes #2041

1. run `flankw`
    1. if flank is not present wrapper will download the latest version
    1. if flank is present, but has a different revision or version than the latest one wrapper will download the latest version
    1. if no update is needed flank will not be downloaded
2. then flank will be executed as usual  

## Checklist

- [ ] Documented
- [x] Unix version
- [ ] Windows version

### DIFF
--- a/flank_wrapper/flankw.sh
+++ b/flank_wrapper/flankw.sh
@@ -1,0 +1,29 @@
+download_flank() {
+  echo "Downloading Flank $1..."
+  curl -L https://github.com/Flank/flank/releases/download/$1/flank.jar -o $FLANK_FILE --silent
+}
+
+get_remote_revision() {
+  curl -H "Accept: application/vnd.github.v3+json" https://api.github.com/repos/flank/flank/commits/v21.06.1 -s | grep '"sha":' | head -n 1 | awk '{print $2}' | sed s/\",//g | sed s/\"//g
+}
+
+FLANK_FILE=/usr/local/bin/flank.jar
+REMOTE_VERSION=$(curl -H "Accept: application/vnd.github.v3+json" https://api.github.com/repos/Flank/flank/releases/latest -s | grep "tag_name" | awk '{print $2}' | sed s/\",//g | sed s/\"//g)
+
+if [ ! -f "$FLANK_FILE" ]; then
+  download_flank $REMOTE_VERSION
+else
+  /usr/bin/env java -jar $FLANK_FILE --version > flank-version
+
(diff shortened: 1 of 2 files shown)


bootstraponline (changes requested):
I think we'll want to adopt a self contained binary approach, as mentioned on slack. Kotlin native or GraalVM can get us a binary from Kotlin.

- Dependency on curl. What happens if cURL is not available? What happens if cURL errors? 
- How do we test shell script code?
- We should add sha256 hash checking per https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:verification This will involve updating the Flank publish process to generate a .sha256. We do this already on the dashboard. https://github.com/Flank/flank-dashboard/releases/tag/metrics-cli-snapshot
- Flank supports windows, mac, and linux. I'm not sure we want to ship a bat file on Windows. Testing the bat logic automatically would be painful.
- GitHub API can get rate limited without providing a credential. As the tool will be running on a CI machine in the cloud, the IPs are likely already at the rate limit. We might be able to get around this by relying on https://github.com/Flank/flank/releases/latest to redirect to the latest version page, then directly discover the download URLs from the result.

I think for the evergreen epic, let's start with a design doc and discuss how it'll meet the requirements.

flank_wrapper/flankw.sh
```diff
@@ -0,0 +1,29 @@
+download_flank() {
+  echo "Downloading Flank $1..."
+  curl -L https://github.com/Flank/flank/releases/download/$1/flank.jar -o $FLANK_FILE --silent
+}
+
+get_remote_revision() {
+  curl -H "Accept: application/vnd.github.v3+json" https://api.github.com/repos/flank/flank/commits/v21.06.1 -s | grep '"sha":' | head -n 1 | awk '{print $2}' | sed s/\",//g | sed s/\"//g
```
adamfilipow92, line 7: I think here we should use $REMOTE_VERSION instead of hardcoded v21.06.1 wdyt?

piotradamczyk5: As we agreed with standup we will deliver Kotlin based version

piotradamczyk5: but any way this is huge mistake from my side :D thanks

PR #2046, opened 2021-06-24
feat: Parameterized tests options

### LINKED ISSUE #2023: Parameterized Tests: Add new options #1

Parameterized tests need options to help in sorting out the issues they might represent. 

**Describe the solution**
Three options should be added within the scope of this ticket namely
* `ignore-all` - All parameterized tests should be ignored and not sharded
* `collect-all-single` -  All parameterized tests should be collected and sharded in a single separate shard.
* `default` - default implementation

This should be applied only to android and an example of the API:

```
### Parameterized Tests
  ## Specifies how to handle tests which contain Parameterization.
  ## 4 options are available
  ## default: treat Parameterized tests as normal and shard accordingly
  ## ignore-all: Parameterized tests are ignored and not sharded
  ## shard-into-single: Parameterized tests are collected and put into a single shard
  ## shard-smart: Smart collection of all the parameterized test and run separately on a per case basis
  ## Note: if left blank default is used.
  # parameterized-tests: default
```

Take note this will require tests & integration tests. Therefore an example parameterized test application needs to be built to accommodate this and test it thoroughly.

More information about the discussion is found here: https://github.com/Flank/flank/issues/1720

Sloox:
Fixes #2023 

## Test Plan

Need to add parameterized test artifacts to flank.

## Checklist
- [x] Documented
- [x] Unit tested


Sloox:
@flank-it

Sloox:
@flank-it

test_runner/src/main/kotlin/ftl/run/model/AndroidTestContext.kt
```diff
@@ -15,6 +15,7 @@ data class InstrumentationTestContext(
     val ignoredTestCases: IgnoredTestCases = emptyList(),
     val environmentVariables: Map<String, String> = emptyMap(),
     val testTargetsForShard: ShardChunks = emptyList(),
+    val parameterizedTestsOption: String = "",
```
pawelpasterz, line 18: `parameterizedTests` can be acquired from `args`, I think we do not need to have id duplicated here.

test_runner/src/main/kotlin/ftl/run/platform/android/CreateAndroidTestContext.kt
```diff
@@ -75,11 +76,14 @@ private fun InstrumentationTestContext.downloadApks(): InstrumentationTestContex
...
 
-private fun InstrumentationTestContext.calculateShards(
+private fun InstrumentationTestContext.calculateShardsNormally(
     args: AndroidArgs,
     testFilter: TestFilter = TestFilters.fromTestTargets(args.testTargets, args.testTargetsForShard)
 ): InstrumentationTestContext = ArgsHelper.calculateShards(
-    filteredTests = getFlankTestMethods(testFilter),
+    filteredTests = getFlankTestMethods(
+        testFilter = testFilter,
+        parameterizedTests = args.parameterizedTests ?: FlankDefaults.DEFAULT_PARAMETERIZED_TESTS
```
pawelpasterz, line 85: `AndroidGcloudConfig` has a default value for `parameterizedTests` therefor I think `parameterizedTests` should not be nullable (in `AndroidArgs`) To make it happen you will also need to make small change in `CreateAndroidArgs`:
`parameterizedTests = gcloud.parameterizedTests` -> `parameterizedTests = gcloud::parameterizedTests.require()`

With the above changes, we can get rid of 'elvis defaults' calls in other places. I think it's worth changing, we will have only one place to update if we decide to change a default value.
Let me know what you think

test_runner/src/main/kotlin/ftl/run/platform/android/CreateAndroidTestContext.kt
```diff
@@ -89,11 +93,42 @@ private fun InstrumentationTestContext.calculateShards(
...
+private fun String.shouldShardIntoSingle() = (this == "shard-into-single")
+
 private fun InstrumentationTestContext.calculateDummyShards(
     args: AndroidArgs,
-    testFilter: TestFilter = TestFilters.fromTestTargets(args.testTargets, args.testTargetsForShard),
+    testFilter: TestFilter = TestFilters.fromTestTargets(args.testTargets, args.testTargetsForShard)
 ): InstrumentationTestContext {
-    val filteredTests = getFlankTestMethods(testFilter)
+    val filteredTests =
+        getFlankTestMethods(testFilter, args.parameterizedTests ?: FlankDefaults.DEFAULT_PARAMETERIZED_TESTS)
```
pawelpasterz, line 131: related to the previous comment

test_runner/src/main/kotlin/ftl/run/platform/android/CreateAndroidTestContext.kt
```diff
@@ -119,13 +155,16 @@ internal fun InstrumentationTestContext.getFlankTestMethods(
             .filter(testFilter.shouldRun)
             .filterNot(parameterizedClasses::belong)
             .map(TestMethod::toFlankTestMethod).toList()
-            .plus(parameterizedClasses.onlyShouldRun(testFilter))
+            .plus(parameterizedClasses.onlyShouldRun(testFilter, parameterizedTests.shouldIgnore()))
```
pawelpasterz, line 158: `#shouldIgnore` is a simple method, I think we can return an empty list here instead of inside `onlyShouldRun` method
```suggestion
            .plus(if (parameterizedTests.shouldIgnore()) emptyList() else parameterizedClasses.onlyShouldRun(testFilter))
```

Not saying it have to look like this, but I'm sure you got the idea.
WDYT?

test_runner/src/main/kotlin/ftl/run/platform/android/CreateAndroidTestContext.kt
```diff
@@ -110,7 +145,8 @@ private fun InstrumentationTestContext.calculateDummyShards(
 
 @VisibleForTesting
 internal fun InstrumentationTestContext.getFlankTestMethods(
-    testFilter: TestFilter
+    testFilter: TestFilter,
+    parameterizedTests: String = FlankDefaults.DEFAULT_PARAMETERIZED_TESTS
```
pawelpasterz, line 149: I think the default value here is redundant. I could not find a method call without `parameterizedTests`

test_runner/src/main/kotlin/ftl/run/platform/android/ResolveApks.kt
```diff
@@ -25,7 +25,8 @@ private fun AndroidArgs.mainApkContext() = appApk?.let { appApk ->
             test = testApk.asFileReference(),
             environmentVariables = emptyMap(),
             testTargetsForShard = testTargetsForShard,
-            args = this
+            args = this,
+            parameterizedTestsOption = parameterizedTests.orEmpty()
```
pawelpasterz, line 29: Related to comment

test_runner/src/main/kotlin/ftl/run/platform/android/ResolveApks.kt
```diff
@@ -49,6 +50,7 @@ private fun AndroidArgs.additionalApksContexts() = additionalAppTestApks.map {
                 clientDetails = it.clientDetails ?: clientDetails
             ),
             testTargets = it.testTargets ?: testTargets
-        )
+        ),
+        parameterizedTestsOption = parameterizedTests.orEmpty()
```
pawelpasterz, line 54: Related to comment

pawelpasterz (commented):
It would be also worth to add possible values of `--parameterized-test` option in description (`AndroidGcloudConfig`)

pawelpasterz:
And I think we could also add this option to `additional-app-test-apks`

test_runner/src/main/kotlin/ftl/run/model/AndroidTestContext.kt
Sloox: Will check. It might be left over here.

test_runner/src/main/kotlin/ftl/run/platform/android/CreateAndroidTestContext.kt
Sloox: good spot thanks for the advice and its been implemented :)

test_runner/src/main/kotlin/ftl/run/platform/android/CreateAndroidTestContext.kt
Sloox: 👍

test_runner/src/main/kotlin/ftl/run/platform/android/CreateAndroidTestContext.kt
Sloox: Looking at simply neatness i think having an if statement in a chained set of function calls looks worse than separating it out like is being done for all the other chained function calls.

test_runner/src/main/kotlin/ftl/run/platform/android/CreateAndroidTestContext.kt
Sloox: 👍

test_runner/src/main/kotlin/ftl/run/platform/android/ResolveApks.kt
Sloox: 👍

test_runner/src/main/kotlin/ftl/run/platform/android/ResolveApks.kt
Sloox: 👍

test_runner/src/main/kotlin/ftl/run/platform/android/CreateAndroidTestContext.kt
Sloox: This will stay as there are multiple tests that rely upon it.

pawelpasterz: If it must, let it be. It's just a bit smelly ( 😿  look, you made a kitty sad)

test_runner/src/main/kotlin/ftl/run/platform/android/CreateAndroidTestContext.kt
pawelpasterz: Well, it doesn't have to be a chained call. I'm saying that we can avoid passing a boolean parameter that will be used only to decide if we should return an empty list or not. That's all :)

pawelpasterz (approved):

pawelpasterz:
@flank-it

adamfilipow92 (approved):

PR #2050, opened 2021-06-28
fix: Corellium client request retry 

jan-gogo:
The PR is fixing incorrect behaviour of the retry method that was always looping 5 times.

Additionally adds `--dirty` option to `flankScripts assemble flank` this will speed up builds for local testing purpose.


pawelpasterz:
> Additionally adds --dirty option to flankScripts assemble flank this will speed up builds for local testing purpose.

Honestly, I thought no one uses this locally 🙌🏻

pawelpasterz (approved):

Sloox (approved):

PR #2062, opened 2021-07-01
feat: Merge test method duration for parameterized classes

### LINKED ISSUE #2052: Parameterized tests support
## Motivation

Flank-Corellium doesn't support various types of parameterized tests.
So, the following features may not works or works partially:

*  Standard parameterized tests annotated with `@RunWith(Parameterized::class)` can be shared and executed but test cases duration from previous runs for smart sharding cannot be resolved properly due to modified parameterized test cases names in test results.
* Test cases from classes annotated with `@RunWith(JUnitParamsRunner::class)` are required to be run as a whole class without splitting on to single test cases which currently is not supported.

## Goal

Have implemented full support for currently known types of parameterized tests

## Design

* Tests from classes annotated `@RunWith(Parameterized::class)` :
    * Are running the same as normal test cases.
    * Calculating test cases duration requires a merge of parameterized runs because one parameterized test case typically is producing many executions for different parameters. The parameterized test case duration for sharding is a summary of this test case runs with different parameters.
* Tests from classes annotated `@RunWith(JUnitParamsRunner::class)`:
    * Cannot be split and must be run as a whole class.
    * Needs to be recognized by class instead of method.
    * Methods of this type need to be filtered out before sharding, only classes must be taken into account.

## Dependencies

Affected modules:

* `:tool:apks` - Expand `Apk.ParseTestCases` implementation for parsing classes annotated by `@RunWith(JUnitParamsRunner::class)` as test cases.
* `:tool:junit` - Add additional function for merging parameterized test cases with durations.
* `:corellium:domain` - Expand `RunTestCorelliumAndroid.Context.loadPreviousDurations` task for merging parameterized tests duration before running `List<JUnit.TestResult>.calculateTestCaseDurations`

jan-gogo:
Fixes #2052 

## Changes

* Adds to module `:tool:junit` function for merging test methods into test-class with accumulated duration.
* Merges test methods for given class names while loading previous durations.

Build flank:
```shell
. .env
flankScripts assemble flank -d
```
Run the following command twice:
```shell
flank corellium test android run -c="./test_configs/flank-corellium.yml"
```
The second run should calculate durations for parameterized classes.
Check generated `android-shards.json` each parameterized class should have a duration different than the default (120).

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/tool/junit/src/main/kotlin/flank/junit/JUnit.kt
+++ b/tool/junit/src/main/kotlin/flank/junit/JUnit.kt
@@ -38,7 +38,7 @@ fun JUnit.Report.writeAsXml(writer: Writer) {
 }
 
 /**
- * Merge [JUnit.TestResult] test methods by class names to accumulate duration and return them as classes of test methods.
+ * Merge [JUnit.TestResult] test methods by class names to accumulate duration and return them as classes.
  */
 fun List<JUnit.TestResult>.mergeTestCases(byClasses: Set<String>): List<JUnit.TestResult> =
     mergeDurations(byClasses)
--- a/tool/junit/src/main/kotlin/flank/junit/JUnit.kt
+++ b/tool/junit/src/main/kotlin/flank/junit/JUnit.kt
@@ -5,6 +5,7 @@ import com.fasterxml.jackson.databind.annotation.JsonSerialize
 import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlProperty
 import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlRootElement
 import flank.junit.internal.calculateMedianDurations
+import flank.junit.internal.mergeDurations
 import flank.junit.internal.parseJUnitTestResults
 import flank.junit.mapper.TimeSerializer
 import flank.junit.mapper.mapToTestSuites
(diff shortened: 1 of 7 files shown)


Sloox (changes requested):
@jan-gogo can this wait for https://github.com/Flank/flank/pull/2046

tool/junit/src/main/kotlin/flank/junit/mapper/Structural.kt
```diff
@@ -31,13 +31,17 @@ internal fun List<JUnit.TestResult>.mapToTestSuites(): List<JUnit.Suite> = this
...
         suite.testcases.map { case ->
+            startAt = endAt
+            endAt = startAt + (case.time * 1000).toLong()
             JUnit.TestResult(
                 suiteName = suite.name,
                 testName = case.name,
                 className = case.classname,
-                startAt = 0,
-                endsAt = (case.time * 1000).toLong(),
+                startAt = startAt.also { startAt = endAt },
```
Sloox, line 43: wont this simply result in 
endAt = endAt
& 
startAt = endAt?

Sloox (approved):
Seems alright. Minor comment :)

tool/junit/src/main/kotlin/flank/junit/mapper/Structural.kt
jan-gogo: I don't understand your suggestion, but it's reminded me that I had forgotten to add a proper test. So, I added a small update. If you see a shorter solution, feel free to check out the branch and test it using `StructuralKtTest.mapToTestResultsTest`

jan-gogo: Right, this part `.also { startAt = endAt }` was not necessary

tool/junit/src/main/kotlin/flank/junit/JUnit.kt
```diff
@@ -36,6 +37,12 @@ fun JUnit.Report.writeAsXml(writer: Writer) {
     xmlPrettyWriter.writeValue(writer, this)
 }
 
+/**
+ * Merge [JUnit.TestResult] test methods by class names to accumulate duration and return them as classes of test methods.
```
jan-gogo, line 41:
```suggestion
 * Merge [JUnit.TestResult] test methods by class names to accumulate duration and return them as classes.
```

zuziakaxel-shape (approved):

axelzuziak-gogo (approved):

Issue #2044, opened 2021-06-24
Flank metrics for corellium


adamfilipow92:
Currently, we collecting data only from FIrebase run. To be able to use business metrics from corellium run we need to implement the following events:

- configuration
- total cost
- cost by device type
- test duration
- flank version
- bundle id / package name

All events are described in this doc [flank_mixpanel_metrics.md](https://github.com/Flank/flank/blob/cabab70653271861b4f10fcb74f4ab33b1a31eae/docs/flank_mixpanel_metrics.md)

axelzuziak-gogo:
Still blocked by Corellium API #2133

PR #2086, opened 2021-07-12
feat: Move Mixpanel metrics to tools:analytics

adamfilipow92:
Fixes #

This is part of #2044. Everything should work as before. Tests should pass.


Sloox (approved):
Looks good @adamfilipow92

test_runner/src/main/kotlin/ftl/reports/CostReport.kt
```diff
@@ -28,21 +24,6 @@ object CostReport : IReport {
             totalBillablePhysicalMinutes += it.billableMinutes.physical
         }
 
-        val virtualCost = calculateVirtualCost(totalBillableVirtualMinutes.toBigDecimal())
```
adamfilipow92, line 31: Removed because currently causing failed tests. We will merge all events into single https://github.com/Flank/flank/issues/2087 so this event is currently redundant

piotradamczyk5 (approved):

PR #2094, opened 2021-07-20
feat: Added Sdk Suppress Handling

### LINKED ISSUE #2053: Flank does not handle `@SdkSuppress` annotation correctly
**Describe the bug**

#### copy-pasted from the original issue https://github.com/Flank/flank/issues/818#issuecomment-870766196

We're seeing this issue as well, usually because we have `@SdkSuppress` on on a test which happens to be the only test on the shard.

Although in our case, FTL doesn't detect that there were no tests in the shard - instead, it reports a single test with the name `null` as seen here:

image
image

```
java.lang.ClassNotFoundException: Invalid name: no tests found
at java.lang.Class.classForName(Native Method)
at java.lang.Class.forName(Class.java:400)
at androidx.test.internal.runner.TestLoader.doCreateRunner(TestLoader.java:72)
at androidx.test.internal.runner.TestLoader.getRunnersFor(TestLoader.java:105)
at androidx.test.internal.runner.TestRequestBuilder.build(TestRequestBuilder.java:804)
at androidx.test.runner.AndroidJUnitRunner.buildRequest(AndroidJUnitRunner.java:575)
at androidx.test.runner.AndroidJUnitRunner.onStart(AndroidJUnitRunner.java:393)
at android.app.Instrumentation$InstrumentationThread.run(Instrumentation.java:1932)
```

We also don't see the message that you mentioned in the output, "One or more shards failed because there are no test cases inside." The table in the output just says that "N tests failed".

@bootstraponline Would this be considered the same issue, or should I file a different one? Or is this something that needs to be looked into on the FTL side? (Or even ATO?) Is this something that can be solved in Flank?

**Expected behavior**

Flank should skip tests when sdk version requirement is not fulfilled
Once fixed -- update IT with new case

piotradamczyk5:
Fixes #2053

1. Run Flank with `SdkSuppress` methods (ex. use `test_projects/android/app_sdksuppress`) project
2. Play with different API versions(using YAML `device`->`version` and check if flank has any problems)

## Checklist

- [x] Unit tested
- [x] Tests project added


test_runner/src/main/kotlin/ftl/run/platform/android/CreateAndroidTestContext.kt
```diff
@@ -99,11 +102,13 @@ private fun InstrumentationTestContext.calculateShards(
 ): InstrumentationTestContext = if (args.parameterizedTests.shouldShardSmartly()) {
     var flankTestMethods = getFlankTestMethods(testFilter, args.parameterizedTests)
     val parameterizedTests = flankTestMethods.filter { it.isParameterizedClass }
-    flankTestMethods = flankTestMethods.filterNot { it.isParameterizedClass }
+    val sdkSuppressTests = flankTestMethods.filter { it.sdkSuppressLevels != null }
+    flankTestMethods = flankTestMethods.filterNot { it.isParameterizedClass }.filterNot { it.sdkSuppressLevels != null }
```
Sloox, line 106: can this not be :
```suggestion
    flankTestMethods = flankTestMethods.filterNot { it.isParameterizedClass || it.sdkSuppressLevels != null }
```

test_runner/src/main/kotlin/ftl/run/platform/android/CreateAndroidTestContext.kt
```diff
@@ -99,11 +102,13 @@ private fun InstrumentationTestContext.calculateShards(
...
     val shards = calculateParameterizedShards(flankTestMethods, args)
     val shardCount = if (args.parameterizedTests.isSingleParameterizedShard()) 1 else parameterizedTests.size
     val parameterizedShard = calculateParameterizedShards(parameterizedTests, args, shardCount)
-    shards.copy(shards = shards.shards + parameterizedShard.shards)
+    val sdkSuppressShard = calculateParameterizedShards(sdkSuppressTests, args, shardCount)
+    shards.copy(shards = shards.shards + parameterizedShard.shards + sdkSuppressShard.shards)
 } else calculateShardsNormally(args, testFilter)
```
Sloox, line 112: I have a feeling this logic should also be applied within the _calculateShardsNormally_ methodology.
For now it seems like it only works when we are sharding.

piotradamczyk5: I will investigate that case and let you know about the results

piotradamczyk5: @Sloox I have investigated this problem and looks that it is resolved by filtering tests during creating `FlankTestMethod`. I could not reproduce the reported problem, but I think that provided solution should fix the issue.

Sloox: Alright. Ill approve it then :)

Sloox (approved):

adamfilipow92 (approved):

PR #2124, opened 2021-08-06
fix: iOS nm global tests discovery

### LINKED ISSUE #2070: Flank can't find swift test methods from `final` test classes
**Describe the bug**

If you for some reason label your test class as `final`, flank can't find it's test methods as it uses `nm -gU` to list symbols, which would only list global defined symbols.

**To Reproduce**

1. Define a swift test class and mark it as `final`.
2. Build test products and run `flank ios run`
3. See that swift test cases are not found by flank.

**Expected behavior**

>

**Additional context**

I think removing `-g` option should be enough to get non-global methods too. This is already what has been used for ObjC test methods.

Sloox:
Fixes #2070

Needs a round of testing

## Checklist

- [x] Documented
- [x] Unit tested


Sloox:
@flank-it

zuziakaxel (approved):

piotradamczyk5 (approved):

PR #2127, opened 2021-08-09
refactor: Bash command execution

### LINKED ISSUE #1770: Refactor Bash/Shell Environment

Due to the inclusion of Windows, Linux and MacOs and the possible iterations in the future (M1 mac for example) a proper shell/bash environment needs to be created and handled.

**Describe the solution**

Specifically it can be expanded as follows:
1. Handle all the iterations of shell environments for each OS (Bash, BinBash, CMD, Powershell etc)
2. A way to store and handle external commands to be run through the Shell above (NM, Swift demangle etc)
3. Make sure its easy to extend when needed and compatible with the Os we have identified.
4. Follow good principles as set out by the flank

piotradamczyk5:
Fixes #1770

Code is refactored according to #1770. All commands works as previously


Sloox:
@flank-it

pawelpasterz (approved):

Sloox (approved):
Nice 👍

PR #2131, opened 2021-08-12
feat: Implement env parsing

### LINKED ISSUE #1882: set test-targets for test-run based on env vars
I want to set the test-targets for my ios test-run based on env vars. That would allow me for example to run only the tests for features which got changed in a PR as a statuscheck

*example config* 
`...`
`test-targets:`
`  - $TESTTARGETS`
`...`

pawelpasterz:
Fixes #1882

2. export env variable with test targets, i.e. 
    ```
    export TEST_TARGETS="class your.test.Clazz, notAnnotation your.any.Annotation"
    ```
3. update your config with env:
    ```
    test-targets:
      - $TEST_TARGETS
    ```
4. start flank test run and see test targets being applied

## Checklist

- [x] Documented
- [x] Unit tested

### DIFF
--- a/test_runner/src/main/kotlin/ftl/util/Utils.kt
+++ b/test_runner/src/main/kotlin/ftl/util/Utils.kt
@@ -128,14 +128,3 @@ fun <T> KMutableProperty<T?>.require() =
 fun getGACPathOrEmpty(): String = System.getenv("GOOGLE_APPLICATION_CREDENTIALS").orEmpty()
 
 fun saveToFlankLinks(vararg links: String) = File("flank-links.log").writeText(links.joinToString(System.lineSeparator()))
-
-fun getEnv(name: String): String? = System.getenv(name)
-
-fun List<String>.parseEnvsIfNeeded() = this
-    .partition { it.startsWith("$") }
-    .let { (envs, commonTestTargets) ->
-        commonTestTargets + envs
-            .mapNotNull { getEnv(it.drop(1)) }
-            .flatMap { it.split(",") }
-            .map { it.trim() }
-    }
--- a/test_runner/src/main/kotlin/ftl/util/Utils.kt
+++ b/test_runner/src/main/kotlin/ftl/util/Utils.kt
@@ -128,3 +128,14 @@ fun <T> KMutableProperty<T?>.require() =
 fun getGACPathOrEmpty(): String = System.getenv("GOOGLE_APPLICATION_CREDENTIALS").orEmpty()
 
 fun saveToFlankLinks(vararg links: String) = File("flank-links.log").writeText(links.joinToString(System.lineSeparator()))
+
+fun getEnv(name: String): String? = System.getenv(name)
+
+fun List<String>.parseEnvsIfNeeded() = this
+    .partition { it.startsWith("$") }
+    .let { (envs, commonTestTargets) ->
+        commonTestTargets + envs
+            .mapNotNull { getEnv(it.drop(1)) }
+            .flatMap { it.split(",") }
+            .map { it.trim() }
+    }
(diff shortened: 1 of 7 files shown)


pawelpasterz:
@flank-it

test_runner/src/main/kotlin/ftl/util/Utils.kt
```diff
@@ -128,3 +128,14 @@ fun <T> KMutableProperty<T?>.require() =
...
+fun getEnv(name: String): String? = System.getenv(name)
+
+fun List<String>.parseEnvsIfNeeded() = this
+    .partition { it.startsWith("$") }
+    .let { (envs, commonTestTargets) ->
+        commonTestTargets + envs
+            .mapNotNull { getEnv(it.drop(1)) }
+            .flatMap { it.split(",") }
+            .map { it.trim() }
+    }
```
piotradamczyk5, line 141: I guess it could be
```kotlin
fun List<String>.parseEnvsIfNeeded() = partition { it.startsWith("$") }
    .let { (envs, commonTestTargets) -> commonTestTargets + envs }
    .mapNotNull { getEnv(it.drop(1)) }
    .flatMap { it.split(",") }
    .map { it.trim() }
```

and then we avoid nested streams

piotradamczyk5: this function is used only in `test_runner.src.main.kotlin.ftl.args.CreateAndroidArgs|CreateIosArgs` so it would be better to put it somewhere there instead of publishing project-wide util

test_runner/src/test/kotlin/ftl/fixtures/flank.local.yml
```diff
@@ -10,6 +10,7 @@ gcloud:
   async: true
   test-targets:
   - class com.example.app.ExampleUiTest#testPasses
+  - $FROM_ENV
```
piotradamczyk5, line 13: maybe it should be placed in a separate file to avoid having empty test-targets for anybody who does not set this `FROM_ENV` env?

test_runner/src/main/kotlin/ftl/util/Utils.kt
pawelpasterz: Totally agree, thanks!

test_runner/src/test/kotlin/ftl/fixtures/flank.local.yml
pawelpasterz: sure, why not 👍🏻

test_runner/src/main/kotlin/ftl/util/Utils.kt
pawelpasterz: > I guess it could be
> 
> ```kotlin
> fun List<String>.parseEnvsIfNeeded() = partition { it.startsWith("$") }
>     .let { (envs, commonTestTargets) -> commonTestTargets + envs }
>     .mapNotNull { getEnv(it.drop(1)) }
>     .flatMap { it.split(",") }
>     .map { it.trim() }
> ```
> 
> and then we avoid nested streams

I think that won't work, I need to parse strings with `$` differently. The proposed solution will spread the list into two and then combine them into one and perform mapping on the combined one.

But, as I wrote, I agree with your suggestion regarding package

test_runner/src/test/kotlin/ftl/fixtures/flank.local.yml
pawelpasterz: On second thought, I think it should stay as it is now. Yeah, in most cases it will produce nothing (we will have only `ExampleUiTest#testPasses`) but that would also verify if flank can skip empty variables without any problem. 
Let me know what do you think

piotradamczyk5 (approved):

Sloox (approved):

PR #2154, opened 2021-08-25
feat: Increase timeout for http request

### LINKED ISSUE #2148: Increase timeout on getApkDetails
- [context](https://firebase-community.slack.com/archives/C72V6UW8M/p1629456282056100)

The `getApkDetails` endpoint often fails with 503 errors due to timing out.

```
  com.google.api.client.googleapis.json.GoogleJsonResponseException: 503 Service Unavailable
POST https://testing.googleapis.com/v1/applicationDetailService/getApkDetails
{
  "code": 503,
  "errors": [
    {
      "domain": "global",
      "message": "The service is currently unavailable.",
      "reason": "backendError"
    }
  ],
  "message": "The service is currently unavailable.",
  "status": "UNAVAILABLE"
}
```

We should add the `X-Server-Timeout`  HTTP header to increase the timeout

https://cloud.google.com/apis/docs/system-parameters

pawelpasterz:
Fixes #2148 

Build and IT finish successfully


pawelpasterz:
@flank-it

test_runner/src/main/kotlin/ftl/http/HttpTimeoutIncrease.kt
```diff
@@ -9,6 +9,7 @@ class HttpTimeoutIncrease(credentials: Credentials) : HttpCredentialsAdapter(cre
     override fun initialize(request: HttpRequest?) {
         super.initialize(request)
 
+        request?.headers?.put("X-Server-Timeout", 60)
```
Sloox, line 12: Will this apply to all network calls?

Sloox: Should we be worried about it, if it does?

bootstraponline: > Timeout (in seconds, float value) for the server to finish processing the request. This system param only applies to REST APIs for which client-side timeout is not applicable.

60 second timeout seems incredibly short? Google cloud is quite slow depending on what endpoint we're using.

pawelpasterz: @bootstraponline What value do you think, would be reasonable?

PR #2272, opened 2022-08-09
feat: Output cost report as JSON

mathroule:
Fixes #2197

`CostReport.txt` is unstructured, making it difficult to parse it. Add a new report `CostReport.json` with this structure:
```
{
  "currency": "USD",
  "cost": {
    "virtual": 0.07,
    "physical": 0.25,
    "total": 0.32
  },
  "billable-minutes": {
    "virtual": 4,
    "physical": 3,
    "total": 7
  }
}
```

A new report named `CostReport.json` should be generated.

## Checklist

- [x] Documented
- [x] Unit tested
- [x] Integration tests updated

PR #2273, opened 2022-08-11
feat: Add billable minutes to JSON output report

### LINKED ISSUE #2197: CostReport as a structurized file like JSON, YAML, etc.

As a developer, I want to have CostReport.txt file as a structurized file like JSON or YAML so I can parse it easily to get values that I need.

**Describe the solution**
Create for ex. JSON file with all the information:

```
{
  "total": 7.7,
  "currency": "USD",
  "tests": [
    {
      "type": "physical",
      "duration": 15,
      "cost": 7.3
    },
    {
      "type": "virtual",
      "duration": 5,
      "cost": 0.4
    }
  ]
}
```

Separate files to easily parse content. For ex.:
* file `virtual_cost.txt` with content "$0.3"
* file `virtual_duration.txt` with content "10"
* file `physical_cost.txt` with content "$1.1"
* file `physical_duration.txt` with content "15"

mathroule:
Fixes #2197
Replaces #2272

The JSON output report `outputReport.json` will contain a new entry `billable-minutes`:
```
  "cost": {
    ...
  },
  "billable-minutes": {
    "physical": 0,
    "virtual": 1,
    "total": 1
  },
  "test_results": {
    ...
  },
```

## Checklist

- [ ] Documented
- [x] Unit tested


mathroule:
The build is failing because `GRADLE_ENTERPRISE_ACCESS_KEY` is missing.
@bootstraponline do you know how to fix that? Thanks

common/src/main/kotlin/flank/common/OutputReportKeys.kt
```diff
@@ -4,4 +4,5 @@ const val OUTPUT_ARGS = "args"
 const val OUTPUT_WEBLINKS = "weblinks"
 const val OUTPUT_TEST_RESULTS = "test_results"
 const val OUTPUT_COST = "cost"
+const val OUTPUT_BILLABLE_MINUTES = "billable-minutes"
```
adamvduke, line 7: It seems a little unintuitive to me to have keys within the same structure mixing hyphens and underscores. Do you have a strong preference for using a key with a hyphen rather than underscore?

test_runner/src/test/kotlin/ftl/reports/output/OutputReportLoggersTest.kt
```diff
@@ -74,9 +74,23 @@ class OutputReportLoggersTest {
...
+        // given
+        val physical: BigDecimal = BigDecimal.ONE
+        val virtual: BigDecimal = BigDecimal.TEN
+        val total: BigDecimal = BigDecimal.ZERO
+
+        // when
+        outputReport.logBillableMinutes(physical, virtual, total)
+
+        // then
+        assertThat(outputReport.outputData).containsKey("billable-minutes")
```
adamvduke, line 94: Is it worth asserting anything about the keys/values of the new billable minutes structure? I see that the test above does a similar check only looking for the top level key. It seems like the intention is to feed this output to another program/process, so defining the entire contract here in the test would help ensure it can't be broken in the future.

common/src/main/kotlin/flank/common/OutputReportKeys.kt
mathroule: Agreed, mixing both is not ideal. Actually, it's already mixed in the code like in https://github.com/Flank/flank/blob/6c6eb5377e6b748a39203719809f4ade0660f11a/integration_tests/src/test/kotlin/utils/OutputReportParser.kt#L20-L30

 I have no strong preference too, but I can change it to use hypen.

integration_tests/src/test/kotlin/utils/OutputReportParser.kt
```diff
@@ -16,6 +17,8 @@ val OutputReport.firstTestSuiteOverview: SuiteOverview
 data class OutputReport(
     val args: Any,
     val cost: OutputReportCostNode? = null,
+    @JsonProperty("billable-minutes")
```
adamvduke, line 20: Does this property definition need to match the `OUTPUT_BILLABLE_MINUTES` value? I see an integration test failure that looks like:
```
integration.MultipleDevicesIT > run tests on multiple devices - android FAILED
    com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "billable_minutes" (class utils.OutputReport), not marked as ignorable (6 known properties: "billable-minutes", "weblinks", "error", "test_results", "args", "cost"])
```

adamvduke:
@flank-it

adamvduke:
Sorry for the delay and back and forth on this. Thanks for taking the time to contribute!

PR #2322, opened 2022-11-05
chore: Migrate set-output GitHub Actions command

mathroule:
GitHub Actions changed the way `set-output` command works. See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands.


adamvduke (approved):

adamvduke:
Thanks for doing this!

PR #2366, opened 2023-04-27
fix: bringing integration tests back to life

Fry-kun:
* Updated default iOS test device from iphone8/12 to iphone13pro/15.7
  FTL recently deprecated iphone8/12, so integration tests have been
  failing
* Updated local copy of FTL catalog to recent version
* Replaced iOS integration test app EarlGreyExample with FlankExampleTest EarlGreyExample is failing on newer OS/xcode. Current copy is based on EarlGreyExample v1, which is no longer maintained. V2 is pretty hard to get up and running, so switching to FlankExampleTest is the easiest way to unblock integration tests.


adamvduke:
@flank-it

test_projects/ios/EarlGreyExample/flank.ios.yml
```diff
@@ -3,7 +3,7 @@ gcloud:
   test: ./earlgrey_example.zip
   xctestrun-file: ./dd_tmp/Build/Products/EarlGreyExampleSwiftTests_iphoneos13.4-arm64e.xctestrun
   device:
-  - model: iphone8
-    version: 12.0
+  - model: iphone13pro
+    version: 15.7
```
kaibolay, line 7: Does this work for this `EarlGreyExample`?
If not, is this example needed for anything else or should it be deleted entirely?

adamvduke (approved):

Fry-kun:
Fixes #2356

PR #2360, opened 2023-04-06
fix: rate limiting issue when downloading flank with no version supplied

rfogar2:
Fixes #2344

Tests should work
Gradle command should work

## Checklist

- [ ] Documented
- [ ] Unit tested
- [ ] Integration tests updated

PR #2369, opened 2023-05-02
fix: Remove usage of secrets to fix failing builds

rfogar2:
Fixes #

Illegal use of `secrets`, see comment in this PR for more context

https://github.com/Flank/flank/pull/2360

I'm working on a way to resolve the original issue (rate-limiting) without breaking things, but this pull request will at least fix the new problem in the meantime.

.

## Checklist

- [ ] Documented
- [ ] Unit tested

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -31,7 +31,7 @@ runs:
       shell: bash
     - id: download_flank
       run: |
-        ./flankScripts github download_flank --version=${{ inputs.version }} --token=${{ secrets.GITHUB_TOKEN }}
+        ./flankScripts github download_flank --version=${{ inputs.version }}
       shell: bash
     - id: authentication
       run: |


rfogar2:
@Fry-kun can you approve and merge this?

adamvduke (approved):

adamvduke:
Seems like this partial rollback of #2360 should be safe, because there is a guard against adding the token to the request if it's blank, and it wasn't being passed previously.

https://github.com/Flank/flank/blob/3a213579b0b8ed7ca018314f0619146408292b35/flank-scripts/src/main/kotlin/flank/scripts/data/github/GithubApi.kt#L197-L201

baygaillardclasspass:
I see the PR fixing this got reverted. Did this ever get fixed? We're running into this now that we're on GHA instead of Jenkins where we maintained our own download script.

ryancfogarty:
> I see the PR fixing this got reverted. Did this ever get fixed? We're running into this now that we're on GHA instead of Jenkins where we maintained our own download script.

No sorry, I never got around to working out a fix that didn't break things.

baygaillardclasspass:
I suppose we can't use your github action then :( it's entirely unusable for larger teams.

adamvduke:
> I suppose we can't use your github action then :( it's entirely unusable for larger teams.
@baygaillardclasspass 
Are you specifying a version? I don't work at Google anymore, but from what I remember this issue was largely caused by multiple unauthenticated requests when trying to use the action without a version. Not specifying a version causes the download process to make extra github requests to determine the latest version.

baygaillardclasspass:
I will bring that back to my devs, thank you for the tip!

PR #2370, opened 2023-05-02
fix: add github_token to secrets

acidlynx:
Fixes #

.

## Checklist

- [ ] Documented
- [ ] Unit tested


rfogar2:
This looks like a breaking change. Could we instead make the new required field not required and default it to an empty string? This will have the same behaviour as previously.

andreschab90:
any updates here @acidlynx ?

Fry-kun:
This is stale

PR #2373, opened 2023-06-07
fix: update various dependencies

### LINKED ISSUE #2372: Outdated Google API Client library leads to vulnerabilities
**Describe the bug**
Version 1.35.2 of the Google API Client is vulnerable to the following CVEs due to dependencies it uses:

- [CVE-2022-3510 ](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-3510)
- [CVE-2022-3509](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-3509)
- [CVE-2022-3171](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-3171)

The latest version (2.2.0) is not vulnerable. 

**To Reproduce**

Include flank in a build that checks for CVE issues in dependencies. See a failure do to the above CVEs. 

**Expected behavior**
Flank does not introduce security vulnerabilities in projects that use it. 

**Details**

Refer to [Versions.kt](https://github.com/Flank/flank/blob/master/buildSrc/src/main/kotlin/Versions.kt#L35) to see that Flank is using an outdated version of `GOOGLE_API`.

adamvduke:
Fixes #2372

This change seems large because it is deleting a significant amount of vendored/generated code. Functionally, the bulk of the change is updating a handful of dependencies, which includes a dependency was previously vendored/generated, but can use a pre-built artifact going forward.

Everything under firebase_apis/test_api is vendored/generated based on the json API discovery document for the testing API or similar. Based on [this comment](https://github.com/Flank/flank/blob/ee8b46391715807aab555a6e93953177a8429c0f/test_runner/build.gradle.kts#L205-L208), it seems like this was done to leverage code that wasn't available as part of the publicly available testing api jar at the time. The unit and integration tests pass using the currently available testing api jar, and so it's no longer necessary to maintain the firebase_apis/test_api sub-project, and flank can depend on the newest/current publicly available jar.

Unit and integration tests are still passing

## Checklist

- [ ] Documented
- [X] Unit tested


adamvduke:
@flank-it

adamvduke:
@flank-it

adamvduke:
@flank-it

tonybaroneee (approved):

